PR #68, opened 2018-08-21
search: AND by default

jma:
* change the OR by AND for a default query


iGormilhit:
@jma Could you suggest how to test it? My guess is:

- `git checkout` into the PR
- `pip install -e .[all]`
- restart the app
- ?

I have this book in my local instance: `"title": "Harry Potter à l'école des sorciers"`:
- `q=potter` retrieves it.
- `q=sorciers` does too.
- `q=école sorciers` retrieves nothing, I don't get why. Maybe I'm doing it wrong.

PR #78, opened 2018-09-10
facets

rerowep:
Tests:

See facets working correctly.

### DIFF
--- a/rero_ils/static/css/rero_ils/search.css
+++ b/rero_ils/static/css/rero_ils/search.css
@@ -126,10 +126,6 @@ img.reroils-frontpage-rero21 {
     }
 }
 
-invenio-search-facets li {
-    text-transform: capitalize;
-}
-
 /*****************************************
    Size of the search input, in the navbar
    Smaller for medium screens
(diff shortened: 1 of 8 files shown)


rero_ils/static/templates/rero_ils/range.html
```diff
@@ -1,9 +1,13 @@
...
+          <a data-toggle="collapse" href="#collapse-year" translate>years</a>
+        </h3>
     </div>
-    <div class="panel-body">
-        <div ng-style="{width: '80%', margin: '0 auto'}" id="year_hist"></div>
-        <div ng-style="{width: '80%', margin: '0 auto'}" id="year_select"></div>
+    <div id="collapse-year" class="panel-collapse collapse in">
+      <div class="panel-body">
+          <div ng-style="{width: '80%', margin: '0 auto'}" id="year_hist"></div>
+          <div ng-style="{width: '80%', margin: '0 auto'}" id="year_select"></div>
```
iGormilhit: Styling into the HTML? Is there another way to do that?

rerowep: Ok will change the search.css.
I don't know who has implemented the styling in the range.html. Do you know a better way, maybe it is better to change it?

iGormilhit: `git blame` will tell you who did this in the first place. :smiley: 
It should be in a `[sc]ss` somewhere. I have to investigate how and where.

iGormilhit: @rerowep I tried something for the `range.html`, but it didn't work well. Let's leave it as it is for now.

rerowep: Have updated the CSS and reimplemented the toggle icon. (something went wrong during my last push :-0)

rero_ils/static/css/rero_ils/search.css
```diff
@@ -151,10 +147,35 @@ invenio-search-facets li {
     }
 }
 
+
 invenio-search-range .panel-heading h3,
 invenio-search-facets .panel-heading h3 {
     text-transform: capitalize;
 }
```
iGormilhit: Either suppress lines 151 to 154 or change them into:
```css
invenio-search-range .panel-heading h3:first-letter,
invenio-search-facets .panel-heading h3:first-letter {
    text-transform: uppercase;
}
```

iGormilhit:
One more thing, concerning the commit message. Prefer something like:

> search: facets improvements
>
>* BETTER: reorder the facets and display libraries instead of locations
>* FIX: remove the CSS capitalization
>* NEW: enable the facets collapsing feature

It's compliant with https://github.com/rero/rero-ils/blob/master/CONTRIBUTING.rst#get-started (item number 6).

PR #115, opened 2018-10-16
Patron Types: Add resource

Garfield-fr:
* NEW: Add patron types resource

Co-authored-by: Peter Weber <peter.weber@rero.ch>

### DIFF
--- a/rero_ils/modules/patrons_types/utils.py
+++ b/rero_ils/modules/patrons_types/utils.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of RERO ILS.
+# Copyright (C) 2018 RERO.
+#
+# RERO ILS is free software; you can redistribute it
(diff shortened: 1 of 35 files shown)


rero_ils/modules/patrons_types/utils.py
```diff
@@ -0,0 +1,51 @@
...
+def save_patron_type(data, record_type, record_class, parent_pid):
+    """Save a record into the db and index it.
+
+    If the patron_type does not exists, it well be created
+    and attached to the organisation.
+    """
+    data = clean_dict_keys(data)
+    data['organisation_pid'] = current_patron.organisation.pid
+
+    patrontype = record_class.create(data)
```
BadrAly: If patron_type record exists , only update is needed, else create record.
Now, when you edit a patron type, system will duplicate the patron type instead of updating the record. According to the user story the patron type name is "[unique within the organisation]"

PR #141, opened 2018-11-09
Issue: Identify the two separate displays in the person detailed view

### LINKED ISSUE #137: Identify the two separate displays in the person detailed view
The librarian can chose between two displays in the person detailed view.
The buttons to select the display should provide a mouseover text, for example : "All sources" and "By source".

Garfield-fr:
* NEW: Add title on tab list of presentation
* closes #137


NicolasLabat:
The time until the message is displayed seems OK (< 1 sec.) but not instantaneous like for other messages (ex : can_delete).
We should standardize this in the whole application (time, colors).

PR #149, opened 2018-11-26
replace function

rerowep:
new api function replace

### DIFF
--- a/tests/modules/test_api.py
+++ b/tests/modules/test_api.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of RERO ILS.
+# Copyright (C) 2017 RERO.
+#
+# RERO ILS is free software; you can redistribute it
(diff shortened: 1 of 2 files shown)


tests/modules/test_api.py
```diff
@@ -0,0 +1,39 @@
...
+
+def test_api_update_replace(db, minimal_document_record):
+    """Test api replace."""
+    minimal_document_record['publicationYear'] = 1999
+    doc = Document.create(minimal_document_record)
+    del minimal_document_record['publicationYear']
+    doc.update(minimal_document_record, dbcommit=True)
+    assert doc.get('publicationYear') == 1999
+    doc.replace(minimal_document_record, dbcommit=True)
+    assert doc.get('publicationYear', True)
```
BadrAly: Need to change this assert to something like:
assert not doc.get('publicationYear')

rerowep: changed to not in

PR #154, opened 2018-12-10
API: library is open

rerowep:
* NEW: library is_open function

Signed-off-by: Peter Weber <Peter.Weber@rero.ch>

open the console and try to test the new date functions from library

### DIFF
--- a/rero_ils/modules/libraries/api.py
+++ b/rero_ils/modules/libraries/api.py
@@ -24,12 +24,24 @@
 
 """API for manipulating libraries."""
 
+from datetime import datetime, timedelta
+
+from dateutil import parser
+from dateutil.rrule import FREQNAMES, rrule
 from invenio_search.api import RecordsSearch
 
 from ..api import IlsRecord
+from ..utils import strtotime
 from .fetchers import library_id_fetcher
 from .minters import library_id_minter
 from .providers import LibraryProvider
+
+
+# define Python user-defined exceptions
+class LibraryNeverOpen(Exception):
+    """Raised when the library has no open days."""
+
+    pass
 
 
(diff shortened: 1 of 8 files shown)


rero_ils/modules/libraries/api.py
```diff
@@ -58,3 +70,159 @@ def get_all_library_names(cls):
...
+            times_open = times_open or (time_to_test >= start_time) and \
+                                       (time_to_test <= end_time)
+        return times_open
+
+    def _is_in_period(self, datetime_to_test, exception_date):
+        """Test if date is period."""
+        start_date = parser.parse(exception_date['start_date'])
+        end_date = exception_date.get('end_date')
+        if end_date:
+            end_date = parser.parse(exception_date['end_date'])
```
jma: `parser.parse(end_date)`, no?

rero_ils/modules/libraries/api.py
```diff
@@ -58,3 +70,159 @@ def get_all_library_names(cls):
...
+                open=times_open,
+                date=date,
+                exception_dates=exception_dates,
+                day_only=day_only
+            )
+            if exception != times_open:
+                times_open = not times_open
+        return times_open
+
+    def next_open(self, date=datetime.now(), previous=False):
```
jma: not sure that default parameter calling function works... i.e. `datetime.now()`.

rero_ils/modules/libraries/api.py
rerowep: Yes, is better like this

rero_ils/modules/libraries/api.py
rerowep: I have tested it. It is working.

PR #172, opened 2019-01-14
Layout: item detailed view

iGormilhit:
:warning: Not to be merged into `master`, but into `refactoring-layout` branch (as mentioned above)

* BETTER: split scss for detailed view and page
* BETTER: remove custom margin and padding by bootstrap classes
* BETTER: adapt item detailed view to new page template
* NEW: display document type icon
* FIX: add padding to  the main footer

----------

### To test

- log in as a librarian
- display an item detailed view
- try item with transaction (in transit, borrowed, requested)
- try with large, medium, small screens


NicolasLabat:
Active transactions on an item are displayed on the item detailed view in a section of the page under the title "Transactions". If there are no transactions, this title doesn't appear. 
But for items with status "missing", "Transactions" appears and is empty. It seems to make no sense.
Example : 10000000016

iGormilhit:
@NicolasLabat You're right, I have to fix this. Thanks.

iGormilhit:
@NicolasLabat Fixed, the transaction section appears only if the item is on loan, in transit or is requested.

PR #176, opened 2019-01-22
Layout: person detailed view

iGormilhit:
* BETTER: implement new layout to person detailed views
* BETTER: move person macro to person module

-------

### To test

- `invenio collect -v`
- display `http://localhost:5000/persons/77` (use the tab navigation)
- try with different screens size


BadrAly:
> @BadrAly I've pushed a new version of my commit, fixing this, using a more traditional method to add `:`.

Yes, problem is fixed.

PR #183, opened 2019-01-25
VIEWS: fix $ref relations

jma:
⚠️ the admin views are not yet back

- FIX: document search view
- FIX: document detailed view
- FIX: patron profile view
- FIX: item detailed view
- BETTER: library facet are based on library pid

Signed-off-by: Johnny Mariéthoz <johnny.mariethoz@rero.ch>

### TESTS

- ./scripts/bootstrap
- ./run-tests.sh
- ./scripts/setup

### DIFF
--- a/rero_ils/modules/documents/templates/rero_ils/detailed_view_documents.html
+++ b/rero_ils/modules/documents/templates/rero_ils/detailed_view_documents.html
@@ -1,5 +1,4 @@
 {%- extends config.RECORDS_UI_BASE_TEMPLATE %}
-{% set data = record.dumps() %}
 {% from 'rero_ils/macros/macro.html' import div_row, div_json %}
 
 {%- block page_body %}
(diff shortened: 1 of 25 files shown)


NicolasLabat:
Documents detailed view correctly displayed for librarians but not for patrons.
Error messages : AttributeError: type object 'Item' has no attribute 'requested_by_patron' OR AttributeError: type object 'Item' has no attribute 'loaned_to_patron'

Garfield-fr:
Bug on document list: The status is still not available even if the documents are available.

rero_ils/modules/documents/templates/rero_ils/detailed_view_documents.html
```diff
@@ -1,5 +1,5 @@
 {%- extends config.RECORDS_UI_BASE_TEMPLATE %}
-{% set data = record.dumps() %}
+{% set document = record %}
```
Garfield-fr: Why you don't use directly record on template ?

jma: It is exactly what I do. `record.dump()` is useless with `$ref` and instead of using a generic name `data`, I think it is better to use a more specific name variable i.e. `document`. Any other suggestion?

PR #208, opened 2019-03-05
admin UI: fix and translations

iGormilhit:
* NEW Adds a label for search views.
* BETTER Improves the result count message.
* BETTER Improves the patrons roles display.
* FIX Fixes the layout for search views without facets.
* FIX Fixes missing translations.
* BETTER Improves patron types, item types, circulation policies fixtures.
* FIX Pins isort to >=4.3.5,<4.3.10
* Corrects some typos.
* Removes a remain JSON print on the item detailed view template.

----

## To test

- `./script/bootstrap`
- `./script/setup`
- Check the translation in the professional UI.
- Check the layout in the professional UI, with or without facets.
- Check the display of patron roles in the professional UI patrons search (brief views).
- Check the patron types, item types and circulation policies fixtures (name and description).


data/circulation_policies.json
```diff
@@ -16,8 +16,8 @@
   },
   {
     "$schema": "http://ils.rero.ch/schema/circ_policies/circ_policy-v0.0.1.json",
-    "name": "library: short",
-    "description": "library: short loan (7 days)",
+    "name": "Site d'Aoste: short",
```
BadrAly: if a librarian adds this policy to another library, then the policy name is no longer valid.

rero_ils/modules/loans/cli.py
```diff
@@ -185,8 +185,8 @@ def get_loanable_items():
     """."""
     # TODO: to be replace by circulation policies
     exclude_item_type_pids = [
-        ItemType.get_pid_by_name('no-checkout'),
-        ItemType.get_pid_by_name('on-site')
+        ItemType.get_pid_by_name('No checkout'),
```
BadrAly: We need to move this logic to circ policies ASAP. if a librarian changes the name of the item type, this will not work

rero_ils/modules/loans/jsonschemas/loans/loan-ils-v0.0.1.json
```diff
@@ -50,10 +50,10 @@
       "type": "string",
       "title": "Request pickup location ID"
     },
-    "request_expire_date": {
+    "request_expiry_date": {
```
BadrAly: I do not know if this will break `invenio_circulation` or not. The code of `invenio_circulation`  rely on 'request_expire_date'

rero_ils/modules/loans/mappings/v6/loans/loan-ils-v0.0.1.json
```diff
@@ -34,7 +34,7 @@
         "start_date": {
           "type": "date"
         },
-        "request_expire_date": {
+        "request_expiry_date": {
```
BadrAly: same here

data/circulation_policies.json
iGormilhit: Ok. We need the input of the sleeping Consortium WG on this.

rero_ils/modules/loans/cli.py
iGormilhit: Yes, you're right. I think that's why there's a `TODO` on line 186. It was already there.

rero_ils/modules/loans/jsonschemas/loans/loan-ils-v0.0.1.json
iGormilhit: In `invenio-circulation` this string is to be find only in the schemas and the mappings.

https://github.com/inveniosoftware/invenio-circulation/search?q=request_expire_date&unscoped_q=request_expire_date

iGormilhit:
@BadrAly On the patron ID, library ID, and also item type ID, you're right. I may be wrong, but I think that serializers are needed to have this data in the record in the first place, which I didn't edit or write.

BadrAly:
New names of `itty`, `ptty` and `cipo` are better thank before.

PR #212, opened 2019-03-11
SERIALIZER: Fix resolve

Garfield-fr:
* FIX: check value on resolve serializer


rero_ils/modules/serializers.py
```diff
@@ -42,7 +42,8 @@ def serialize(self, pid, record, links_factory=None, **kwargs):
         :param links_factory: Factory function for record links.
         """
         if request and request.args.get('resolve'):
```
jma: This can be done in one line:
```python
if request and request.args.get('resolve') == '1':
```

Garfield-fr: done

PR #228, opened 2019-03-20
user interface: menus structure

iGormilhit:
* NEW Adds a configuration for menu categories icons.
* BETTER Attributes categories to organize submenus.
* BETTER Adds filter to remove special characters in categories name.

---

## To test

- restart your app
- check the categorisation of menus


Garfield-fr:
Is it possible to put the Admin menu last ?

iGormilhit:
@Garfield-fr Not at this point. But you're right, according to the specifications, we need to be able to decide the order of the main menu and submenus, so in a future PR, we will rewrite the template.

PR #237, opened 2019-03-23
Persons language mappings

rerowep:
to test: run setup (documents have to be recreated)

change language in interface and look for author name changes


rerowep:
Attention: Setup takes quite a while because of 9000 MEF persons references !!!

PR #251, opened 2019-04-02
library admin: improvement

### LINKED ISSUE #222: Overlap of opening hours
When editing library opening hours, the system indicates a false overlap when the eary period is entered after the late period (ex: 14.00-16.00, 10.00-12.00).
It should accept the two periods and reorder them automatically.

Garfield-fr:
* NEW Reorders opening hours for 1 day if the second hour is less than the first hour
* BETTER Improves the overlap validator if the second hour is less than the first hour
* FIX Closes #222

### DIFF
--- a/ui/src/app/records/custom-editor/libraries/library.ts
+++ b/ui/src/app/records/custom-editor/libraries/library.ts
@@ -55,6 +55,7 @@ export class Library {
   update(obj) {
     Object.assign(this, obj);
     this.cleanOpeningHours(this.opening_hours);
+    this.reorderOpeningHours(this.opening_hours);
   }
 
   createOpeningHours() {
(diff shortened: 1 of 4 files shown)


ui/src/app/records/custom-editor/libraries/library.ts
```diff
@@ -88,6 +89,16 @@ export class Library {
     });
   }
 
+  reorderOpeningHours(openingHours) {
+    openingHours.forEach(opening => {
+      if (opening.times.length > 1) {
+        opening.times.sort(function(a, b) {
```
jma: you can use inline function such as:
```typescript
...sort((a,b) => moment(a.start_time, 'HH:mm').diff(moment(b.start_time, 'HH:mm')) );
```

Garfield-fr: done

PR #256, opened 2019-04-03
Circulation: renewal due date from current_date

### LINKED ISSUE #231: Renewal date
Current situation:
The renewal date is calculated on the basis of the due date.

Desired situation:
The renewal date should be calculated on the basis of the current day (current day + renewal duration). The renewal should be blocked if the new due date is equal or before the current due date.

BadrAly:
* NEW Blocks renewal if renewal due date less than checkout due date
* FIX Fixes #231.

- Create the following scenario:
  Checked-out item its renewal-due-date is less than its check-out-due-date
 The renewal will be rejected

- For all renewals, the renewal-due-date is calculated from current-date

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -34,6 +34,7 @@ You overwrite and set instance-specific configuration by either:
 from __future__ import absolute_import, print_function
 
 from datetime import timedelta
+from functools import partial
 
 from invenio_circulation.pidstore.pids import CIRCULATION_LOAN_FETCHER, \
     CIRCULATION_LOAN_MINTER, CIRCULATION_LOAN_PID_TYPE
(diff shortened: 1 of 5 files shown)


rero_ils/config.py
```diff
@@ -975,9 +975,12 @@ def _(x):
         item_available=is_item_available_for_checkout
     ),
     extension=dict(
-        from_end_date=True,
-        duration_default=get_default_extension_duration,
-        max_count=get_default_extension_max_count
+        from_end_date=partial(
```
iGormilhit: Do we actually *need* this parameters? Will it *ever* be `True`?

BadrAly:
@pronguen PR is ready for your testing @ ils.dev.rero.ch

pronguen:
It is for now impossible to test consistently:

- When check-out duration is set to 3 days, due date given is 15th April (we are 11th April, all days are marked as open in the library, result should be 14th April)
- On the same item, when renewal duration is set to 10 days, due date given is 26th April (we are 11th April, all days are marked as open in the library, result should be 21th April)

The first extension is not based on the current date

BadrAly:
> * When check-out duration is set to 3 days, due date given is 15th April (we are 11th April, all days are marked as open in the library, result should be 14th April)

14th April is a week-end normally library is closed. If you are using the library `Bibliothèque cantonale valdôtaine, site de Pont-saint-Martin` it is also closed on Mondays. So the 16th is correct

pronguen:
I did the test with Astrid, who is working at "Bibliothèque du Lycée de la Vallée d'Aoste". I checked the opening hours, and changed them to have all days open, in order to simplify the test.

PR #267, opened 2019-04-17
circulation: possibility to check-out in-transit items

### LINKED ISSUE #230: Check-out of an item "in transit"
An item which is "in transit" cannot be checked out. This should be possible, if the item has no request, or if the request is for the user concerned by the check-out.

BadrAly:
* FIX Fixes #230: allows checkout of in-transit items.

This two scenarios works:

- checkout out an item (having no requests) with loan_status= ITEM_IN_TRANSIT_TO_HOUSE to any patron.

- checkout out an item with loan_status= ITEM_IN_TRANSIT_FOR_PICKUP to any requested patron from any checkout library.

### DIFF
--- a/rero_ils/modules/items/api.py
+++ b/rero_ils/modules/items/api.py
@@ -125,11 +125,18 @@ def add_loans_parameters_and_flush_indexes(function):
             if not patron_pid:
                 raise CirculationException(
                     "Patron PID not specified")
-            data = {
-                'item_pid': item.pid,
-                'patron_pid': patron_pid
-            }
-            loan = Loan.create(data, dbcommit=True, reindex=True)
+            if function.__name__ == 'checkout':
+                request = get_request_by_item_pid_by_patron_pid(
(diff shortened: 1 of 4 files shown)


pronguen:
2 problems observed

# Use case 1

1. The item of Lib A is requested by Simonetta, with pick-up in Lib B
1. The item is checked-out in Lib C for Simonetta, without action "receive"
1. Check-out is allowed, but the request persist

Problem: the request of Simonetta should be satisfied at the check-out.

# Use case 2

1. The item of Lib A is requested by Simonetta, with pick-up in Lib B
1. The item is checked-out in Lib C for Alain, without action "receive"
1. Check-out is allowed

Problem: check-out should be allowed only for the first patron in the request queue (for now). Later, the librarian will have the option to override the limit.

This meen that the second scenario described in this pull request is not correct (not conform to https://github.com/rero/rero-ils/issues/230):

> checkout out an item with loan_status= ITEM_IN_TRANSIT_FOR_PICKUP to any requested patron from any checkout library

PR #269, opened 2019-04-17
page not found and handle http error (angular)

Garfield-fr:
NEW Adds route wildcard for Page Not Found (404)
NEW Adds interceptor to catch http error
Signed-off-by: Bertrand Zuchuat bertrand.zuchuat@rero.ch

How to test:

start angular server with pipenv run npm run proxy
Test on http://localhost:4200
Test url that does not exist. You are a "Page not found" error
Test for interceptor: Logout the current Url. Check url http://localhost:4200/records/item_types


jma:
This approach using interceptor break some existing logic. In some cases the services check the http status and return null if a 404 status is returned. Try search: `e.status` in the code. In the interface you can go to circulation checkinout and enter a non existing barcode.

ui/src/app/core/interceptors/httpError.interceptor.ts
```diff
@@ -0,0 +1,42 @@
...
+    private exceptionService: ExceptionService,
+    private router: Router,
+    private responseStatusService: ResponseStatusService
+  ) {}
+
+  intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    return next.handle(request).pipe(
+        retry(1),
+        catchError((error: HttpErrorResponse) => {
+          if (error.status !== 200) {
```
jma: This approach using interceptor break some existing logic. In some cases the services check the http status and return null if a 404 status is returned. Try search: e.status in the code. In the interface you can go to circulation checkinout and enter a non existing barcode.

Garfield-fr: Intercept http error equal or greater than 400 except 404

PR #278, opened 2019-05-06
user interface: pickup library instead of pickup location

jma:
* BETTER Replace pickup location by pickup library in the patron profile and circulation ui.
* BETTER Remove useless column in the patron profile.
* Changed name variable to be more clear.

### To Test
- boostrap

### DIFF
--- a/rero_ils/modules/loans/api.py
+++ b/rero_ils/modules/loans/api.py
@@ -104,9 +104,11 @@ class Loan(IlsRecord):
 
         if loan.get('pickup_location_pid'):
             location = Location.get_record_by_pid(loan['pickup_location_pid'])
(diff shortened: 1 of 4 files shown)


jma:
> Select a Pickup Location
> To be consistent, you may want to change this as well:
> 
> `<h6 class="dropdown-header">{{ _('Select a Pickup Location') }}</h6> `
> to:
> `<h6 class="dropdown-header">{{ _('Select a Pickup Library') }}</h6>`
> in
> `modules/documents/templates/rero_ils/detailed_view_documents.html`
> 
> Other than that, approved.

Here location is from the user point of view, not the `Location` instance class.

PR #282, opened 2019-05-07
global: test coverage and docs for locations

BadrAly:
* BETTER increase test coverage and add docs for locations.

./run_tests should show that modules/locations coverage is at 100%


BadrAly:
> Could we fix `celery` to version 4.2.0 here?
> `pip install celery==4.2.0`
> 
> We have following error during startup of server:
> 
> ```
>  File "msgpack/_packer.pyx", line 281, in msgpack._packer.Packer._pack
> kombu.exceptions.EncodeError: can't serialize datetime.datetime(2019, 5, 8, 6, 59, 31, 694863)
> ```
It is probably better to have a different commit for this.

PR #296, opened 2019-05-14
fixtures: load prepared loans json file

### LINKED ISSUE #254: Fixtures: write a better and faster way to create circulation transactions
scripts/setup:

Prepare a file of circulation transactions to load to database instead of the current method to randomly create transactions. Similar to items.json

BadrAly:
* FIX Fixes #254.

./run_tests.sh
./scripts/setup.sh


data/loans_generate.json
```diff
@@ -0,0 +1,89 @@
+Tree
```
jma: Not well indented

BadrAly: done

PR #297, opened 2019-05-15
rest API: access restriction by organisation

BadrAly:
* NEW allows access to records for users of same organisation.

* login with a user from an organisation, you should not be able to access all  type of records of the second organisation:
URL : http://localhost:5000/api/<pid_type>/< pid>

### DIFF
--- a/rero_ils/modules/api.py
+++ b/rero_ils/modules/api.py
@@ -307,3 +307,8 @@ class IlsRecord(Record):
     def can_delete(self):
         """Record can be deleted."""
         return len(self.reasons_not_to_delete()) == 0
+
+    @property
+    def org_pid(self):
+        """Get organisation pid for circulation policy."""
+        return self.replace_refs()['organisation']['pid']
--- a/rero_ils/modules/locations/api.py
+++ b/rero_ils/modules/locations/api.py
@@ -100,3 +100,16 @@ class Location(IlsRecord):
         if links:
             cannot_delete['links'] = links
         return cannot_delete
+
+    @property
+    def library_pid(self):
+        """Get library pid for location."""
+        return self.replace_refs()['library']['pid']
+
+    @property
+    def org_pid(self):
+        """Get organisation pid for location."""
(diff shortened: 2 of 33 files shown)


rero_ils/modules/api.py
```diff
@@ -307,3 +307,8 @@ def reasons_not_to_delete(self):
     def can_delete(self):
         """Record can be deleted."""
         return len(self.reasons_not_to_delete()) == 0
+
+    @property
+    def org_pid(self):
```
jma: please replace by `organisation_pid`.

rero_ils/modules/api.py
```diff
@@ -307,3 +307,8 @@ def reasons_not_to_delete(self):
...
+        """Get organisation pid for circulation policy."""
+        return self.replace_refs()['organisation']['pid']
```
jma: need to check if organisation property exists, else returns `None`.

rero_ils/modules/locations/api.py
```diff
@@ -100,3 +100,16 @@ def reasons_not_to_delete(self):
...
+
+    @property
+    def library_pid(self):
+        """Get library pid for location."""
+        return self.replace_refs()['library']['pid']
+
+    @property
+    def org_pid(self):
+        """Get organisation pid for location."""
+        from ..libraries.api import Library
```
jma: check if it is possible put this at the top of the file.

rero_ils/modules/patrons/api.py
```diff
@@ -212,8 +212,19 @@ def initial(self):
     @property
     def patron_type_pid(self):
         """Shortcut for patron type pid."""
-        patron_type_pid = self.get('patron_type').get('pid')
-        return patron_type_pid
+        from ..patron_types.api import PatronType
+        if self.get('patron_type'):
+            ptty = PatronType.get_record_by_pid(self['patron_type']['pid'])
```
jma: should be a `$ref` no? In such case use `replace_refs`.

PR #305, opened 2019-05-17
Cover art

rerowep:
For the test the `bootstrap` must be run.
To see the cover art for ebooks the data has to have cover_art links in the data.
Tis cover_art links are created if [PR303](https://github.com/rero/rero-ils/pull/303)
is used for harvesting data from the ebooks server with harvested data from API [PR24](https://github.com/rero/rero-ebooks/pull/24)


jma:
I think this is not the right approach. This can be done directly on the jinja and angular template using a `if else` case. No?

rerowep:
> I think this is not the right approach. This can be done directly on the jinja and angular template using a `if else` case. No?

The hole image class is constructed in the js, this is the reason why I opted for this solution.

PR #310, opened 2019-05-22
config: Sentry support

jma:


rerowep:
We have a `test_ping` do we need a `test_error`?

PR #322, opened 2019-05-29
indexing: update item types mapping

Garfield-fr:
* Adds analyzer on name and description fields.
* Adds es config on test to support templates.
* Fixes type on description field.

How to test:
- `script/setup`
- `http://localhost:5000/admin/records/item_types`
- Find a term and check if the result is correct (value in fields name or description)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -30,7 +30,9 @@ from os.path import dirname, join
 
 import mock
 import pytest
+from elasticsearch.exceptions import RequestError
 from invenio_circulation.proxies import current_circulation
+from invenio_search import current_search, current_search_client
 from utils import flush_index, mock_response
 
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -52,6 +52,39 @@
...
+def es(appctx):
+    """Setup and teardown all registered Elasticsearch indices.
+
+    Scope: module
+    This fixture will create all registered indexes in Elasticsearch and remove
+    once done. Fixtures that perform changes (e.g. index or remove documents),
+    should used the function-scoped :py:data:`es_clear` fixture to leave the
+    indexes clean for the following tests.
+    """
+    from elasticsearch.exceptions import RequestError
```
jma: please put it at the top of the file if it is possible

Garfield-fr: done

PR #323, opened 2019-06-04
permissions: allows librarian to manipulate users of its library only

BadrAly:
* Reduces librarian rights to the library level for the resource patron


rero_ils/permissions.py
```diff
@@ -86,6 +86,7 @@ def can_delete_organisation_records_factory(record, *args, **kwargs):
 
     user must have librarian or system_librarian role.
     returns False if a librarian tries to delete a system_librarian.
+    false if librarian tries to delete a librarian from another library.
```
jma: can be rewords as: "returns False if a librarian tries to delete a system_librarian and if librarian tries to delete a librarian from another library."

PR #330, opened 2019-06-06
indexing: update libraries mapping

Garfield-fr:
* Adds analyzer on name field.
* Improves global analyzer.

How to test:

* `script/setup`
* http://localhost:5000/admin/records/libraries
* Find a term and check if the result is correct (value in fields name or description)


tests/data/data.json
```diff
@@ -415,6 +415,107 @@
       }
     ]
   },
+  "lib5": {
+    "$schema": "https://ils.rero.ch/schema/libraries/library-v0.0.1.json",
+    "address": "Centre culturel",
+    "code": "AOSTE",
+    "email": "reroilstest+aoste@gmail.com",
+    "name": "Bibliothèque d'Aoste",
```
BadrAly: The `org2` is the `district of Sion`, better to change the name of the library from `Aoste` to one of these:
`Arbaz`, `Grmisuat`, `Savièse`, ou `Vesyonnaz`

tests/fixtures/organisations.py
```diff
@@ -160,6 +160,24 @@ def lib_sion(app, org_sion, lib_sion_data):
     return lib
 
 
+@pytest.fixture(scope="module")
+def lib_aoste_data(data):
```
BadrAly: same here

tests/fixtures/organisations.py
```diff
@@ -160,6 +160,24 @@ def lib_sion(app, org_sion, lib_sion_data):
...
+    """Aoste library data."""
+    return deepcopy(data.get('lib5'))
+
+
+@pytest.fixture(scope="module")
+def lib_aoste(app, org_sion, lib_aoste_data):
```
BadrAly: same here

tests/data/data.json
Garfield-fr: done

tests/fixtures/organisations.py
Garfield-fr: done

tests/fixtures/organisations.py
Garfield-fr: done

PR #333, opened 2019-06-07
indexing: update mef persons mapping

Garfield-fr:
* Adds analyzer on biographical_information, preferred_name_for_person,
  authorized_access_point_representing_a_person and variant_name_for_person fields.

How to test:

* `script/setup`
* http://localhost:5000/admin/records/persons
* Find a term and check if the result is correct


jma:
it would be great to have in the detailed message commit an explanation of what is a "global char_group tokenizer" or better the name and the description of the new analyzer.

tests/data/data.json
```diff
@@ -420,7 +420,7 @@
     "address": "Centre culturel",
     "code": "APROZ",
     "email": "reroilstest+aproz@gmail.com",
-    "name": "Bibliothèque d'Aproz",
+    "name": "Biblioth\u00e8que d'Aproz",
```
jma: both (é and \u00e8)should be supported

PR #342, opened 2019-06-14
search: boosting fields on resources

Garfield-fr:
* Adds fields boosting configuration.
* Improves query parser to use the boosting configuration.

How to test:
* Upgrade Elasticsearch to version 6.6.0 (show in docker-services.yml)
* `script/setup`
* http://localhost:5000
* Find some terms and check if the result is correct (show boosting in config.py)

### DIFF
--- a/rero_ils/query.py
+++ b/rero_ils/query.py
@@ -29,6 +29,7 @@ from __future__ import absolute_import, print_function
 from elasticsearch_dsl.query import Q
 from flask import current_app, request
 from invenio_records_rest.errors import InvalidQueryRESTError
+from invenio_records_rest.sorter import eval_field
 
 from .modules.patrons.api import current_patron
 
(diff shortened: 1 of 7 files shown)


rero_ils/query.py
```diff
@@ -49,28 +49,48 @@ def and_search_factory(self, search, query_parser=None):
     :param search: Elastic search DSL search instance.
     :returns: Tuple with search instance and URL arguments.
     """
```
jma: Please add boosting support in the comment.

rero_ils/query.py
```diff
@@ -49,28 +49,48 @@ def and_search_factory(self, search, query_parser=None):
...
-    def _default_parser(qstr=None):
+    def _default_parser(qstr=None, query_boosting=[]):
         """Default parser that uses the Q() from elasticsearch_dsl."""
         if qstr:
-            return Q('query_string', query=qstr, default_operator='AND')
+            if len(query_boosting) == 0 or ':' in qstr:
```
jma: `if query_boosting...`

rero_ils/query.py
```diff
@@ -49,28 +49,48 @@ def and_search_factory(self, search, query_parser=None):
...
+                return Q('query_string', query=qstr, default_operator='AND')
+            if qstr and len(query_boosting) > 0:
```
jma: `if qstr and query_boosting`

rero_ils/query.py
```diff
@@ -49,28 +49,48 @@ def and_search_factory(self, search, query_parser=None):
...
+                return Q('bool', should=[
```
jma: We need to find an other approach as boosting does not use query string.

tests/fixtures/metadata.py
```diff
@@ -100,6 +100,24 @@ def ebook_3(app, ebook_3_data):
...
 
+@pytest.fixture(scope="module")
+def ebook_4_data(data):
+    """Load ebook 4 data."""
+    return deepcopy(data.get('ebook4'))
+
+
+@pytest.fixture(scope="module")
+def ebook_4(app, ebook_4_data):
+    """Load document record."""
```
jma: please change the comment

rero_ils/query.py
Garfield-fr: done

rero_ils/query.py
Garfield-fr: done

rero_ils/query.py
Garfield-fr: done

tests/fixtures/metadata.py
Garfield-fr: done

rero_ils/query.py
Garfield-fr: done

PR #347, opened 2019-06-18
US341: Email notifications

BadrAly:


BadrAly:
> Notification "due date" and "availability" work fine!
> 
> One improvements can be made on circulation policy settings:
> 
>     * Currently "Days before due date" cannot be equal 0
> 
>     * It should be possible to set 0: in that case, the notification is sent exactly at the due date.
> 
> 
> PR can be integrated then.

done

PR #371, opened 2019-07-02
document: Publisher format

sebastiendeleze:
FIX Ignore empty publisher in dojson conversion.

Signed-off-by: Sébastien Délèze <sebastien.deleze@rero.ch>

### DIFF
--- a/rero_ils/modules/documents/dojson/contrib/marc21tojson/model.py
+++ b/rero_ils/modules/documents/dojson/contrib/marc21tojson/model.py
@@ -290,11 +290,11 @@ def marc21_to_publishers_publicationDate(self, key, value):
                 self['freeFormedPublicationDate'] = data
         indexes[tag] = index + 1
         lasttag = tag
-    publishers.append(publisher)
-    if publishers == [{}]:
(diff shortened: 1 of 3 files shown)


sebastiendeleze:
My proposition for ignoring empty publishers and so avoid {} in publishers array. I'm not sure if that works because I cannot setup the project locally..

rero_ils/modules/documents/dojson/contrib/marc21tojson/model.py
```diff
@@ -290,11 +290,9 @@ def marc21_to_publishers_publicationDate(self, key, value):
...
         indexes[tag] = index + 1
         lasttag = tag
-    publishers.append(publisher)
-    if publishers == [{}]:
-        return None
-    else:
-        return publishers
+
+    if publisher:
+        publishers.append(publisher)
```
BadrAly: I think your are missing `return something`?

sebastiendeleze: Yes of course, sorry..

PR #373, opened 2019-07-03
fixtures: fix dojson publishers conversion (3rd time)

### LINKED ISSUE #367: TypeError: 'NoneType' object is not iterable
Sentry entry: https://sentrydev.test.rero.ch:8443/share/issue/0c089092b63c4994b3bfd1b5d79b41d4/

This affects all documents detailed views without publishers data ~with items that have circulation transactions~ of the deployed `v0.2.0` release (`ils.test.rero.ch`).

iGormilhit:
* FIX Fixes #367.
* FIX Makes sure dojson conversion does not return an empty array for
publishers.
* FIX Asserts in the tests that there's not empty array for publishers.

## to test

- `./run-tests.sh`
- `./script/setup` should succeed.

### DIFF
--- a/rero_ils/modules/documents/dojson/contrib/marc21tojson/model.py
+++ b/rero_ils/modules/documents/dojson/contrib/marc21tojson/model.py
@@ -293,8 +293,10 @@ def marc21_to_publishers_publicationDate(self, key, value):
 
     if publisher:
         publishers.append(publisher)
-
-    return publishers
+    if not publishers:
+        return None
+    else:
+        return publishers
 
 
 @marc21tojson.over('formats', '^300..')
--- a/tests/unit/test_documents_dojson.py
+++ b/tests/unit/test_documents_dojson.py
@@ -435,7 +435,7 @@ def test_marc21_to_publishers_publicationDate():
     """
     marc21json = create_record(marc21xml)
     data = marc21tojson.do(marc21json)
-    assert not data.get('publishers')
+    assert 'publishers' not in data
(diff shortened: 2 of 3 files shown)


tests/unit/test_documents_dojson.py
```diff
@@ -436,6 +436,7 @@ def test_marc21_to_publishers_publicationDate():
     marc21json = create_record(marc21xml)
     data = marc21tojson.do(marc21json)
     assert not data.get('publishers')
+    assert data.get('publishers') != []
```
sebastiendeleze: I think you can replace asserts : 
```
assert not data.get('publishers')
assert data.get('publishers') != []
```

by:
```
assert "publishers " not in data
```
It should be enough to test that publishers key does not exist.

sebastiendeleze: But it's a detail

iGormilhit: Ok, corrected. Seems to work, indeed.

rero_ils/modules/documents/dojson/contrib/marc21tojson/model.py
```diff
@@ -293,8 +293,10 @@ def marc21_to_publishers_publicationDate(self, key, value):
 
     if publisher:
         publishers.append(publisher)
-
-    return publishers
+    if not publishers:
+        return None
+    else:
```
BadrAly: Optional: The `else` here is not necessary, you can safely delete line #298
`setup.py` works fine, all is good

PR #395, opened 2019-07-11
[spike] holdings: introduce holding level

BadrAly:
* Adds holdings level between documents and items.
* Adds automatic holding creation for new items based on item type and item location.
* Adds automatic attaching of new items to a holding record if a valid holding record exists.
* Adds automatic holding deletion once last item is deleted.
* Adds automatic relinking if necessary to a new holding with item is updated.
* Adds loading fixtures is now using reroils class instead of invenio record class.

# How to Test:
bootstrap
setup

1. create an item then check the json of the item to see if there is a holding field is added.
2. you can display the holding record via '/api/holdings/XX'
3. delete the  item and make sure the holding is also deleted by checking   '/api/holdings/XX'

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -56,6 +56,7 @@ from rero_ils.utils import get_agg_config
 
 from .modules.circ_policies.api import CircPolicy
 from .modules.documents.api import Document
+from .modules.holdings.api import Holding, HoldingsIndexer
 from .modules.item_types.api import ItemType
 from .modules.items.api import Item, ItemsIndexer
 from .modules.libraries.api import Library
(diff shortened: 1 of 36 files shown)


BadrAly:
>     2. Why is the item_type inside a group with location (see below)? Moreover, this group is repeatable (array) but it should not. The initial aim was to rename item_type with circulation_type.

Fixed.

```
"circulation_type": {
     "$ref": "https://ils.rero.ch/api/item_types/4"
},
"document": {
    "$ref": "https://ils.rero.ch/api/documents/300"
},
"location": {
    "$ref": "https://ils.rero.ch/api/locations/10"
}
```

BadrAly:
> 1. We should really think about the relevance of duplicating item types, location and document in the item, since it could create some inconsistency.

Once the `circulation` module is adapted, we can safely remove the location/item/document `refs` from the item record.

PR #402, opened 2019-07-13
documentation: update INSTALL.rst

vrabe:
The tag name has been changed to `rero/rero-ils-base` in Dockerfile.
But it remains the old tag name here.


iGormilhit:
@vrabe By the way, thanks for this! :slightly_smiling_face:

PR #410, opened 2019-07-25
circulation: upgrade to invenio circulation v1.0.0a14

BadrAly:
* Adds $ref to item in then loan record.
* Improves circulation exceptions and parameters.

# To test:
- bootstrap

### DIFF
--- a/tests/api/test_loans_rest.py
+++ b/tests/api/test_loans_rest.py
@@ -136,6 +136,7 @@ def test_due_soon_loans(client, librarian_martigny_no_email,
     item = item_lib_martigny
     item_pid = item.pid
     patron_pid = patron_martigny_no_email.pid
+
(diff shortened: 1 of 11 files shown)


tests/api/test_loans_rest.py
```diff
@@ -198,8 +198,8 @@ def test_overdue_loans(client, librarian_martigny_no_email,
     patron_pid = patron_martigny_no_email.pid
     assert not item.is_loaned_to_patron(patron_martigny_no_email.get(
         'barcode'))
-    assert item.can_delete
-    assert item.available
+    assert not item.can_delete
```
jma: any explanation for this? Was it a bug?

BadrAly: Probably is a fixture problem, I removed it because this test already done in the same file in an earlier function. I also created a task to ensure that `item.can_delete` considers only active loans.

PR #411, opened 2019-07-26
tests: optional execution of external services tests.

BadrAly:
* Moves all external services tests to a separate test file.
* Adds script for executing only external services tests.
* Excludes external services tests from main tests script.

# How to test:

- ./run_tests.sh
- ./run_tests.sh external

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,6 +87,7 @@ before_script:
 
 script:
   - ./run-tests.sh # E2E tests are skipped if E2E env is set to 'no'
+  - ./run-tests.sh external # test external services only
 
 after_success:
   - coveralls
--- a/pytest.ini
+++ b/pytest.ini
@@ -26,7 +26,7 @@
 pep8ignore =
     docs/conf.py ALL
     config.py E501
-addopts = --pep8 --doctest-glob="*.rst" --doctest-modules --cov=rero_ils --cov-report=term-missing --ignore=setup.py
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -87,6 +87,7 @@ before_script:
 
 script:
   - ./run-tests.sh # E2E tests are skipped if E2E env is set to 'no'
+  - ./run-ext-tests.sh # test external services only
```
jma: Please use the same script with an option.

pytest.ini
```diff
@@ -26,7 +26,7 @@
 pep8ignore =
     docs/conf.py ALL
     config.py E501
-addopts = --pep8 --doctest-glob="*.rst" --doctest-modules --cov=rero_ils --cov-report=term-missing --ignore=setup.py
+addopts = --pep8 --doctest-glob="*.rst" --doctest-modules --cov=rero_ils --cov-report=term-missing --ignore=setup.py -m "not externaltest"
```
jma: instead of ignoring tests you can use a specific pytest options such as: http://doc.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

.travis.yml
BadrAly: done

pytest.ini
BadrAly: as discussed, it is better to use the ignore option

PR #456, opened 2019-08-23
oaiharvesting: bulk indexing of oai records

rerowep:
* Improves indexing of harvested oai records.


rero_ils/modules/utils.py
```diff
@@ -30,3 +32,29 @@ def strtotime(strtime):
         hour=int(splittime[0]),
         minute=int(splittime[1])
     )
+
+
+def bulk_index(uuids, process=False, verbose=False):
+    """Bulk index records."""
+    if verbose:
+        click.echo(' add to index: {count}'.format(count=len(uuids)))
```
Garfield-fr: Remove space before "add"

PR #461, opened 2019-08-26
 circulation: holdings level adaptation

BadrAly:
* Adapts circulation policies to look for items info from parent holding record.


rero_ils/modules/items/api.py
```diff
@@ -482,6 +481,17 @@ def item_type_pid(self):
             item_type_pid = item_type.get('pid')
         return item_type_pid
 
+    @property
+    def holding_circulation_category_pid(self):
```
jma: can you add tests for the new properties?

BadrAly: done

PR #472, opened 2019-09-03
fixtures: generate new files

BadrAly:
* Replaces big 10k (items, holdings, docs) files by small 1k files.
* Improves code to insert pids to identifier table.
* Adds extra language codes, needed for the transformation.
* Fixes import problem for the transformation.


rero_ils/dojson/cli.py
```diff
@@ -17,6 +17,7 @@
 
 """Click command-line utilities."""
 
+import json
```
jma: why?

rero_ils/modules/documents/jsonschemas/documents/document-minimal-v0.0.1.json
```diff
@@ -89,18 +89,21 @@
             "description": "Language value.",
             "type": "string",
             "enum": [
-              "fre",
```
jma: why did you change the list of languages? If it is really the case, you need to change also the form option.

rero_ils/dojson/cli.py
BadrAly: This is to correct a bug when generating the json file.

rero_ils/modules/documents/jsonschemas/documents/document-minimal-v0.0.1.json
BadrAly: As discussed, adding a note to the commit message

BadrAly:
> It can be a good idea to change the fixture file names to something more explicit such as big instead of 1k, small instead of 300. Includes the resource name in the fixtures file name would be also good.

done

scripts/setup
```diff
@@ -115,12 +118,12 @@ pipenv run invenio fixtures create --pid_type doc --schema 'http://ils.rero.ch/s
 
 # create items
 # NEEDS item type !!!
-# to generate 10k items file 10k documents must exist in DB
-# pipenv run invenio fixtures create_items -i 5 -t ${DATA_PATH}/items10k.json -h ${DATA_PATH}/holdings10k.json
-# pipenv run flask utils check_json -r data/holdings10k.json
+# to generate 1k items file 1k documents must exist in DB
```
rerowep: 1change to `1k` and `big`

scripts/setup
```diff
@@ -115,12 +118,12 @@ pipenv run invenio fixtures create --pid_type doc --schema 'http://ils.rero.ch/s
...
+# pipenv run invenio fixtures create_items -i 2 -t ${DATA_PATH}/items_big.json -h ${DATA_PATH}/holdings_big.json
+# pipenv run flask utils check_json -r data/holdings_big.json
 # to generate 300 items file 300 documents must exist in DB
```
rerowep: change `300` to `small`

PR #473, opened 2019-09-03
editor: shows/hides main (1th level) fields

jma:
* Add buttons to editor to hide or show optional fields.
* With initial data, populated fields are diplayed, empty fields are
hidden.
* All document fields are available in the editor.
* Fixe language and identifiedBy fields for BNF importation.


jma:
> Some minor changes in the commit message:
> 
> ```
> editor: shows/hides main (1st level) fields
> 
> * Adds buttons to editor to hide or show optional fields.
> * With initial data, populated fields are displayed, empty fields are
> hidden.
> * All document fields are available in the editor.
> * Fixes language and identifiedBy fields for BNF importation.
> ```

Done

jma:
> Problems:
> 
> * In 2nd-level fields, selecting text with the mouse is not possible. However, it is possible in the 1st-level fields.

Not fixed, you can open an issue if you want.

> * Repetitive subfields (publisher.name, publisher.place) are not fully displayed in the editor: only one is displayed.

This is an issue of the editor library. An issue has been opened: https://github.com/hamzahamidi/Angular6-json-schema-form/issues/144


> * Existing ISBN are displayed in the detailed view, but if I edit a record, it disappears from the detailed view (same if I create a new record).

Fixed.

> * When adding a second "language", the key "type" (hidden in the form) is not added and the form cannot be validated and saved.

Fixed.

> 
> Improvements:
> 
> * When adding a subfield with the "plus" or when adding a field listed in the top section, place the mouse cursor directly in the new input box.
To be done.
> * When adding a field listed in the top section, display the form so that the newly added field is visible.
To be done.

> * In the top section (add fields), list all kind of fields that can be added, even those already in the form that are repetitive.
To be done.

> * Some constraints should be added for the field "identifiedBy":
>   
>   * "source" exists only if type=bf:Local and is in this case required

Done but required part (not possible with the actual version of jsonschema as I know).

>   * "acquisitionsTerms" exists only if type=bf:Isbn
Done.

Issue #225, opened 2019-03-18
Display of "My account"


pronguen:
In some pages (like admin/records/patron_types), my account is displayed "My account" instead of the initial letters of the users.
The initial letters should always be displayed when a user is authenticated.

pronguen:
@iGormilhit @Garfield-fr 
The behaviour is unstable. After refreshing the cash, I observed that the menu is sometimes displayed as "My account", sometimes as initial letters "AM".
To reproduce this case
1. go to ils.test.rero.ch
2. login as reroilstest@gmail.com
3. change the page and the language several times and observe the display of "My account"

PR #476, opened 2019-09-04
fix user initials view

rerowep:
* Fixes user initials not displayed correctly.

How to test:
Try `dev` and this fix with following command running:
`watch -n 0.1 curl -X OPTIONS localhost:5000`


BadrAly:
Add `Closes #225` to the commit message.

PR #494, opened 2019-09-11
Publication statement view

AoNoOokami:
Test with ISBN 9785852550378


rero_ils/modules/documents/templates/rero_ils/detailed_view_documents.html
```diff
@@ -62,6 +62,13 @@ <h1 class="col-sm-10 col-md-11">{{ record.title }}</h1>
                  else record.publicationYear) }}
             {% endif %}
 
+            <!-- PUBLICATION STATEMENT -->
+            {% if record.provisionActivity %}
+              {% for record in record.provisionActivity %} 
+                {{ dl_value(record.type[3:], record.statement) }}
```
rerowep: Maybe we have to translate `record.type`

PR #496, opened 2019-09-12
circulation : fix issue two loans instead of one

### LINKED ISSUE #484: Two loans instead of one
When a patron requests then checkouts out an item, the system creates two loans instead of one

1. A patron places a request on an item to pickup at an external library, system creates a loan (pid 1 for example) with a state `PENDING`
2. A librarian sends the item to the pickup library, the loan state become `IN_TRANSIT_FOR_PICKUP`
3. A librarian receives the item at the destination library, the loan state becomes `ITEM_AT_DESK`
4. A librarian checks out the item to a patron, the loan state becomes `CANCELLED` and a new loan (pid=2) is created with state `ON_LOAN`

**Expected behavior**
System should continue to work on the first loan and never creates a new loan.

benerken:
* Closes #484
* Moves prior_checkout_actions() function from api_views.py to api.py

Co-Authored-by: Benoit Erken <erken.benoit@gmail.com>

### DIFF
--- a/rero_ils/modules/items/api.py
+++ b/rero_ils/modules/items/api.py
@@ -297,7 +297,6 @@ class Item(IlsRecord):
         for loan in self.get_requests():
             data.setdefault('pending_loans',
                             []).append(loan.dumps_for_circulation())
-
         return data
 
     @classmethod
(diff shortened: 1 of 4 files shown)


rero_ils/modules/items/api.py
```diff
@@ -772,11 +771,37 @@ def checkin(self, current_loan, **kwargs):
...
     @add_loans_parameters_and_flush_indexes
     def checkout(self, current_loan, **kwargs):
         """Checkout item to the user."""
+        new_data = self.prior_checkout_actions(kwargs)
+        loan = Loan.get_record_by_pid(new_data.get('pid'))
+
+        if loan:
+            current_loan = loan
+        else:
+            current_loan = Loan.create(new_data, dbcommit=True, reindex=True)
```
jma: you can replace this by:
```python
current_loan = loan or Loan.create(new_data, dbcommit=True, reindex=True)
```

rero_ils/modules/items/api.py
```diff
@@ -772,11 +771,37 @@ def checkin(self, current_loan, **kwargs):
             item, validate_action = self.validate_request(**request)
         return self, actions
 
+    def prior_checkout_actions(self, data):
```
jma: data can be renamed as action_params if you want

rero_ils/modules/items/api.py
benerken: Done !

rero_ils/modules/items/api.py
benerken: Done.

PR #497, opened 2019-09-12
#1019 - Refactoring units testing api calls

blankoworld:
* create a `post_dict_to(client, entrypoint, data, headers)` function to replace all `client.post()` functions in tests
* replace all `client.post()` calls in tests

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -21,6 +21,7 @@
 import json
 
 import mock
+from flask import url_for
 from invenio_accounts.testutils import login_user_via_session, \
     login_user_via_view
 from invenio_circulation.api import get_loan_for_item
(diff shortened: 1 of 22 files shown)


tests/utils.py
```diff
@@ -63,6 +64,29 @@ def get_json(response):
     return json.loads(response.get_data(as_text=True))
 
 
+def post_dict_to(client, endpoint, input_dict, headers=None):
```
jma: as in most tests we used the same header (`json_header`) this can be the default for the headers parameters

jma: the function name seems a bit cryptic to me, I would also change input_dict as request_body, data_body or similar.

blankoworld: `post_dict_to()` name could be:
* post_request() / post_request_to()
* posting() / posting_to()
* post() / post_to()
* postdata() / post_data() / post_data_to()

I prefer "postdata".

Ok for `json_header`. What about just `data` instead of `input_dict` ?

jma: perfekt

PR #505, opened 2019-09-16
circ_policies ui: increase API size limit

### LINKED ISSUE #405: Increase size of result set during API calls
Release 0.2.3

We have created a test instance with : 
3 organisations
22 libraries
75 locations
9 patron_types
37 item_types

In the admin screen for "circulation policies", we can choose the organisation/libraries, the patron types and the item_types but by default, the software displays ONLY the first 10 entries.
(So we are unable to configure the circ policies for all libraries, item_types or patron types.)

Capture du 2019-07-17 15-24-55

For now, it calls the APIs like this : 

Capture du 2019-07-17 15-30-28

You need to increase the "default size" of APIs result sets when you get the list of libraries, items types etc...
or you could add a "size" parameter when you call the APIs
Ex:  /api/item_types/    =>   /api/item_types/?size=100
/api/locations/    =>   /api/locations/?size=100

Hope it helps,
Ben & the team :-)

zannkukai:
* Closes #405

Increase the default elasticsearch size result. 10000 is the maximum size for the default search type.

### DIFF
--- a/ui/src/app/core/item-type/item-type.service.ts
+++ b/ui/src/app/core/item-type/item-type.service.ts
@@ -18,14 +18,16 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { Injectable } from '@angular/core';
-import { HttpClient, HttpHeaders } from '@angular/common/http';
+import { HttpClient, HttpHeaders, HttpParams } from '@angular/common/http';
 import { ApiService } from '../api/api.service';
 
 const httpOptions = {
   headers: new HttpHeaders({
     'Accept': 'application/json',
     'Content-Type': 'application/json'
-  })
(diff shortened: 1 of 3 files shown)


sebastiendeleze:
Hi, this part will be refactoring in the scope of the new angular library. Custom services for resources won't exist anymore. There will be only one service for managing API calls and you'll be able to call a method like "getRecords" which accepts the size as parameter.

For example, instead of : 

```typescript
this.libraryService.libraries();
```

You will do: 
```typescript
this.recordService.getRecords('libraries', [query], [page], [size]);
```

I think so this change is maybe not relevant.

zannkukai:
Il will continue this bug fix after Angular refactoring.

PR #506, opened 2019-09-17
#1036 - bootstrap: delete useless virtualenv

blankoworld:
This line: https://github.com/rero/rero-ils/blob/f601159395c99757ce2d2c147c560fe334b45526/scripts/bootstrap#L78
creates some problem: a useless virtualenv is created.
But using `npm install` without `pipenv` will use an older version of
npm.
To avoid this problem, we should inform pipenv which environment it
should use.

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -74,9 +74,7 @@ cd ui; pipenv run yarn install --frozen-lockfile; pipenv run yarn run deploy
 cd ${CWD}
 pipenv run invenio npm
 static_folder=$(pipenv run invenio shell --no-term-title -c "print('static_folder:%s' % app.static_folder)"|grep static_folder| cut -d: -f2-)
-cd $static_folder
-pipenv run npm install
-cd ${CWD}
+pipenv run npm install --prefix "${static_folder}"
 
 # build the web assets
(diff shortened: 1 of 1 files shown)


scripts/bootstrap
```diff
@@ -66,17 +66,13 @@ virtualenv_path=`pipenv --venv`
...
 cd ui; pipenv run npm ci; pipenv run npm run deploy
 
 # install the npm dependencies
-cd ${CWD}
 pipenv run invenio npm
 static_folder=$(pipenv run invenio shell --no-term-title -c "print('static_folder:%s' % app.static_folder)"|grep static_folder| cut -d: -f2-)
-cd $static_folder
-pipenv run npm install
-cd ${CWD}
+pipenv run npm install $static_folder -g --prefix "${virtualenv_path}"
```
jma: -g seem not required

blankoworld: I delete the `-g` option.
I understand the impact for a local OS, between all users and local user. But what about a pipenv environment? Does it have an impact in such environment?

jma: The -g make sense for command line tools such as angular node-sass etc. not for libraries.

PR #509, opened 2019-09-17
circ_policies ui: increase API size limit

### LINKED ISSUE #405: Increase size of result set during API calls

zannkukai:
* Closes #405


ui/src/app/records/custom-editor/circulation-settings/circulation-policy.service.ts
```diff
@@ -22,56 +22,38 @@ import { Injectable } from '@angular/core';
...
-      httpOptions
-    );
-  }
-
   loadAllItemTypesPatronTypesCirculationPolicies() {
     return forkJoin(
-      this.itemTypeService.itemTypes(),
-      this.patronTypeService.patronTypes(),
-      this.loadAllCirculationPolicy()
+      this.recordsService.getRecords('global', 'item_types', 1, 9999),
```
jma: these 9999 can be replaced by a variable or better in a configuration service (can be recordsService for the moment).

zannkukai: Added 2 constants into `RecordServices` class : DEFAULT_REST_RESULTS_SIZE, MAX_REST_RESULTS_SIZE.

PR #512, opened 2019-09-18
circulation : fix checkin of item that should go in transit

### LINKED ISSUE #462: Checkin of item that should go in transit
Librarian of library A does a checkin of a borrowed item:
* Expected result: the item goes in transit, as it belongs to library B
* Effective behaviour: the item is on shelf and avoids the transit status.

Tested on RERO ILS v0.3.1

benerken:
* Fixes an existing issue, may closes #462
    
Co-Authored-by: Benoit Erken <erken.benoit@gmail.com>

### DIFF
--- a/rero_ils/static/scss/rero_ils/frontpage.scss
+++ b/rero_ils/static/scss/rero_ils/frontpage.scss
@@ -18,7 +18,6 @@
  */
 
  .rero-ils-frontpage {
-  height: 45vh;
   background-image: url(#{$static-images}/homepage_image_bleu.jpg);
   background-size: cover;
   background-position: center;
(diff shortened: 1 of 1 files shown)


rero_ils/modules/items/api.py
```diff
@@ -711,8 +711,7 @@ def item_location_retriever(cls, item_pid, **kwargs):
         #       to deal with multiple pickup locations for a library
         item = cls.get_record_by_pid(item_pid)
         if item:
-            library = item.get_library_of_last_location()
-            return library.get_pickup_location_pid()
+            return item.get_owning_pickup_location_pid()
```
BadrAly: I recall that the `PO` asked to not rely on the the `pickup location` of the `owning library`. Instead he asked to find where the item is actually now by checking the `library` of the last transaction `loan`. This is a use case not documented in the Circulation state chart. The outcome, keep the two lines:
```
library = item.get_library_of_last_location()
return library.get_pickup_location_pid()
```

PR #523, opened 2019-09-24
#1027 - item availability instead of status

blankoworld:
Item detail view contains status, as last field. But no info appears
about the item availability.

The availability appears in other view as a colored circle. Green circle
while available. Red if not available.

This change adds the colored circle before the item status. It also
changes the field description to *Availability*.

### DIFF
--- a/rero_ils/modules/items/templates/rero_ils/detailed_view_items.html
+++ b/rero_ils/modules/items/templates/rero_ils/detailed_view_items.html
@@ -50,9 +50,13 @@
             _('Library'),
             library.name)
           }}
-          {% if record.status %}
-            {{ dl(_('Status'), _(record.status)) }}
-          {% endif %}
+          {{ dl(
+              _('Availability'),
+              '<i class="fa fa-circle text-{}"></i> {}'.format(
+                'success' if record.available else 'danger',
+                _(record|item_availability_text)
+              )
+          )}}
         </dl>
       </article>
       {% if current_user | can_access_item(record) %}


rero_ils/modules/items/templates/rero_ils/detailed_view_items.html
```diff
@@ -51,7 +55,10 @@ <h1 class="col-sm-11 col-md-11">{{ _('Barcode') }} {{ record.barcode }}</h1>
             library.name)
           }}
           {% if record.status %}
-            {{ dl(_('Status'), _(record.status)) }}
+            {{ dl(
+              _('Availability'),
+              '<i class="fa fa-circle text-' + availability + '"></i> ' + record|item_availability_text)
```
BadrAly: you may remove lines 37-40 and  replace  `availability` by :
```
{{ 'success' if record.available else 'danger' }}
```

blankoworld: @BadrAly I did it L.37 because L.60 didn't accept `{{` and `}}` and without them it didn't use the end of the line which break the page.

BadrAly: ok

blankoworld:
> Considering the commit message, it may be too verbose:
> 
> ```
> ui: replace status by availability in item detailed view
> 
> Item detail view should contains availability instead of status.
> ```

@iGormilhit : I don't use "detailed" word as we limit the length to 50 chars.
Is that OK like this:

> ui: replace status by availability in item view

?

iGormilhit:
@blankoworld Ok for me. Thanks.

rero_ils/modules/items/templates/rero_ils/detailed_view_items.html
```diff
@@ -34,6 +34,10 @@ <h1 class="col-sm-11 col-md-11">{{ _('Barcode') }} {{ record.barcode }}</h1>
   <section>
     <!-- Item data -->
     <section class="py-4">
+      {% set availability = 'danger' %}
```
jma: this is not necessary

PR #524, opened 2019-09-25
#971 - ui: display git commit hash on frontpage

blankoworld:
Homepage displays current RERO-ILS version. Version as in "release".
It's not so accurate.
To get more precision we use git hash version. We so:

* fetch GIT_HASH environment variable for Docker containers
* if no GIT_HASH environment variable we read files in .git directory

Info : it's not mandatory to have the complete hash. Just 8 first chars is enough.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,12 @@ RUN chown -R invenio:invenio ${WORKING_DIR}
 
 USER 1000
 
+# Set last GIT commit hash. Example:
+# docker build --build-arg GIT_HASH=$(git rev-parse @) .
+ARG GIT_HASH
+# Last GIT commit hash
+ENV INVENIO_RERO_ILS_APP_GIT_HASH ${GIT_HASH:-''}
+
 ENV INVENIO_COLLECT_STORAGE='flask_collect.storage.file'
 RUN ./scripts/bootstrap --deploy && rm -fr ui/node_modules
 
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -1015,7 +1015,10 @@ RERO_ILS_APP_BASE_URL = 'https://ils.rero.ch'
 RERO_ILS_PERMALINK_RERO_URL = 'http://data.rero.ch/01-{identifier}'
 RERO_ILS_PERMALINK_BNF_URL = 'http://catalogue.bnf.fr/ark:/12148/{identifier}'
 
-#: RERO_ILS MEF specificconfigurations.
+#: Git commit hash. If set, a link to github commit page is displayed on RERO-ILS frontpage.
+RERO_ILS_APP_GIT_HASH = None
+
+#: RERO_ILS MEF specific configurations.
 RERO_ILS_MEF_HOST = 'mef.rero.ch'
 RERO_ILS_MEF_URL = 'https://{host}/api/mef/'.format(host=RERO_ILS_MEF_HOST)
 RERO_ILS_MEF_RESULT_SIZE = 100
--- a/rero_ils/templates/rero_ils/frontpage.html
+++ b/rero_ils/templates/rero_ils/frontpage.html
@@ -92,7 +92,11 @@
           <p class="card-text pl-2 align-self-center">
             {{_('Software source code')}}
             <a href="https://github.com/rero/rero-ils" class="rero-ils-external-link" title="{{ _('Link to the RERO ILS code source on GitHub') }}">{{_('on GitHub')}}</a>, {{ _('version') }}
-            <a href="https://github.com/rero/rero-ils/releases/tag/v{{ version }}" class="rero-ils-external-link" title="{{ _('Link to the RERO ILS code source on GitHub, last released version') }}">{{ version }}</a>
+            {% if config.RERO_ILS_APP_GIT_HASH %}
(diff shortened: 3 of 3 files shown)


rero_ils/utils.py
```diff
@@ -80,3 +81,63 @@ def get_agg_config(index_name, field):
             current_app.config.get('RERO_ILS_DEFAULT_AGGREGATION_SIZE')
         )
     ))
+
+
+def git_get_current_branch():
```
jma: I'm not sure that is robust. I will probably do call a `git` command directly or use a python package to deal with git. Something like:
```python
import subprocess
label = subprocess.check_output(["git", "describe"]).strip()
```

rero_ils/utils.py
```diff
@@ -80,3 +81,63 @@ def get_agg_config(index_name, field):
...
+    """Get current GIT branch or None.
+
+    Use only files to determine in which branch we are.
+
+    :return: string of current GIT branch or None
+    """
+    res = None
```
jma: you can remove this

rero_ils/utils.py
```diff
@@ -80,3 +81,63 @@ def get_agg_config(index_name, field):
...
+    current_dir = os.path.dirname(__file__)
+    filepath = os.path.join(current_dir, '../.git/HEAD')
+    if os.path.exists(filepath):
+        with open(filepath, 'r') as headfile:
+            res = headfile.read().strip().split('/')[-1]
```
jma: replace directly by a `return`

rero_ils/utils.py
```diff
@@ -80,3 +81,63 @@ def get_agg_config(index_name, field):
...
+
+
+def git_get_last_commit_hash():
+    """Get last GIT commit hash.
+
+    Use only files to determine what's the last commit hash.
+
+    :return: string of last GIT commit hash or None
+    """
+    res = None
```
jma: same as preview

rero_ils/utils.py
```diff
@@ -80,3 +81,63 @@ def get_agg_config(index_name, field):
...
+    :return: string of last GIT commit hash, 8 first chars. Or None.
+    """
+    res = ''
+    # Attempt 1
+    _git_hash = os.environ.get("GIT_HASH", default=None)
+    if _git_hash:
+        return _git_hash[:8]
+    # Attempt 2
+    _hash = git_get_last_commit_hash()
+    if _hash is not None:
```
jma: can be replaced by:
```python
if _hash:
```

rero_ils/utils.py
```diff
@@ -80,3 +81,63 @@ def get_agg_config(index_name, field):
...
+        - master if production
+        - dev if development
+        - only 8 first chars
+    3. If no file: return empty string.
+
+    :return: string of last GIT commit hash, 8 first chars. Or None.
+    """
+    res = ''
+    # Attempt 1
+    _git_hash = os.environ.get("GIT_HASH", default=None)
```
jma: please do not use `_` except if it is a python key word or private property

jma: I propose to not use environment variable but flask config instead: `RERO_ILS_APP_GIT_HASH` or `RERO_ILS_APP_GIT_VERSION`. This will avoid variable injection in the template file as config is available in templates. It is also more flexible as we can use env variable or config file.

rero_ils/views.py
```diff
@@ -207,6 +208,7 @@ def index():
     """Home Page."""
     return render_template('rero_ils/frontpage.html',
                            version=__version__,
+                           git_version=git_short_hash(),
```
jma: probably useless

Dockerfile
```diff
@@ -28,6 +28,9 @@ RUN chown -R invenio:invenio ${WORKING_DIR}
 
 USER 1000
 
+# Get last GIT commit hash
+ENV RERO_ILS_APP_GIT_HASH $(git rev-parse @)
```
jma: you should prefix the env with `INVENIO_`

rero_ils/config.py
```diff
@@ -982,7 +983,12 @@ def _(x):
 RERO_ILS_PERMALINK_RERO_URL = 'http://data.rero.ch/01-{identifier}'
 RERO_ILS_PERMALINK_BNF_URL = 'http://catalogue.bnf.fr/ark:/12148/{identifier}'
 
-#: RERO_ILS MEF specificconfigurations.
+#: RERO_ILS Git displayed hash
+RERO_ILS_APP_GIT_HASH = None
+if os.getenv('RERO_ILS_APP_GIT_HASH', False):
```
jma: This is useless. See https://invenio-app.readthedocs.io/en/latest/usage.html#instance-configuration for more details.

Dockerfile
```diff
@@ -28,6 +28,9 @@ RUN chown -R invenio:invenio ${WORKING_DIR}
 
 USER 1000
 
+# Get last GIT commit hash
+ENV INVENIO_RERO_ILS_APP_GIT_HASH $(git rev-parse @)
```
jma: This will not work as the .git does not exists on the Docker image. It would be better to use ARG: Set build-time variables (--build-arg) section in https://docs.docker.com/engine/reference/commandline/build. This is also important to have the choice to set this variable as we do not want to do it for production!

You can try by yourself by building the docker image and run a command inside:
```bash
docker build --rm -t rero/rero-ils:test .
docker run --rm -ti rero/rero-ils:test git rev-parse @
or
docker run --rm -ti rero/rero-ils:test /bin/bash
and
echo $INVENIO_RERO_ILS_APP_GIT_HASH
```
The last command return `$(git rev-parse @)` which is not correct.

PR #536, opened 2019-09-30
#1021 - refactoring: delete unused imports

blankoworld:
More and more imports are useless in the project.

To avoid this problem, 2 things are needed:

* delete current imports
* check imports regularly

To delete current imports it uses a program: `autoflakes`

```bash
autoflake --remove-all-unused-imports -i -r --exclude ui \
--ignore-init-module-imports .
```

Also checks that no more useless imports remains. It also uses
`autoflakes` during `run-tests.sh` script.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -46,6 +46,7 @@ invenio-access = "<1.2.0"
 [dev-packages]
 Flask-Debugtoolbar = ">=0.10.1"
 Sphinx = ">=1.5.1"
+autoflake = ">=1.3.1"
 check-manifest = ">=0.35"
 coverage = ">=4.4.1"
 isort = ">=4.3.5,<4.3.10"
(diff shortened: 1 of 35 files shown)


blankoworld:
> Are the celery worker tasks: ` indexer, accounts, ebooks-harvester, notification-creation` still functional and executed in a celery worker?

That's a good point. Have I any way to check this quickly @rerowep ?

rerowep:
> > Are the celery worker tasks: ` indexer, accounts, ebooks-harvester, notification-creation` still functional and executed in a celery worker?
> 
> That's a good point. Have I any way to check this quickly @rerowep ?

* `bootstrap`
* start server with celery workers `pipenv run celery worker -A invenio_app.celery --autoscale=8,1 --beat -l WARNING --without-heartbeat & pid_celery=$!`
* `setup`
* control if you see errors in the server log after some time

PR #558, opened 2019-10-09
tests: add PID verifications with commit/rollback

blankoworld:
Record that failed validation generates a PID. It should not.

If any validation error occurs during a Record validation,
the PID should be canceled in database (using a rollback).

Some tests have been done to check this.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/tests/ui/test_api.py
+++ b/tests/ui/test_api.py
@@ -27,11 +27,13 @@ from invenio_pidstore.models import PIDStatus, RecordIdentifier
 from invenio_pidstore.providers.base import BaseProvider
 from invenio_search import current_search
 from invenio_search.api import RecordsSearch
-from utils import flush_index
+from jsonschema.exceptions import ValidationError
 
 from rero_ils.modules.api import IlsRecord, IlsRecordError
 from rero_ils.modules.fetchers import id_fetcher
 from rero_ils.modules.minters import id_minter
+from rero_ils.modules.providers import Provider
+from utils import flush_index
 
 
 class IdentifierTest(RecordIdentifier):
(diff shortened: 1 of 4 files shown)


rero_ils/modules/api.py
```diff
@@ -137,7 +138,12 @@ def create(cls, data, id_=None, delete_pid=False,
         if not id_:
             id_ = uuid4()
         cls.minter(id_, data)
-        record = super(IlsRecord, cls).create(data=data, id_=id_, **kwargs)
+        # We don't need to keep the PID if any validation error
+        try:
+            record = super(IlsRecord, cls).create(data=data, id_=id_, **kwargs)
+        except (RecordValidationError, ValidationError) as e:
+            db.session.rollback()
```
rerowep: I'am not sure this is atomic. I think we have to use `with db.session.begin_nested():`

rero_ils/modules/api.py
```diff
@@ -137,7 +138,12 @@ def create(cls, data, id_=None, delete_pid=False,
         if not id_:
             id_ = uuid4()
         cls.minter(id_, data)
-        record = super(IlsRecord, cls).create(data=data, id_=id_, **kwargs)
+        # We don't need to keep the PID if any validation error
+        try:
+            record = super(IlsRecord, cls).create(data=data, id_=id_, **kwargs)
+        except (RecordValidationError, ValidationError) as e:
```
rerowep: One letter variables `e` are not recommended.

rero_ils/modules/api.py
```diff
@@ -137,7 +138,12 @@ def create(cls, data, id_=None, delete_pid=False,
...
+            db.session.rollback()
+            raise e
```
rerowep: Maybe `raise` will interrupt the import of fixtures. We have to be clear if we like to interrupt the import or if we have to intercept the exception and display a error message for example.

rero_ils/modules/api.py
blankoworld: I think so too. But then I check https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L172-L190 which already uses `db.session.begin_nested()`.

That's why I just use `db.session.rollback()`.

rero_ils/modules/api.py
blankoworld: Which fixtures do you talk about? Those from `./scripts/setup`?

I will check that.

blankoworld: On my side, it seems to be OK.

tests/ui/test_api.py
```diff
@@ -201,3 +203,64 @@ def test_ilsrecord(app, es_default_index, ils_record, ils_record_2):
...
+FailedPidProvider = type(
+    'FailedPidProvider',
+    (Provider,),
+    dict(
+        identifier=FailedPidIdentifier,
+        pid_type='failed'
+    )
+)
+
+# cipo minter
```
rerowep: What is `cipo`

rero_ils/modules/api.py
blankoworld: Minter add the pid to the session here: https://github.com/inveniosoftware/invenio-pidstore/blob/master/invenio_pidstore/models.py#L562
The `with` statement should commit at the end. So PID should be reserved.

But I didn't notice that the PID remains after the rollback yet. I should investigate more.

rero_ils/modules/api.py
blankoworld: I change it to `validation_error`

tests/ui/test_api.py
blankoworld: Its comes from **ci**rculation **po**licies. My bad!

PR #579, opened 2019-10-23
cli: reserve a range of pids

BadrAly:
* Assigns all unused pids the status NEW.
* Assigns all reserved pids the status RESERVED.
* Improves method to insert identifer pids and reset sequence.

## How to test?

-  script/setup
- make sure sequences are set correctly (pgAdmin)

- for information how to use the new cli:
pipenv run invenio utils reserve_pid_range --help

- test the cli and reserve a range of pids and consult the console to check the status of the reserved pids.

- an example to reserve 10 pids for patron future loading and assign status NEW to all unused pids:
`pipenv run invenio utils reserve_pid_range -p ptrn -n 10 -u`

## Code review check list

- [x] File names.
- [x] Functions names.
- [x] Functions docstrings.

### DIFF
--- a/rero_ils/modules/cli.py
+++ b/rero_ils/modules/cli.py
@@ -38,7 +38,8 @@ from flask import current_app
 from flask.cli import with_appcontext
 from flask_security.confirmable import confirm_user
 from invenio_accounts.cli import commit, users
-from invenio_pidstore.models import PersistentIdentifier
+from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 from invenio_records.api import Record
 from invenio_records_rest.utils import obj_or_import_string
 from invenio_search.cli import es_version_check
@@ -299,9 +300,9 @@ def create(infile, pid_type, schema, verbose, dbcommit, reindex, append):
             error_file.write(']')
 
     if append:
-        table = record_class.provider.identifier
+        identifier = record_class.provider.identifier
         try:
-            append_fixtures_new_identifiers(table, sorted(pids))
+            append_fixtures_new_identifiers(identifier, sorted(pids), pid_type)
         except Exception as err:
             pass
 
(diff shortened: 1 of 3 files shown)


rero_ils/modules/cli.py
```diff
@@ -475,3 +477,57 @@ def compile_json(src_jsonfile, output, verbose):
...
+        ))
+    if unused:
+        for pid in range(1, identifier.max()):
+            if not db.session.query(
+                    identifier.query.filter(identifier.recid == pid).exists()
+            ).scalar():
+                record_class.provider.create(pid_type, pid_value=pid,
+                                             status=PIDStatus.NEW)
+                db.session.add(identifier(recid=pid))
+            db.session.commit()
```
zannkukai: Despite I tried to used "-u" option, all requested pid are set to RESERVED into PostgreSQL.
I tried with multiple record_type, multiple record_number.
My Invenio knowledge is not suffisant to determine if it's a bug or not.... But for my tests the `if` condition is never satisfied.

zannkukai (approved):

rero_ils/modules/cli.py
BadrAly: If you have the following scenario in db:
1. REGISTERED pids 1,2,3,7,8,9
2. you execute the cli to reserve 10 pids, with the option `-u`
3. here is what you get:
3.1: REGISTERED pids 1,2,3,7,8,9 [pid_status in table = 'R']
3.2: NEW pids: 4,5,6, [pid_status in table = 'N']
3.3. RESERVED pids: 10->19  [pid_status in table = 'K']

so the option `u` is only to fill the gaps.

rero_ils/modules/cli.py
```diff
@@ -475,3 +477,57 @@ def compile_json(src_jsonfile, output, verbose):
...
+@utils.command('reserve_pid_range')
+@click.option('-p', '--pid_type', 'pid_type', default=None,
+              help='pid type of the resource')
+@click.option('-n', '--records_number', 'records_number', default=None,
+              help='Number of records to load')
+@click.option('-u', '--unused', 'unused', is_flag=True, default=False,
+              help='Set unused pids status to NEW ')
+@with_appcontext
+def reserve_pid_range(pid_type, records_number, unused):
+    """Reserve a range of pids for eventual records loading.
```
iGormilhit: Is `eventual` correct here?

iGormilhit (changes requested):
Commit message: title with a verb. `cli: reserve a range of pids` ? 

done

rero_ils/modules/cli.py
BadrAly: I used `future` instead.

jma (approved):

PR #614, opened 2019-11-11
Workshop Issues Fixing

jma:

- To test the issue fixing by the Workshop presenter.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


pronguen:
To improve:
1. Facets value: when clicking on button "More", display 30 values instead of only 10. Make this improvement on admin and public views ideally, only on public in any case.
2. RERO ILS seems not to detect the default policy. It applies a default loan duration of 30 days, even if the default policy is set to 5. It does not allow renewals.

jma:
> To improve:
> 
> 1. Facets value: when clicking on button "More", display 30 values instead of only 10. Make this improvement on admin and public views ideally, only on public in any case.
> 2. RERO ILS seems not to detect the default policy. It applies a default loan duration of 30 days, even if the default policy is set to 5. It does not allow renewals.

This two issues should be solved.

pronguen:
To correct:
* is is possible to save a patron without selection a communication channel and language. This two fields are required, and it created bugs later if they are not filled.

PR #630, opened 2019-11-17
api: fix rest api person when a single record is requested

lauren-d:
* Implements custom routing converter defining 'espid' type.
* Implements new elasticsearch record class (detached from the database).
* Clean residual MEF api harvesting.
* Change title of 'pid' and 'id' in mef person jsonschema.
* Refactoring of MEF person api.
* Rewrite fixtures and tests for mef persons.
* Improve code coverage for mef persons module.

- Implements task: https://tree.taiga.io/project/rero21-reroils/task/1120

## How to test?

- bootstrap and setup
- go to ~/api/persons/21414859

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


rero_ils/modules/api.py
```diff
@@ -319,3 +326,124 @@ def organisation_pid(self):
         if self.get('organisation'):
             return self.replace_refs()['organisation']['pid']
         return None
+
+
+class ElasticsearchRecord(dict):
```
rerowep: Why not using `invenio_record.api.RecordBase` as `class` and only surcharge the changed functionality?

lauren-d: RecordBase should be use with SQLAchemy model. In this case, the idea is to have a Record class not linked to the database. In additon, if we want use RecordBase, all properties or methods must be surcharged. If is better, I will use RecordBase instead of dict

jma:
In your commit message: detached of -> detached from

jma:
Very interesting approach! Some suggestions:
- the creation date and update date can be retrieve from the es index and updated when the record is indexed for the first time (create + index), when the commit and patch method is called
- here ES can be considered as a replacement of the DB, in such case I will implement create, commit, patch, with ES transactions and I will do nothing in the index method
- to be exhaustives some Record methods still missing: patch, get_records, etc. See https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py for more details
- version_id can be retrieve by ES itself: https://www.elastic.co/fr/blog/elasticsearch-versioning-support

These suggestions can take while to implement, we can start with this implementation to do it later.

jma:
This is done in an other PR and the MEF becomes a regular resource.

PR #644, opened 2019-11-26
circulation policies: fix settings bug

Garfield-fr:


AoNoOokami:
It doesn't fix the bug. I created an item type and a patron type, I tried to create a ci-po with those settings. The patron type is ok, but not the item type, whose pid is always 1 (instead of 3 in my test).

PR #664, opened 2019-12-10
Fix CSP problem in Chromium for images

blankoworld:
* Adds new CSP (Content Security Policy) for img-src

I get some errors on Chromium for https://ils.test.rero.ch:

RERO_ILS_img-src_problems

## How to test?

1. Launch rero-ils server
1. Go to https://localhost:5000/
1. Press F12 on Chrome/Chromium
1. No errors should appears about `img-src` problems

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -288,7 +288,8 @@ APP_DEFAULT_SECURE_HEADERS = {
             'https://www.google-analytics.com',
             'https://services.test.rero.ch',
             'https://services.rero.ch'
-        ]
+        ],
+        'img-src': ['data:', "'self'"]
     },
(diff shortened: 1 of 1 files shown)


blankoworld:
As explained on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src#common_sources) it seems that it's a little bit unsafe to accept `data:` sources in CSP.

I so check more my browser configuration. It seems that this `data:` item comes from **Ghostery** plugins (either on Firefox or Chrom[e|ium]).

Except if you have another example, I closed this pull request.

PR #677, opened 2019-12-18
docker: reduce image size of 500MB

blankoworld:
* deletes Dockerfile.base
* merges Dockerfile.base and Dockerfile
* improves Dockerfile layer to reduce size of 500MB
* uses Debian buster as default image

To reduce docker images size for deployment.

## How to test?

```
docker build -t rero/rero-ils:slim .
docker images|grep rero
```

Then compare size of different rero-ils images (rero/rero-ils:latest and rero/rero-ils:slim for an example).

Finally, run `docker-compose -f docker-compose.full.yml up -d` to check

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


Dockerfile
```diff
@@ -15,25 +15,60 @@
...
+
+# for some files to be written (example: egg-info)
+RUN chown -R invenio:invenio ${WORKING_DIR} \
+  && chmod -R go+w ${WORKING_DIR}
+
+# require debian packages
+# Set locales to C.UTF-8 to avoid some problem:
+# Cf. https://click.palletsprojects.com/en/5.x/python3/#python-3-surrogate-handling
+RUN set -ex \
+  && buildDeps=' \
```
rerowep: && doing the same as `set -e` maybe we can use ; instead

blankoworld: This is done.

blankoworld:
@rerowep : Could you check changes in `docker` directory regarding your production setup? This is to avoid any break in your production with these changes.

docker/nginx/conf.d/default.conf
```diff
@@ -22,10 +22,10 @@
 # which is expected to be running on port 5000/5001.
 
 upstream ui_server {
-  server localhost:5000 fail_timeout=0;
+  server web-ui:5000 fail_timeout=0;
```
rerowep: It is not working on deployment:
`nginx: [emerg] host not found in upstream "web-ui:5000" in /etc/nginx/conf.d/default.conf:25`

blankoworld: You probably have a different name for the ui in your production `docker-compose.yml` file.
We have to use the same name for this container. Either you rename it `web-ui` in your production `docker-compose.yml` file or I rename it by those you choose.
What's your preference?

PS: Another solution is to put always same port on localhost (for example 5000) but it differs from Docker best practices…

rerowep: I changed the name, still not working!

rerowep: The port is fixed to 5000 in `docker/nginx/conf.d/default.conf `, so whats the problem ?

blankoworld: The problem is that in Docker you will have 2 ports: those inside the container (here 5000) and the port outside the container (for example 32774).

It's adviced to use `web-api` as link to other containers. And `5000` as the used port.
32774 is not known by our scripts. But `web-api` (keyword for our container) and `5000` (internal port) are.

We should be careful to not use `5000:5000` in docker-compose because it block 5000 port. Which is incoherent with docker usage. Docker is built to permit to launch multiple times a container (with multiple random ports).

We will find a solution Monday ;)

jma: Didi you?

blankoworld: We did!

jma: I do not understand why you have removed the docker-compose.full. I do not think that it reduce the image size and brake the Invenio choice. What is the advantage of doing this?

blankoworld: I do it for 2 reasons:
1. first we talk about the usage in daily meeting and I understand that in RERO-ils you will only use the Dockerfile and files located in **docker** directory for production use. As a result, I consider the `docker-compose.full.yml` file remains for users that want to discover RERO-ils or to launch it in production.
2. every day usage: as a `docker-compose.yml` file existe, you always have collision with the other name `docker-compose.full.yml` in command line. Try it once or twice. More and you become crazy. It's more readable and more understable a `prod.yml` file.

I understand that we break the Invenio choice, but isn't that already done by all changes RERO made in docker directory and all Dockerfiles?
Moreover Invenio can do mistakes, as https://github.com/rero/rero-ils/issues/713 shows it.

blankoworld:
> Why not a new directory `docker.prod` and not a subdirectory `prod`in `docker` ?

@rerowep In Daily Meeting we talk about this, I understood we put all in `docker` directory, even if it could be a mess.
Do you prefer I use a `docker.prod` instead?

> Get following error in UI: `FileNotFoundError: [Errno 2] No such file or directory: '/invenio/var/instance/static/scss/rero_ils/styles.scss'``

If you encount this error, you're not in production use. A production environment shouln't compile files. SCSS files are only built as a preparation for production.
I explain it here: https://github.com/rero/rero-ils/issues/713

[Flask itself advices to run in production mode](https://flask.palletsprojects.com/en/1.1.x/tutorial/deploy/#run-with-a-production-server).
But it's not enough: you need to configure Invenio to be in production mode.

I use these variable for that:

```ini
INVENIO_ENV='production'
INVENIO_SESSION_COOKIE_SECURE=True
INVENIO_ASSETS_AUTO_BUILD=''
```
The `ASSETS_AUTO_BUILD` is very important. Without it, your application will **always** recompile every templates and every CSS. And so it will search for `scss` and `templates` directories…

PR #692, opened 2020-01-10
script: add rero-ils-ui install from tgz

AoNoOokami:
* Updates bootstrap script to install rero-ils-ui from .tgz if needed.

To simplify rero-ils-ui integration from PR.
To allow its deployment.

## How to test?

1. In RERO-ILS-UI
* `npm run pack`
* Copy path/to/rero-rero-ils-ui-0.0.x.tgz

2. In RERO-ILS
* Go to static folder of invenio instance
* `npm install @rero/rero-ils-ui@latest`
* OPTIONAL: `cat package.json`
* OPTIONAL: You should see: `"@rero/rero-ils-ui": "0.0.8",` (or other latest version)
* Go back to rero-ils project folder
* `./scripts/bootstrap -t path/to/rero-rero-ils-ui-0.0.x.tgz`
* Go to static folder of invenio instance (again)
* `cat package.json`
* You should see: `"@rero/rero-ils-ui": "file:../../../rero-ils-ui/rero-rero-ils-ui-0.0.x.tgz",`

3. Check also `./scripts/bootstrap --deploy` and `./scripts/bootstrap --ci` behaviour

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -57,29 +57,41 @@ success_msg() {
 # Displays program name
 msg "PROGRAM: ${PROGRAM}"
 
-# The real bootstrap script starts 20 lines below ;-)
+# options may be followed by one colon to indicate they have a required argument
+if ! options=$(getopt -o dsb -l -t deploy,ci,tgz_path: -- "$@")
+then
+    # something went wrong, getopt will put out an error message for us
+    exit 1
+fi
+
+# The real bootstrap script starts with app installation below ;-)
 flags=()
 deploy=false
 cmd="pipenv sync"
 ci=false
-while test $# -gt 0; do
-    case "$1" in
-        --deploy)
-            # We are probably inside travis, check for env variables
-            deploy=true
-            shift
-            ;;
-        --ci)
-            # We are probably inside travis, check for env variables
-            ci=true
-            shift
-            ;;
-        *)
-            # Use the flags that are passed to the script directly to install
-            flags+=("$1")
-            shift
-            ;;
-    esac
+TGZ_PATH=""
+while test $# -gt 0
+do
+  case "$1" in
+    -d|--deploy)
+      # We are probably inside travis, check for env variables
+      deploy=true ;;
+    -c|--ci)
+      # We are probably inside travis, check for env variables
+      ci=true ;;
+    -t|--tgz_path)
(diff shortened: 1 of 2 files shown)


scripts/bootstrap
```diff
@@ -54,32 +54,39 @@ success_msg() {
...
-            ci=true
-            shift
-            ;;
-        *)
-            # Use the flags that are passed to the script directly to install
-            flags+=("$1")
-            shift
-            ;;
-    esac
+while test $# -gt 0;
```
blankoworld: I think `;` is not needed if you go to the line after. It was only if you put `do` after that.

scripts/bootstrap
```diff
@@ -119,6 +126,11 @@ info_msg "Search static folder location"
 static_folder=$(pipenv run invenio shell --no-term-title -c "print('static_folder:%s' % app.static_folder)"|grep static_folder| cut -d: -f2-)
 info_msg "Install static folder npm dependencies in: ${static_folder}"
 pipenv run npm install --prefix "${static_folder}"
+if [ ! -d $TGZ_PATH ]
```
blankoworld: What about using `-z` test to check if TGZ_PATH is set? And then `-d`?

AoNoOokami: Because of `tgz_path:` at line 61, a value is expected. If not, there will be an error message saying `missing parameter`. So I think it's redundant. I'll check.

AoNoOokami: Ok, no error message saying `missing parameter`. But if no value is given to TGZ_PATH, the condition is considered as `false` and bootstrap script is run normally.

scripts/bootstrap
```diff
@@ -57,29 +57,42 @@ success_msg() {
...
+      # We are probably inside travis, check for env variables
+      deploy=true ;;
+    -c|--ci)
+      # We are probably inside travis, check for env variables
+      ci=true ;;
+    -t|--tgz_path)
+      TGZ_PATH=$2
+      if [ -z $2 ]
+        then
+          error_msg "Please give path to .tgz"
```
blankoworld: Use error_msg+exit instead of "error_msg" and then "exit" ;)

PR #699, opened 2020-01-15
setup: speed up and clean improvements

blankoworld:

- Because of the setup load 54 times the virtual environment (`pipenv run`), we need to reduce it to 1 load.
- To reduce logs from setup (delete all useless warnings)

For that I use this tip from Pipenv: https://pipenv.readthedocs.io/en/latest/advanced/#custom-script-shortcuts .

For Python I also use a "ignore DeprecationWarning" state.

## How to test?

```bash
# First prepare the environment
docker-compose down && pipenv --rm && ./scripts/bootstrap && docker-compose up -d && ./docker/wait-for-services.sh
# Then launch the setup like this:
PIPENV_QUIET=1 pipenv run setup
```

The setup should only display info messages, things about the setup.

Python DeprecationWarnings should have disappeared.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


scripts/setup
```diff
@@ -68,6 +68,9 @@ success_msg() {
 }
 
 
+export FLASK_SKIP_DOTENV=1  # avoid 'There are .env or .flaskenv…' messages
+export PYTHONWARNINGS="ignore::DeprecationWarning"
```
rerowep: Better to make this optional.

Pipfile
```diff
@@ -73,3 +73,4 @@ python_version = "3.6"
 test = "python setup.py test"
 build_sphinx = "python setup.py build_sphinx"
 dev = "pytest --no-cov -vs tests"
+setup = "./scripts/setup"
```
jma: `PIPENV_QUIET=1 ./script/setup` instead or better include `PIPENV_QUIET=1` in the ./scripts/setup?

scripts/setup
blankoworld: I do it via a `--warnings` (or `-w`) option to enable warnings.

Pipfile
blankoworld: `PIPENV_QUIET=1` have an impact on the pipenv command only if used outside the pipenv itself. So before the `pipenv run setup` or before the script IF it includes `pyenv shell` inside.

It is to avoid the *Courtesy note* (Cf. https://github.com/pypa/pipenv/issues/3068). But our pipenv version have a bug on it.

blankoworld: @jma: `pyenv shell` seems to need more preparation before being usable for this.

```
❯ ./scripts/setup
pyenv: shell integration not enabled. Run `pyenv init' for instructions.
❯ pyenv init
# Load pyenv automatically by appending
# the following to ~/.zshrc:

eval "$(pyenv init -)"
```

It remains more comfortable to use the [custom shortcut functionnality of pipenv](https://pipenv.readthedocs.io/en/latest/advanced/#custom-script-shortcuts): 

  * no more preparation for the developer
  * a message explains to developer which command to use

scripts/setup
```diff
@@ -75,12 +75,19 @@ STOP_EXECUTION=true
 CREATE_LAZY=""
 DONT_STOP=""
 PREFIX=""
+ENABLE_WARNINGS=false
```
rerowep: The warnings should be active by default.

blankoworld: I change the behaviour as you expect. Have a look!

PR #700, opened 2020-01-16
deployment: adaptions for rero-ils-ui

rerowep:
* Adds rero-ils-ui hash to frontpage.
* Updates Dockerfile to use of rero-ils-ui package file.

## How to test?

- Frontpage can be tested by changing `RERO_ILS_APP_GIT_HASH` and `RERO_ILS_UI_GIT_HASH` in 
`config.py`
- real deployment can be tested with`minikube` or `docker-desktop kubernetes` and the deployment scripts.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,12 +28,12 @@ RUN chown -R invenio:invenio ${WORKING_DIR}
 
 USER 1000
 
-# Set last GIT commit hash. Example:
-# docker build --build-arg GIT_HASH=$(git rev-parse @) .
 ARG GIT_HASH
-# Last GIT commit hash
 ENV INVENIO_RERO_ILS_APP_GIT_HASH ${GIT_HASH:-''}
+ARG GIT_UI_HASH
+ENV INVENIO_RERO_ILS_UI_GIT_HASH ${GIT_UI_HASH:-''}
+ARG UI_TGZ=""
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -28,12 +28,12 @@ RUN chown -R invenio:invenio ${WORKING_DIR}
...
-# Last GIT commit hash
 ENV INVENIO_RERO_ILS_APP_GIT_HASH ${GIT_HASH:-''}
+ARG GIT_UI_HASH
+ENV INVENIO_RERO_ILS_UI_GIT_HASH ${GIT_UI_HASH:-''}
+ARG UI_TGZ=""
 
 ENV INVENIO_COLLECT_STORAGE='flask_collect.storage.file'
-RUN ./scripts/bootstrap --deploy && rm -fr ui/node_modules
 
+RUN ./scripts/bootstrap --deploy ${UI_TGZ} && rm -fr ui/node_modules
```
blankoworld: Either you miss `-t ${UI_TGZ}` or you should rename the argument to a more self-explained one. Something like `ADDITIONAL_ARGS` or `BOOTSTRAP_ARGS`. What do you think?

rerowep: UI_TGZ is with the -t

blankoworld: So it's a wrong named variable :(

PR #706, opened 2020-01-22
ui: display a different logo/color for each orga.

blankoworld:
* Adds RERO_ILS_THEME_ORGANISATION_CSS_ENDPOINT configuration variable to gives a
way for each organisation to have a specific stylesheet

- It implements [US1200](https://tree.taiga.io/project/rero21-reroils/us/1200?milestone=252915)

## How to test?

1. Launch server
2. Go to each given URL:
  * https://localhost:5000/highlands/
  * https://localhost:5000/aoste/
  * https://localhost:5000/fictive/
  * https://localhost:5000/global/
3. Each URL should have a different logo and color in the header

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/templates/rero_ils/head.html
+++ b/rero_ils/templates/rero_ils/head.html
@@ -63,5 +63,6 @@
   {%- block header %}{% endblock header %}
   {%- block css %}
     {% assets "rero_ils_main_css" %}<link href="{{ ASSET_URL }}" rel="stylesheet">{% endassets %}
+    <link href="{{ config.get('RERO_ILS_THEME_ORGANISATION_CSS_ENDPOINT') }}{{ viewcode }}.css" rel="stylesheet" title="{{ viewcode }}">
   {%- endblock css %}
 {%- endblock head %}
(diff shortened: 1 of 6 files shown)


rero_ils/templates/rero_ils/head.html
```diff
@@ -63,5 +63,6 @@
   {%- block header %}{% endblock header %}
   {%- block css %}
     {% assets "rero_ils_main_css" %}<link href="{{ ASSET_URL }}" rel="stylesheet">{% endassets %}
+    <link href="{{ config.get('RERO_ILS_THEME_ORGANISATION_CSS_ENDPOINT') }}{{ viewcode }}.css" rel="stylesheet" title="{{ viewcode }}">
```
jma: Can you check that this works in production (non debug mode). Probably it will failed due to APP_DEFAULT_SECURE_HEADERS policies.

blankoworld: You're right this approach is the simplest I found regarding given constraint: change the CSS even if the server is launched. So yes I know that's a naive one. And that in the future it will evolve.

It avoids problems with static files that are (re)generated at each deployment, that the production don't keep them in kubernetes and so on.

I also think about `APP_DEFAULT_SECURE_HEADERS` policy, and found nothing about CSS and images.
Seems to be OK with `ENV=production` in **invenio.cfg** file.

PR #836, opened 2020-03-04
tests: fix problems with daylight saving time

blankoworld:
* Fixes test that have problem twice a year with daylight saving time
  timezone

To fix problems about timezones in tests.

## How to test?

```bash
pipenv run pytest -vvs --no-cov --disable-warnings tests/api/test_loans_rest.py::test_due_soon_loans
```

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/tests/api/test_loans_rest.py
+++ b/tests/api/test_loans_rest.py
@@ -27,7 +27,7 @@ from flask import url_for
 from invenio_accounts.testutils import login_user_via_session
 from invenio_circulation.api import get_loan_for_item
 from invenio_circulation.search.api import LoansSearch
-from utils import flush_index, postdata
+from utils import check_timezone_date, flush_index, postdata
 
 from rero_ils.modules.items.api import Item
 from rero_ils.modules.libraries.api import Library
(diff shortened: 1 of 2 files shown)


tests/api/test_loans_rest.py
```diff
@@ -170,27 +170,20 @@ def test_due_soon_loans(client, librarian_martigny_no_email,
...
-    assert end_date.minute == 59 and end_date.hour == 23
+    loan_date = ciso8601.parse_datetime(checkout_loan.get('end_date'))
+
+    # should be 22:59 in UTC+0
+    tocheck_date = loan_date
+    assert tocheck_date.minute == 59 and tocheck_date.hour == 22
+
+    # should be 14:59/15:59 in US/Pacific (because of daylight saving time)
+    check_timezone_date(pytz.timezone('US/Pacific'), loan_date, [14, 15])
+    check_timezone_date(pytz.timezone('Europe/Amsterdam'), loan_date)
```
jma: Why here the expected parameter is not defined as this timezone have also a winter and summer time GMT+1 and GMT+2?

blankoworld: I consider that Europe/Zurich was so close to Europe/Amsterdam that we can be confident about the result: it would be 1 or 2 hours of difference.
Do you prefer I add them too?

PR #857, opened 2020-03-18
bootstrap: fix useless tgz file installation

### LINKED ISSUE #856: Bootstrap: attempt to install TGZ_FILE even if no option in command line
**Describe the issue**

While using `pipenv run bootstrap` (without `-t rero-ils-ui.tgz` option), it attempts to install a TGZ file. You can see that:

```
[INFO]: Install RERO-ILS-UI from tgz: 
```

It shouldn't.

**To Reproduce**

1. `pipenv run bootstrap`
1. Check that `[INFO]: Install RERO-ILS-UI from tgz: ` appears without any TGZ filename (end of line is empty, no specific file is given).

**Expected behavior**

* If no option `-t` given, it shouldn't display `[INFO]: Install RERO-ILS-UI from tgz: `
* If option `-t` without any file, it should raise an error like `[ERROR]: package tgz file dos not exist: `
* If option `-t rero-rero-ils-ui-0.0.12.tgz` is given, it should display `[INFO]: Install RERO-ILS-UI from tgz: rero-rero-ils-ui-0.0.12.tgz`

**Software version**

3cd35f9 (current dev).

**Additional context**

The problem comes from `scripts/bootstrap` file, probably because of the use of `[` instead of `[[`.

It happens here: https://github.com/rero/rero-ils/blob/3cd35f99a1cf42a6bee3df41d2d8ded8189139e8/scripts/bootstrap#L84
and here:
https://github.com/rero/rero-ils/blob/3cd35f99a1cf42a6bee3df41d2d8ded8189139e8/scripts/bootstrap#L138

blankoworld:
While using bootstrap script, it attempts to install TGZ_FILE.
Something like `[INFO]: Install RERO-ILS-UI from tgz: ` appears. But no
file were given.

* Fixes #856 - by changing `[` tests to `[[` (bash)

Because of #856 issue.

## How to test?

Launch `pipenv run bootstrap` and check that no `[INFO]: Install RERO-ILS-UI from tgz: ` appears.

Then do the same with a TGZ file, for example those from https://github.com/rero/rero-ils-ui/releases/tag/v0.0.12, like : 

`pipenv run bootstrap -t v0.0.12.tar.gz`

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -58,7 +58,7 @@ success_msg() {
 msg "PROGRAM: ${PROGRAM}"
 
 # options may be followed by one colon to indicate they have a required argument
-if ! options=$(getopt -o dsb -l -t deploy,ci,tgz_path: -- "$@")
+if ! options=$(getopt -o dct: -l deploy,ci,tgz_package: -- "$@")
 then
     # something went wrong, getopt will put out an error message for us
     exit 1
@@ -81,7 +81,7 @@ do
       ci=true ;;
     -t|--tgz_package)
       tgz_file=$2
-      if [ ! -f ${tgz_file} ]
+      if [[ ! -f "${tgz_file}" ]]
       then
         error_msg+exit "package tgz file dos not exist: ${tgz_file}"
       fi ;;
@@ -93,7 +93,7 @@ do
 done
 
 # PIPENV is a mandatory condition to launch this program!
-if [[ -z  "${VIRTUAL_ENV}" ]]; then
+if [[ -z "${VIRTUAL_ENV}" ]]; then
   error_msg+exit "Error - Launch this script via pipenv command:\n\tpipenv run ${PROGRAM}"
 fi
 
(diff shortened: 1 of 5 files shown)


scripts/bootstrap
```diff
@@ -81,7 +81,7 @@ do
       ci=true ;;
     -t|--tgz_package)
       tgz_file=$2
-      if [ ! -f ${tgz_file} ]
+      if [[ ! -f ${tgz_file} ]]
```
rerowep: This one is working like this for me: `if [ ! -f ${tgz_file} ]`

scripts/bootstrap
```diff
@@ -135,7 +135,7 @@ info_msg "Search static folder location"
 static_folder=$(pipenv run invenio shell --no-term-title -c "print('static_folder:%s' % app.static_folder)"|grep static_folder| cut -d: -f2-)
 info_msg "Install static folder npm dependencies in: ${static_folder}"
 npm install --prefix "${static_folder}"
-if [ -f ${tgz_file} ]
+if [[ -f ${tgz_file} ]]
```
rerowep: But this one has to have two [[ ?

scripts/bootstrap
AoNoOokami: Not on Debian devices or VM

PR #868, opened 2020-03-27
subscription: adds listener methods to create a subscription if needed

zannkukai:
When a patron is create/updated, if the patron is linked to a patron type requiring a subscription,
we need to create a subscription transaction if the patron doesn't have any current one.
This is the purpose of this PR.

* Adds a listerner method for patron signals : after_record_create, after_record_update
* Adds unit tests

## How to test?

- Edit a patron_type to have one with a yearly subscription
- Create a new patron and assign it with this patron_type
- Check the patron fee tabs, you should see a new subscription fee

- Update this patron to assign it to a patron_type without subscription
- Update again this patron to re-assign it the patron_type with a subscription
- No new fee should be assign to the patron

- Search for a patron previously linked to the patron_type with subscription
- Edit this patron 
- Check this patron has now a subscription fee.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/patrons/api.py
+++ b/rero_ils/modules/patrons/api.py
@@ -37,8 +37,7 @@ from ..minters import id_minter
 from ..organisations.api import Organisation
 from ..patron_transactions.api import PatronTransaction
 from ..providers import Provider
-from ..utils import get_ref_for_pid
-from ..utils import trim_barcode_for_record
+from ..utils import get_ref_for_pid, trim_barcode_for_record
 
 _datastore = LocalProxy(lambda: current_app.extensions['security'].datastore)
 
@@ -230,6 +229,10 @@ class Patron(IlsRecord):
         loans = self.get_number_of_loans()
         if loans:
             links['loans'] = loans
+        transactions = PatronTransaction.get_transactions_count_for_patron(
+            self.pid, status='open')
+        if transactions > 0:
+            links['transactions'] = transactions
         return links
 
     def reasons_to_keep(self):
@@ -347,7 +350,8 @@ class Patron(IlsRecord):
         subs = filter(is_subscription_valid, self.get('subscriptions', []))
         return list(subs)
 
-    def add_subscription(self, patron_type, start_date, end_date):
+    def add_subscription(self, patron_type, start_date, end_date,
+                         dbcommit=True, reindex=True, delete_pids=False):
         """Add a subscription to a patron type.
 
         :param patron_type: the patron_type linked to the subscription
@@ -356,7 +360,7 @@ class Patron(IlsRecord):
         """
         transaction = PatronTransaction.create_subscription_for_patron(
             self, patron_type, start_date, end_date,
-            dbcommit=True, reindex=True, delete_pid=True)
+            dbcommit=dbcommit, reindex=reindex, delete_pid=delete_pids)
         if transaction:
             subscriptions = self.get('subscriptions', [])
             subscriptions.append({
@@ -370,4 +374,4 @@ class Patron(IlsRecord):
                 'end_date': end_date.strftime('%Y-%m-%d'),
             })
             self['subscriptions'] = subscriptions
-            self.update(self, dbcommit=True, reindex=True)
+            self.update(self, dbcommit=dbcommit, reindex=reindex)
--- a/rero_ils/modules/patrons/listener.py
+++ b/rero_ils/modules/patrons/listener.py
@@ -17,7 +17,11 @@
 
 """Signals connector for patron."""
 
+from datetime import datetime
+
 from .api import Patron, PatronsSearch
+from ..patron_types.api import PatronType
+from ..utils import add_years, get_schema_for_resource
 
 
 def enrich_patron_data(sender, json=None, record=None, index=None,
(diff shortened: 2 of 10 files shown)


rero_ils/modules/patrons/listener.py
```diff
@@ -38,3 +42,30 @@ def enrich_patron_data(sender, json=None, record=None, index=None,
...
+    a new subscription patron transaction should be created or not. If a
+    subscriotion is needed, then the 'subscription' patron attribute will be
+    updated.
+    This method should be connect with 'before_record_insert' and
+    'before_record_update'.
+
+    :param record: the record being performed
+    """
+    if record.get('$schema') != get_schema_for_resource(Patron):
+        return
```
rerowep: Should we return or raise an error?

rero_ils/modules/patrons/api.py
```diff
@@ -221,12 +221,21 @@ def get_number_of_loans(self):
         results = search.source().count()
         return results
 
+    def get_number_of_open_transactions(self):
+        """Get number of open transactions for this patron."""
+        trans = PatronTransaction.get_transactions_pids_for_patron(
+            self.pid, status='open')
+        return len(list(trans))
```
rerowep: I don't know how big the list could be, but normally a query with count should be faster.

rero_ils/modules/patrons/listener.py
```diff
@@ -38,3 +42,30 @@ def enrich_patron_data(sender, json=None, record=None, index=None,
...
+    if record.patron_type_pid is None:
+        return
+    ptty = PatronType.get_record_by_pid(record.patron_type_pid)
+    if ptty.is_subscription_required and not record.has_valid_subscription:
+        # TODO : (2020-03-27)
+        #   At this time, subscription are only possible for one year. In
+        #   the future, the subscription period should be defined as a
+        #   patron_type attribute.
+        start_date = datetime.now()
```
rerowep: Do we need to consider the time zone?

rero_ils/modules/utils.py
```diff
@@ -161,3 +161,24 @@ def add_years(initial_date, years):
     except ValueError:
         return initial_date + (date(initial_date.year + years, 1, 1) -
                                date(initial_date.year, 1, 1))
+
+
+def get_schema_for_resource(resource):
```
rerowep: Good function, maybe we have to open a task to use this function in: IlsRecord, Holding, Loan, PatronTransactionEvent and PatronTRansaction.

tests/unit/test_utils.py
```diff
@@ -54,3 +56,8 @@ def test_add_years():
     assert tow_years_later.month == 3 and tow_years_later.day == 1
     assert four_years_later.month == initial_date.month and \
         four_years_later.day == initial_date.day
+
+
+def test_get_schema_for_resources(app):
+    """Test get_schemas_for_resource function."""
+    assert get_schema_for_resource(Patron) == get_schema_for_resource('ptrn')
```
rerowep: Maybe this test has not the expected effect if both functions calls responding None.
Would it be better to write two asserts like:
```
assert get_schema_for_resource(Patron) == 'https://ils.rero.ch/...'
assert get_schema_for_resource('ptrn') ==  'https://ils.rero.ch/...'
```

rero_ils/modules/patrons/listener.py
zannkukai: This is to filter the record to process. Each time a record is create/update, the signal is emit and the listener is called. But this listener must only be performed if it's a `Patron` record. If the record isn't a `Patron` then do nothing and return (not raise an error)

rero_ils/modules/patrons/api.py
zannkukai: You are right. Potentially a patron could have a loooooot of open transaction. But in the fact, it's inimaginable to have more than 20 open transactions. As the method already exists to get open transactions, and this method return a generator of pid, I just used it to not write a function doing the same but returning a count.

Maybe it's a solution to add an argument to change `get_transactions_pids_for_patrons` -->
```
def get_transactions_pids_for_patrons(patron_pid, status=None, return_type='pids'):
    ....
    if return_type == 'count':
        return query.source().count
    else:
        results = query.source('pid').scan()
        for result in results:
            yield result.pid
```

rero_ils/modules/patrons/listener.py
zannkukai: Don't think so. Start/End subscription date are stored as UTC date with format %Y-%m-%d.
In my point of view, timezone isn't relevant here.

tests/unit/test_utils.py
zannkukai: sure I can do that ;-)
Just a little bit time consuming if (one day) we change the json schema of `Patron` resource. 
Done

rero_ils/modules/patrons/api.py
zannkukai: After testing, we can't use `yield` and `return` in the same function. If Python found usage of `yield` in a function it always return a generator despite if `yield` is not reached.
I will create two separate functions

rero_ils/modules/patrons/api.py
```diff
@@ -353,7 +358,7 @@ def add_subscription(self, patron_type, start_date, end_date):
         """
         transaction = PatronTransaction.create_subscription_for_patron(
             self, patron_type, start_date, end_date,
-            dbcommit=True, reindex=True, delete_pid=True)
+            dbcommit=True, reindex=True, delete_pid=False)
```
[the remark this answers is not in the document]
zannkukai: done

tests/unit/conftest.py
```diff
@@ -25,6 +25,15 @@
 from pkg_resources import resource_string
 
 
+@pytest.fixture(scope='module')
+def create_app():
+    """Create test app."""
+    # from invenio_app.factory import create_ui
+    # create_ui
```
blankoworld, line 32: Why not deleting these commented lines?

PR #878, opened 2020-03-31
tests: implement first tests with cypress

AoNoOokami:
* Implements automated e2e testing.
* Adds cypress in folder tests/e2e/cypress and adapt tests scripts.
* Configures cypress base url to avoid repeating it and check that
server is up on launch.
* Creates first custom command.
* Creates first test (checkout-checkin).
* Adds a script to install cypress.

To integrate Cypress in the project

## How to test?

Documentation available here: https://github.com/rero/developer-resources/pull/13

### Command line
1. `poetry run cypress` (install and launch cypress tests, localhost:5000 server must be running a there should be a setup)

cypress

### With test launcher
1. `cd tests/e2e/cypress`
1. `npm run cypress` to open Cypress (localhost:5000 server must be running)
1. Run the test by clicking on "checkout-checkin.spec.js"

test

The test runner should start the test with the browser your choose

test2

And you should have tests passing like this:

test3

### Run a single test
1. `cd tests/e2e/cypress`
1. `npx cypress run --spec "./cypress/integration/circulation/checkout-checkin.spec.js"`

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?


tests/e2e/cypress/cypress/integration/circulation/checkout-checkin.spec.js
```diff
@@ -0,0 +1,85 @@
...
+    // Go to documents
+    cy.contains('Catalogue').click()
+    cy.contains('Documents').click()
+    cy.contains('Type de document', {timeout: timeOut})
+
+    // Go to the detailed view of a document
+    cy.contains(documentTitle, {timeout: timeOut}).click()
+
+    // Create an item
+    cy.contains('Ajouter...', {timeout: timeOut}).click()
```
jma, line 44: In french?

scripts/bootstrap
```diff
@@ -142,6 +142,10 @@ then
   npm install "${tgz_file}" --prefix "${static_folder}"
 fi
 
+# install cypress
+info_msg "Installing cypress"
+npm install cypress --save-dev --prefix "tests/e2e/cypress"
```
jma, line 147: I would not put here. I would add a new option in `./scripts/setup` to perform cypress tests and I would check if the npm install is already done or not.

AoNoOokami: I think that this should be discussed.

blankoworld:
I tested on GNU/Linux (ArchLinux), with Chromium 83, Firefox 76.0.1 and Electron.
It seems OK.

scripts/bootstrap
```diff
@@ -142,6 +142,10 @@ then
   npm install "${tgz_file}" --prefix "${static_folder}"
 fi
 
+# install cypress
+info_msg "Install cypress"
+npm install --prefix "tests/e2e/cypress"
```
blankoworld, line 147: Does it mean that it will install all previous installed npm dependencies in another directory?

Is that possible to configure cypress to use rero-ils npm dependencies? Or to just install cypress in `${static_folder}`?

AoNoOokami: Cypress is a testing application and I think its place is not in static folder but in tests folder. It will not install `all`previous installed npm dependencies, but only the default ones and those specified in corresponding package.json.

blankoworld: Cypress dependencies takes 53Mb on my side. RERO dependencies takes 150Mb.
What about this place in our Dockerfile? It will grow up the image, but without any improvement on RERO-ils functionnality.

I suggest to add an option to avoid installing these dependencies.

jma: @blankoworld Cypress should not be installed into a Docker image at all. It is just a dev tool. For me it should be installed in his own directory and should be installed on demand. For example by adding an option to the bootstrap or with his own script.

AoNoOokami: I agree with the option in bootstrap or an own script.

blankoworld: I agree that Cypress should not be in Docker image. But with the given script in bootstrap, it is. And that was my question: why not excluding it from the Docker image (and so, we should find a solution).

The solution you suggest is a good idea because cypress will either not be in docker image nor in bootstrap/setup.

scripts/cypress
```diff
@@ -0,0 +1,79 @@
...
+command_exists () {
+    type "$1" &> /dev/null ;
+}
+
+# Displays program name
+msg "PROGRAM: ${PROGRAM}"
+
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
blankoworld, line 66: What do you think about `DIR=$(cd `dirname $0` && pwd)`?

scripts/bootstrap
AoNoOokami: The solution we choose to use is a separated script

scripts/cypress
```diff
@@ -0,0 +1,79 @@
...
+export PATH=${PATH}:${DIR}/../tests/e2e/cypress/node_modules/.bin/
+
+if command_exists cypress
```
blankoworld, line 69: It would be better to use `"cypress"`.

scripts/cypress
```diff
@@ -0,0 +1,79 @@
...
+    error_msg "${1}" && exit 1
+}
+
+success_msg() {
+  colored_msg "${SUCCESS_COLOR}" "SUCCESS" "${1}"
+}
+
+command_exists () {
+    type "$1" &> /dev/null ;
+}
```
blankoworld, line 59: I suggest something like:
```
check_cypress() {
  if test -n $(command -v "$1" &> /dev/null); then
    export CYPRESS=`which "$1"`
    info_msg "Found cypress: ${CYPRESS}"
  else
    info_msg "Install cypress"
    npm install --prefix "tests/e2e/cypress"
    export CYPRESS=`which "$1"`
    info_msg "Cypress path: ${CYPRESS}"
  fi
}
```
and then delete line 69-76.
Finally line 79 change to `${CYPRESS} run -P ./tests/e2e/cypress`.

AoNoOokami: ping @jma ?

AoNoOokami:
> Installed and tested locally (Debian 10)
> Great Job !

Thanks. Can you run a single test? I think I added that in the description just before your comment.

PR #887, opened 2020-04-02
issues: trim item and patron barcodes

BadrAly:
To ensure item and patron barcodes will not have trailing and leading spaces
for records loaded via the APIs.

* Removes the unused transactions resource and update config files accordingly.
* Closes rero/rero-ils-ui#213

https://tree.taiga.io/project/rero21-reroils/task/1400?kanban-status=1224894

## How to test?

`bootstrap`

## Code review check list

- [x] File names.
- [x] Functions names.
- [x] Functions docstrings.
- [ ] Extracted translations?


jma:
> > I would not do it at the backend side. This is not his responsibility. Why for the barcode and not > for other fields, such as email, names, patron barcode, etc.
> 
> This issue was discussed during on the daily meeting and we have agreed to have it in backend side as well.
> 
> Why the barcode only? because the issue addressed only the barcode for now. [rero/rero-ils-ui#213](https://github.com/rero/rero-ils-ui/issues/213)
> 
> Please note the UI is not the only way to create items/patrons on the database.

Oups sorry I have misunderstood the purpose of this PR. I believed that the idea was to clean the barcode during the query or circulation actions. This is not the case this is for the item creation.  I will do the review soon.

PR #893, opened 2020-04-02
permissions: update permission API and configuration

zannkukai:
Some ressource permissions defined into the main ReroILS configuration file wasn't correct.
Some resources (circulation policy, patron type, item type, budget, ...) should only be created by
'system_librarian` user. This PR fix this problem

The permission API are updated to take care of main configuration for 'can_delete' permission.
The permission API only checked if the record is linked to other resources (children). If not, despite
the main configuration, it return True. Now, if a ressource has no linked resources, then the API
check also if user has the permission to delete the record.

## How to test?

- Login as Astrid (system_librarian)
- check `~/api/permissions/circ_policies/1`
- You should "can_update" but cannot "can_delete" (due to children)
- Create an new CIPO
- check `~/api/permissions/circ_policies/[NEW_ID]`
- You should "can_update" and "can_delete"

- Login as Jean (librarian)
- check `~/api/permissions/circ_policies/[NEW_ID]`
- You shouldn't "can_update" and "can_delete" (permission)

## Code review check list

- [x] File names.
- [x] Functions names.
- [x] Functions docstrings.
- [x] Extracted translations?

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -51,9 +51,9 @@ from .modules.acq_invoices.api import AcquisitionInvoice
 from .modules.acq_order_lines.api import AcqOrderLine, AcqOrderLinesIndexer
 from .modules.acq_orders.api import AcqOrder
 from .modules.budgets.api import Budget
-from .modules.budgets.permissions import can_create_budgets_factory, \
-    can_list_budgets_factory, can_update_delete_budgets_factory
+from .modules.budgets.permissions import can_list_budgets_factory
 from .modules.circ_policies.api import CircPolicy
+from .modules.circ_policies.permissions import can_update_circ_policy_factory
 from .modules.documents.api import Document, DocumentsIndexer
 from .modules.holdings.api import Holding, HoldingsIndexer
 from .modules.item_types.api import ItemType
@@ -61,8 +61,7 @@ from .modules.items.api import Item, ItemsIndexer
 from .modules.items.permissions import can_create_item_factory, \
     can_update_delete_item_factory
 from .modules.libraries.api import Library
-from .modules.libraries.permissions import can_create_library_factory, \
-    can_delete_library_factory, can_update_library_factory
+from .modules.libraries.permissions import can_update_library_factory
 from .modules.loans.api import Loan
 from .modules.loans.permissions import can_list_loan_factory, \
     can_read_loan_factory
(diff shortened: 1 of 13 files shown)


rero_ils/config.py
```diff
@@ -665,7 +664,7 @@ def _(x):
         search_factory_imp='rero_ils.query:organisation_organisation_search_factory',
         list_permission_factory_imp=can_access_organisation_patrons_factory,
         create_permission_factory_imp=deny_all,
-        update_permission_factory_imp=can_update_organisations_factory,
+        update_permission_factory_imp=is_system_librarian_organisation_record_factory,
```
BadrAly, line 667: I recall that it was decided that the `organisation` record should not be manipulated by a librarian/sys_librarian. instead it will be managed by rero_centrale or ucl_admins... possibly need to talk to PO about this.

rero_ils/config.py
```diff
@@ -795,9 +794,9 @@ def _(x):
         max_result_window=10000,
         search_factory_imp='rero_ils.query:organisation_search_factory',
         read_permission_factory_imp=can_access_organisation_records_factory,
-        create_permission_factory_imp=can_create_organisation_records_factory,
-        update_permission_factory_imp=can_update_organisation_records_factory,
-        delete_permission_factory_imp=can_delete_organisation_records_factory,
+        create_permission_factory_imp=is_system_librarian_organisation_record_factory,
+        update_permission_factory_imp=is_system_librarian_organisation_record_factory,
```
BadrAly, line 798: I think the librarian needs to have the update permissions of `cipo`. Maybe the excel file of permissions is not up to date. We can have `cipo` at the library level. another point for the PO

rero_ils/modules/permissions.py
```diff
@@ -35,17 +35,29 @@ def jsonify_permission_api_response(
...
     try:
-        rec_class, update_permission = \
+        rec_class, update_permission, delete_permission = \
             get_record_class_update_permission_from_route(route_name)
         record = rec_class.get_record_by_pid(record_pid)
 
         if not record:
             return jsonify({'status': 'error: Record not found.'}), 404
 
+        # We have tow behavior for 'can_delete'. Either the record has linked
```
BadrAly, line 45: two instead of tow

rero_ils/permissions.py
```diff
@@ -62,6 +62,23 @@ def staffer_is_authenticated(user=None):
     return None
 
 
+def user_has_roles(user=None, roles=[], condition='or'):
```
BadrAly, line 65: can you create a test for this a case when you have  two roles as input?

rero_ils/config.py
zannkukai: I used `deny_all` for all organisation permissions

rero_ils/config.py
zannkukai: I create a special permission for `update`.

rero_ils/modules/permissions.py
zannkukai: done

rero_ils/permissions.py
zannkukai: done ! Allow to detect a bug in this function and into other permission functions !

PR #937, opened 2020-04-18
tests: fix pytest-invenio static path location

jma:
pytest-invenio introduces a bug in a recent release. A pull request is
pending (https://github.com/inveniosoftware/pytest-invenio/pull/44).
This is a temporary fix and should be removed once pytest-invenio will
be released with the resolution. This appens only for a local
installation that change the `INVENIO_INSTANCE_PATH` variable.

Signed-off-by: Johnny Mariéthoz <Johnny.Mariethoz@rero.ch>

- Which issue does it fix?

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,10 @@
 """Common pytest fixtures and plugins."""
 
 import json
+import os
+import shutil
+import sys
+import tempfile
 from os.path import dirname, join
 
 import pytest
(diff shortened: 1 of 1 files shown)


blankoworld:
It fails on my computer.

```
E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpgwcxchsa/static/scss/rero_ils/styles.scss' 
```

If we have no `.env` file, and so no INVENIO_INSTANCE_PATH set, it fails.

The bootstrap put files in `sys.prefix`/var/instance/static.

Is there a way to manage the 2 cases?

jma:
> It fails on my computer. Because I have no `.env` file, and so no INVENIO_INSTANCE_PATH is set.
> 
> ```
> E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpgwcxchsa/static/scss/rero_ils/styles.scss' 
> ```
> 
> The bootstrap put files in `sys.prefix`/var/instance/static. Changing your code with `sys.prefix` works.
> 
> Is there a way to manage the 3 cases:
> 
> * first INVENIO_INSTANCE_PATH (if set)
> * then `sys.prefix` if exists
> * finally temporary directory? (doesn't work on my side though)
> 
> If this PR is OK, I think we can delete this method: https://github.com/rero/rero-ils/blob/US1293-doo-invenio32/rero_ils/modules/cli.py#L1534 as it was created because of pytest-invenio instance path problem.
> 
> I tried something here: https://github.com/blankoworld/rero-ils/tree/maj-doo-fix-pytest-invenio , could you please have a look?

I should be ok now, proposed by Peter.

blankoworld:
> > It fails on my computer. Because I have no `.env` file, and so no INVENIO_INSTANCE_PATH is set.
> > ```
> > E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpgwcxchsa/static/scss/rero_ils/styles.scss' 
> > ```
> > 
> > 
> > The bootstrap put files in `sys.prefix`/var/instance/static. Changing your code with `sys.prefix` works.
> > Is there a way to manage the 3 cases:
> > 
> > * first INVENIO_INSTANCE_PATH (if set)
> > * then `sys.prefix` if exists
> > * finally temporary directory? (doesn't work on my side though)
> > 
> > If this PR is OK, I think we can delete this method: https://github.com/rero/rero-ils/blob/US1293-doo-invenio32/rero_ils/modules/cli.py#L1534 as it was created because of pytest-invenio instance path problem.
> > I tried something here: https://github.com/blankoworld/rero-ils/tree/maj-doo-fix-pytest-invenio , could you please have a look?
> 
> I should be ok now, proposed by Peter.

On my side, without .env, I think it would not work. That's why I suggest to use INVENIO_INSTANCE_PATH, then sys.prefix and finally mktemp if no directory exists.
Check here: https://github.com/blankoworld/rero-ils/commit/2834221cb3f2b929cea8484609fd19626e913e0e#diff-484462fced51d1a06b1d93b4a44dd535R142

PR #938, opened 2020-04-18
setup: fix ref. prob. on responsibilityStatement

blankoworld:
When you setup RERO-ils, some document don't have any
responsibilityStatement. Which make the setup failed on some loan
creation.
This commit fixes the problem.

Because of errors in SETUP (`Exception loan requested_by_others: 'NoneType' object is not iterable`).

## How to test?

`pipenv run setup`

Check that no errors (in red) appears during create_loans setup.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/notifications/api.py
+++ b/rero_ils/modules/notifications/api.py
@@ -140,7 +140,7 @@ class Notification(IlsRecord):
 
             from ..documents.views import create_title_responsibilites
             responsibility_statement = create_title_responsibilites(
-                document.get('responsibilityStatement')
+                document.get('responsibilityStatement', [])
             )
             data['loan']['document']['responsibility_statement'] = \
                 next(iter(responsibility_statement or []), '')


blankoworld:
> I’m not sure that your title commit message is good.

You're right. I need to change it. But no idea for the moment :-/

rero_ils/modules/notifications/api.py
```diff
@@ -140,7 +140,7 @@ def replace_pids_and_refs(self):
 
             from ..documents.views import create_title_responsibilites
             responsibility_statement = create_title_responsibilites(
-                document.get('responsibilityStatement')
+                document.get('responsibilityStatement', [])
             )
             data['loan']['document']['responsibility_statement'] = \
```
rerowep, line 145: I should have used `try``except`.

PR #968, opened 2020-04-30
item: multiple inheritance for the item class

BadrAly:
Since the item class is used now to manage two things:
1. the item record
2. the item circulation transactions

Two base classes are created for each section above in
api.record:ItemRecord, api.circulation:ItemCirculation
The main Item class inherites from these two classes.

re-organising the class Item

## How to test?
Must have `rero_ils_ui.dev`
./run_tests.sh

## Code review check list

- [x] File names.
- [x] Functions names.
- [x] Functions docstrings.
- [ ] Extracted translations?


rero_ils/modules/items/api/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+    (Provider,),
+    dict(identifier=ItemIdentifier, pid_type='item')
+)
+# minter
+item_id_minter = partial(id_minter, provider=ItemProvider)
+# fetcher
+item_id_fetcher = partial(id_fetcher, provider=ItemProvider)
+
+
+class ItemsSearch(IlsRecordsSearch):
```
lauren-d, line 47: I think that the primary use of __init__.py is to initialize Python packages. Is it possible to move class somewhere else? maybe into `helpers.py` or why not into `api/record.py`?

rero_ils/modules/items/api/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        doc_types = None
+
+    @classmethod
+    def flush(cls):
+        """Flush indexes."""
+        current_search.flush_and_refresh(DocumentsSearch.Meta.index)
+        current_search.flush_and_refresh(cls.Meta.index)
+
+
+class Item(ItemRecord, ItemCirculation):
```
lauren-d, line 63: same before

rero_ils/modules/items/api/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        return cannot_delete
+
+    def dumps(self, **kwargs):
+        """Return pure Python dictionary with record metadata."""
+        dump = super(Item, self).dumps(**kwargs)
+        dump['available'] = self.available
+        return dump
+
+
+class ItemsIndexer(IlsRecordsIndexer):
```
lauren-d, line 100: Same before

rero_ils/modules/items/api/__init__.py
jma: `__init__.py` can contains code. This is not a problem. But for more clarity we can put in `api/api.py`, but this require to defined an `__all__` variable. Putting this in the helpers is not a good idea, nor in record as the Item inherit from two classes.

PR #971, opened 2020-05-01
item: field call number is now optional

BadrAly:
The item call number is now an optional field in both the
item json schema and item editor. This field is not populated
if not given by the librarian.

The item barcode remains a required field in the item json schema,
on the other side, it is an optional field in the item editor.

If a librarian decides not to add an item barcode, the system
will automatically populate the barcode with the following:
f-current timestamp.

* Adapts schema, tests and files correspondingly.

https://tree.taiga.io/project/rero21-reroils/us/1328?milestone=263424

## How to test?

`bootstrap` + `setup`
in the UI, you should be able to add items without barcodes

## Code review check list

- [x] File names.
- [x] Functions names.
- [x] Functions docstrings.
- [x] Extracted translations?

### DIFF
--- a/rero_ils/modules/items/api.py
+++ b/rero_ils/modules/items/api.py
@@ -46,7 +46,7 @@ from ..minters import id_minter
 from ..organisations.api import Organisation
 from ..patrons.api import Patron, current_patron
 from ..providers import Provider
-from ..utils import trim_barcode_for_record
+from ..utils import generate_item_barcode, trim_barcode_for_record
 from ...filter import format_date_filter
 
 # provider
@@ -177,8 +177,12 @@ class Item(IlsRecord):
         """Create item record."""
         cls._item_build_org_ref(data)
         data = trim_barcode_for_record(data=data)
+        # Since the barcode is a mandatory field, we set it to current
+        # timestamp if not given
+        data = generate_item_barcode(data=data)
         record = super(Item, cls).create(
             data, id_, delete_pid, dbcommit, reindex, **kwargs)
+
         if not data.get('holding'):
             record.link_item_to_holding()
         return record
@@ -192,10 +196,24 @@ class Item(IlsRecord):
         :returns: The updated item record.
         """
         data = trim_barcode_for_record(data=data)
+        data = generate_item_barcode(data=data)
         super(Item, self).update(data, dbcommit, reindex)
         # TODO: some item updates do not require holding re-linking
         self.link_item_to_holding()
 
+        return self
+
+    def replace(self, data, dbcommit=False, reindex=False):
+        """Replace an item record.
+
+        :param data: The record to replace.
+        :param dbcommit: boolean to commit the record to the database or not.
+        :param reindex: boolean to reindex the record or not.
+        :returns: The replaced item record.
+        """
+        # update item record with a generated barcode if does not exist
+        data = generate_item_barcode(data=data)
+        super(Item, self).replace(data, dbcommit, reindex)
         return self
 
     @classmethod
@@ -216,7 +234,10 @@ class Item(IlsRecord):
         data['organisation'] = org_ref
 
     def link_item_to_holding(self):
-        """Link an item to a standard holding record."""
+        """Complete the item record.
+
+        Link an item to a standard holding record.
+        """
         from ..holdings.api import \
             get_standard_holding_pid_by_doc_location_item_type, \
             create_holding
@@ -235,6 +256,7 @@ class Item(IlsRecord):
 
         base_url = current_app.config.get('RERO_ILS_APP_BASE_URL')
         url_api = '{base_url}/api/{doc_type}/{pid}'
+        # update item record with the parent holding record
         self['holding'] = {
             '$ref': url_api.format(
                 base_url=base_url,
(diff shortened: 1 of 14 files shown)


rero_ils/modules/items/jsonschemas/items/item-v0.0.1.json
```diff
@@ -36,7 +34,7 @@
     },
     "barcode": {
       "title": "Barcode",
-      "description": "Barcode of the item.",
+      "description": "Barcode of the item. This field is optional",
```
zannkukai, line 37: See with Igor. I think it should be finished by a ".".
Maybe just "(optional)" string is enough.

rero_ils/modules/items/jsonschemas/items/item-v0.0.1.json
```diff
@@ -53,7 +51,7 @@
     },
     "call_number": {
       "title": "Call number",
-      "description": "Call number of the item.",
+      "description": "Call number of the item. This field is optional",
```
zannkukai, line 54: same here

rero_ils/modules/items/jsonschemas/items/item-v0.0.1.json
iGormilhit: @zannkukai is right, trailing dot. But more than that, is this description actually useful for the end user? I think the title is enough. And "required or optional" should be visually displayed (red * and validation message).

So I would delete the description field. See [JSON schema guidelines, section description](https://github.com/rero/developer-resources/blob/master/guidelines/json-schemas.md#description).

rero_ils/modules/items/jsonschemas/items/item-v0.0.1.json
iGormilhit: Yes, same as before, I recommend deleting the description.

rero_ils/modules/items/jsonschemas/items/item-v0.0.1.json
```diff
@@ -7,8 +7,6 @@
   "required": [
     "$schema",
     "pid",
-    "barcode",
-    "call_number",
```
iGormilhit, line 11: But are these fields actually optional, I mean do we accept an item without barcode? If the barcode is generated by the system, then it is always filled in, it's not optional anymore. It only optional form the editor point of view.

It may be different with the call number, I don't know.

benerken: According to me :
1. call_number 
- could be empty 
- shouldn't get system generated value.
- shouldn't be mandatory in the item editor.

2. barcode
- could be empty (because the librarian doesn't know it when it creates the item) 
- should get system generated value when it's empty
- shouldn't be mandatory in the item editor.

iGormilhit: @benerken It just seems weird to me that a field that has always a value is not required by the validation tool, ie the schema. I do understand that the librarian should be able to leave this field empty. I don't know if we can achieve that. If not, then the field we not be required, thus leaving the possibility to be saved without any barcode (at least theoretically). My question is, could we live with it?

BadrAly: Item barcode is a required field populated by the current timestamp if not given.
Item call number is an optional field

BadrAly:
A new taiga task is added to adapt the item editor

rero_ils/modules/items/api.py
```diff
@@ -192,12 +198,30 @@ def update(self, data, dbcommit=False, reindex=False):
         :returns: The updated item record.
         """
         data = trim_barcode_for_record(data=data)
+        if not data.get('barcode'):
```
rerowep, line 201: We have the same code in the replace function. A `set_barcode` function could be usefull.

rero_ils/modules/items/api.py
```diff
@@ -192,12 +198,30 @@ def update(self, data, dbcommit=False, reindex=False):
...
+    def replace(self, data, dbcommit=False, reindex=False):
+        """Replace an item record.
+
+        :param data: The record to replace.
+        :param dbcommit: boolean to commit the record to the database or not.
+        :param reindex: boolean to reindex the record or not.
+        :returns: The replaced item record.
+        """
+        # update item record with a generated barcode if does not exist
+        if not data.get('barcode'):
```
rerowep, line 219: See above.

rero_ils/modules/items/api.py
```diff
@@ -177,8 +177,12 @@ def create(cls, data, id_=None, delete_pid=False,
         """Create item record."""
         cls._item_build_org_ref(data)
         data = trim_barcode_for_record(data=data)
+        # Since the barcode is a mandatory field, we set it to current
+        # timestamp if not given
+        data = generate_item_barcode(data=data)
```
jma, line 182: this does not correspond to the PO specification. I should  contains the PID value. The same logic should be present for the update.

BadrAly: On discussion with  the PO and Benoit, the specs of the barcode has changed to:
f-current_timestamp instead of f-pid

PR #980, opened 2020-05-07
US1305 data model illustrations colors physical details

rerowep:

- https://tree.taiga.io/project/rero21-reroils/us/1305

## Dependencies

* https://github.com/rero/rero-ils-ui/pull/261

## How to test?

- run bootstrap and setup
- verify all documents views working properly

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/data/documents_big.json
+++ b/data/documents_big.json
@@ -90,8 +90,16 @@
       }
     ],
     "extent": "660 p.",
-    "otherMaterialCharacteristics": "ill.",
-    "formats": [
+    "note": [
+      {
+        "noteType": "otherPhysicalDetails",
+        "label": "ill."
+      }
+    ],
+    "illustrativeContent": [
+      "illustrations"
+    ],
+    "dimensions": [
       "24 cm"
     ],
     "series": [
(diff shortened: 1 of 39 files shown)


JoelleDosimont:
I think there's a problem with the display of series. See, [](https://ilsdev.test.rero.ch/aoste/documents/1155). Only the number is displayed, but it's ok in the description tab. Both in public and professional view.

pronguen:
@JoelleDosimont It is not a problem. The number displayed above is the extent (300$a). In this case, it is a series of volume so they wrote only "1-". See [MARC record](http://data.rero.ch/01-R008586921/marchtml).
A better idea of how to display it, knowing that we cannot formally identify when there is a pagination or a volume enumeration?

JoelleDosimont:
> 
> 
> @JoelleDosimont It is not a problem. The number displayed above is the extent (300$a). In this case, it is a series of volume so they wrote only "1-". See [MARC record](http://data.rero.ch/01-R008586921/marchtml).
> A better idea of how to display it, knowing that we cannot formally identify when there is a pagination or a volume enumeration?

Sorry, didn't see that... If the problem is in the data, data should be cured ;-)
Is there a way of not display the extent when there is no alphabetical character (don't know if it's an option) ?

PR #981, opened 2020-05-08
tests: fix travis trouble with pytest-invenio

blankoworld:
Last days a [new pytest-invenio 1.2.2 version](https://github.com/inveniosoftware/pytest-invenio/commit/9a5d15bbfeda699bcc67805519861409c567c6f0) was released.
It downgrades our pytest-flask from 1.0.0 to 0.15.1 and our Flask from
1.1.2 to 1.04.
These changes break our tests. Mainly with Flask and werkzeug problems.
Cf. https://travis-ci.org/github/blankoworld/rero-ils/jobs/684571964#L1905

For now, this commit limits tests problem by keeping our pytest-invenio
version, pytest-flask and Flask.

* Fix pytest-invenio to 1.2.1 to avoid problems in our tests

This should fix the issue about tests like here: https://travis-ci.org/github/rero/rero-ils/jobs/684554447

## How to test?

* Check Travis tests
* Check changes in Pipefile and Pipfile.lock

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?


tests/api/test_items_rest_views.py
```diff
@@ -726,7 +726,8 @@ def test_update_loan_pickup_location(
     )
     assert res.status_code == 200
     assert data.get('pickup_location_pid') == new_loc_pid
-    # Change loan state to 'ITEM_AT_DESK'
+    # Change loan state to 'ITEM_AT_DESK'.
+    # WARNING: Use loan from test_patron_checkouts_order (item3_lib_martigny)
```
BadrAly, line 730: Is this a valid comment? because it is using the `item3_lib_martigny`
`# WARNING: Use loan from test_patron_checkouts_order (item3_lib_martigny)`

blankoworld: Yes! This explain that we use an item from a previous test!

Pipfile
```diff
@@ -102,7 +102,7 @@ marshmallow = ">=2.15.1,<3.0.0"
 pydocstyle = ">=3.0.0"
 pytest = ">=4.6.4"
 pytest-cov = ">=2.7.1"
-pytest-invenio = ">=1.2.1,<1.3.0"
+pytest-invenio = ">=1.3.0"
```
jma, line 105: This can probably be removed.

blankoworld: Do you talk about the entire line or simply remove the constraint on the version?

In the first case I don't understand how we can use pytest-invenio if no other package depends on it.
In the second case, I'm not sure that it will be safe to let pipenv choose the version as the 1.2.2 is problematic for us.

I think that's the way I do is the best case in term of dependancies.

For your information, that was my ideas before this PR: 

```
pytest-invenio = "==1.2.1"
pytest-invenio = ">=1.3.0"
pytest-invenio = ">=1.2.1,!=1.2.2"
```

PR #994, opened 2020-05-15
project: fix keyboard interruption for scripts

blankoworld:
We often run `poetry run server`. And we stop it with `Ctrl+C`. This
generally gives a keyboard interruption signal that stops the processus.

But this is broken: the processus is already runned!

This commit fixes the issue by sending again the interruption signal to
the runned script.

The `poetry run server` often let invenio server and celery worker running (after Ctrl+C keyboard interruption). When we launch it again: it returns an error about "Port already addressed".

This PR changes the behaviour by sending an interruption signal to launched script when using a keyboard interruption command.

## How to test?

* Launch `poetry run server`
* Wait few seconds (for server to be launched)
* Interrupt the server with [Ctrl] + [C] command
* `ps aux|grep invenio` should not return any launched process
* `ps aux |grep celery` should do the same (except if you have other invenio instances launched)

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/scripts/__init__.py
+++ b/scripts/__init__.py
@@ -17,6 +17,7 @@
 
 """Poetry script utils."""
 
+import os
 import subprocess
 import sys
 
@@ -29,9 +30,6 @@ import sys
 
 def run(prg_name):  # python 3.7+, otherwise define each script manually
     def fn():
-        res = subprocess.run(
-            [prg_name] + sys.argv[1:]
-        )  # run whatever you like based on 'name
-        if res.returncode:
-            sys.exit(res.returncode)
+        # Replace current Python program by prg_name (same PID)
+        os.execvp(prg_name, [prg_name] + sys.argv[1:])
     return fn


scripts/__init__.py
```diff
@@ -29,9 +31,18 @@
...
-        )  # run whatever you like based on 'name
-        if res.returncode:
-            sys.exit(res.returncode)
+        # run whatever you like based on 'name'
+        p = subprocess.Popen(
+            [prg_name] + sys.argv[1:])
+        try:
+            # check regularly p is alive
+            while p.poll() is None:
+                time.sleep(0.1) # time.sleep avoid too many CPU consumption.
```
rerowep, line 40: Why not working with signals?
https://stackabuse.com/handling-unix-signals-in-python/
http://avi.im/blag/2016/sigterm-in-python/

blankoworld: You're right, I can use a signal to capture the keyboard interruption. And so stop the program.

But to catch all other signals? Should I list them all?

I need to check further more the solution, regarding signals.

PR #1013, opened 2020-05-28
translation: fix translations API

zannkukai:
Fixes bug with translations API. If an exception occurred during
execution of this API, the exception should be logged into the
`current_app` logger. But `current_app` context was never loaded from
flask module before.
This PR also hacks translation files data : if a string is not
translated, rather than return an empty string, the original string will
be used.

* Adapts unitest to check theses cases.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/tests/api/test_translations.py
+++ b/tests/api/test_translations.py
@@ -16,21 +16,13 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 """Test translations API."""
-
-
+import mock
 from flask import url_for
 from utils import get_json
 
 
 def test_translations(client, app):
     """Test translations API."""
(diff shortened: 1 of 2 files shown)


tests/api/test_translations.py
```diff
@@ -41,3 +33,34 @@ def test_translations(client, app):
         )
         assert res.status_code == 200
         assert len(get_json(res)) > 0
+
+
+def test_translations_exceptions(client, app):
+
```
jma, line 39: docstring is missing.

zannkukai: done

PR #1019, opened 2020-06-03
release: v0.9.1

iGormilhit:

**There are two commits** because I had to slightly amend the README.rst file.

This patch is needed because #1013 is missing in `v0.9.0`, resulting in many
user interface elements that are missing.

## How to test?

- Check the detailed document view, the back button should be there, and the
  first tab should be labelled "Get". 
- Check the changed files. 

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/RELEASE-NOTES.rst
+++ b/RELEASE-NOTES.rst
@@ -18,6 +18,47 @@
 Release notes
 =============
 
+v0.9.1
+------
+
+This patch is needed because `#1013`_ is missing in ``v0.9.0``, resulting in
+many user interface elements that are missing.
+
+Documentation
+~~~~~~~~~~~~~
+
+-  Updates the chart of links between RERO ILS resources.
+-  Replace circulation chart with a new chart using ``dot`` from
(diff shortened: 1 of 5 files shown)


RELEASE-NOTES.rst
```diff
@@ -18,6 +18,47 @@
...
+-  Adds a markdown file with scenarios based on these actions.
+-  Enriches the mention of contributing libraries with UNamur and
+   U.Saint-Louis Brussels (``README.rst``).
+
+Translations
+~~~~~~~~~~~~
+
+-  Fixes a bug with the translation API that prevented exception to be
+   logged.
+-  Returns English source sting instead of an empty string when the translation
```
zannkukai, line 43: Minor : We are not really return the `English` string, only the source string (could be an ID, a context string, ...)

iGormilhit: :+1:

PR #1010, opened 2020-05-27
circulation: implement all the add_request actions

BadrAly:
Fixes problem when multiple pending requests are allowed for same
patron on same item.

Loans with state ITEM_IN_TRANSIT_TO_HOUSE are no longer blocking
new requests for same patron on same item.

Creates a method item_record_to_a_specific_loan_state to put an item
record in a specific status.

Reorganises the circulation units testing by creating a new directory
for circulation testings.

* Updates tests and fixtures accordingly.

Co-Authored-by: Aly Badr <aly.badr@rero.ch>

## Why are you opening this PR?

taiga_task: https://tree.taiga.io/project/rero21-reroils/task/1415

To implement the request actions defined by the group PO.
ref: https://hackmd.io/_3qU5KRiRZaLSYxaJrlsUA

## How to test?

- `run_tests.sh` 
- only code review at this moment.

## Code review check list

- [ ] Commit message template compliance.
- [ ] Commit message without typos.
- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Unnecessary commited files?
- [ ] Extracted translations?

PR #1020, opened 2020-06-03
circulation: implement the add_request action

BadrAly:
Takes part of the implementations of all actions after an effort
to modelize extensively all the circulation uses cases.

Fixes problem when multiple pending requests are allowed for same
patron on same item.

Loans with state ITEM_IN_TRANSIT_TO_HOUSE are no longer blocking
new requests for same patron on same item.

Creates a method item_record_to_a_specific_loan_state to put an item
record in a specific status.

Creates new circulation fixtures to return item, patron and loan on
a specific state.

Reorganises the circulation units testing by creating new directories
for circulation testings.

Added a class LoanState to better handle the different states of the loan.

* Updates tests and fixtures accordingly.

https://tree.taiga.io/project/rero21-reroils/task/1415?kanban-status=1224895

## How to test?
for now: `run_tests.sh`

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?


rero_ils/modules/loans/api.py
```diff
@@ -42,6 +43,19 @@
...
+
+    CREATED = 'CREATED'
+    PENDING = 'PENDING'
+    ITEM_IN_TRANSIT_FOR_PICKUP = 'ITEM_IN_TRANSIT_FOR_PICKUP'
+    ITEM_IN_TRANSIT_TO_HOUSE = 'ITEM_IN_TRANSIT_TO_HOUSE'
+    ITEM_AT_DESK = 'ITEM_AT_DESK'
+    ITEM_ON_LOAN = 'ITEM_ON_LOAN'
+    ITEM_RETURNED = 'ITEM_RETURNED'
+    CANCELLED = 'CANCELLED'
+
```
blankoworld, line 57: Why not do the same as `class ItemStatus` and use a `object`? It could be more useful because you can use `LoanState.CREATED` instead of `LoanState.CREATED.value`!

jma: I do not understand you point here. `ItemStatus` is very similar and we do not need `.value` here. But perhaps I misunderstood.

BadrAly: @jma You asked to change it to `.value` here -> https://github.com/rero/rero-ils/pull/1010#discussion_r431766491

blankoworld: You suggested it here: https://github.com/rero/rero-ils/pull/1010#discussion_r431766491 .
But to be consistent, either we use `Enum` for ItemStatus and LoanState or we use `object` for both :wink: 

In my opinion, `Enum` is only useful if you need to list all states in a web interface. It's the only advantage.

BadrAly: I will put it back as `object`.

rero_ils/modules/loans/api.py
```diff
@@ -81,7 +94,8 @@ class Loan(IlsRecord):
             'item': 'item'
         }
     }
-    DATE_FIELDS = []
+    DATE_FIELDS = [
+    ]
```
rerowep, line 98: Better to write `DATE_FIELDS = []`

tests/api/circulation/test_loan_utils.py
```diff
@@ -0,0 +1,114 @@
...
+    get_last_transaction_loc_for_item, get_loans_by_patron_pid
+from rero_ils.modules.loans.utils import can_be_requested
+from rero_ils.modules.locations.api import LocationsSearch
+
+
+def test_loan_utils(client, patron_martigny_no_email,
+                    patron2_martigny_no_email, circulation_policies,
+                    item_lib_martigny, librarian_martigny_no_email,
+                    loc_public_martigny):
+    """Test loan methods."""
```
jma, line 38: loan or loan utils methods?

jma: Is this file at the right place? tests/api is done for REST API tests. Moreover this is a test for a test function... not nice but ok.

tests/api/circulation/test_loan_utils.py
```diff
@@ -0,0 +1,114 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Tests REST API for loans."""
```
jma, line 18: bad comment

tests/utils.py
```diff
@@ -186,3 +190,83 @@ def check_timezone_date(timezone, date, expected=[]):
...
+    # perform circulation actions
+    if loan_state in [
+            LoanState.PENDING, LoanState.ITEM_AT_DESK,
+            LoanState.ITEM_ON_LOAN,
+            LoanState.ITEM_IN_TRANSIT_FOR_PICKUP,
+            LoanState.ITEM_IN_TRANSIT_TO_HOUSE
+    ]:
+        item, actions = item.request(**params)
+        loan = Loan.get_record_by_pid(actions[LoanAction.REQUEST].get('pid'))
+        # assert item.status == ItemStatus.ON_SHELF
```
jma, line 245: to be removed?

PR #1025, opened 2020-06-05
ui: keep selected tab active on reload

Garfield-fr:
* Keeps current tab for documents and persons detailed view on page
  reloads.

- selected tabs where lost during reload of page

## How to test?

- select a tab in documents or persons detailed view, select a tab and reload the page or change the language. The selected tab should not change.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/static/js/rero_ils/rero_keep_tab.js
+++ b/rero_ils/static/js/rero_ils/rero_keep_tab.js
@@ -1,0 +1,49 @@
+/*
+
+RERO ILS
+Copyright (C) 2020 RERO
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, version 3 of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+*/
+
+(function ( $, window ) {
+  $.fn.keeptab = function( options ) {
+
+    var opts = $.extend( {
+      tabsName: '.nav-tabs',
+      localeItemName: 'navkeeptab',
+      selector: '.nav-link'
+    }, options );
+
+    $(opts.tabsName).find(opts.selector).on('click', function() {
+      var elementId = undefined;
+      if ($(this).attr('id')) {
+        elementId = '#' + $(this).attr('id');
+      } else if ($(this).attr('href')) {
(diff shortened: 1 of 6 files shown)


Garfield-fr:
> Nice work, but do we have to put the bundle into persons. This is also used by documents, could we put the bundle in a more general place?

The jquery plugin loaded globally (root bundle) and used locally on persons and documents

rero_ils/static/js/rero_ils/rero_keep_tab.js
```diff
@@ -0,0 +1,36 @@
...
+
+*/
+
+(function ( $ ) {
+  $.fn.keeptab = function( options ) {
+    var opts = $.extend( {
+      localeItem: 'navkeeptab',
+      selector: '.nav-link'
+    }, options );
+    $('#' + this.attr('id') + ' ' + opts.selector).on('click', function() {
```
zannkukai, line 26: This should be non relevant if `this` doesn't have an `id`. 
Maybe better to use ``` this.find(opts.selected).on(....) ``` ?

Garfield-fr: I changed this with a new parameter.

rero_ils/static/js/rero_ils/rero_keep_tab.js
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+(function ( $ ) {
+  $.fn.keeptab = function( options ) {
+    var opts = $.extend( {
+      tabsName: '.nav-tabs',
+      localeItemName: 'navkeeptab',
+      selector: '.nav-link'
+    }, options );
+    $(opts.tabsName).find(opts.selector).on('click', function() {
```
zannkukai, line 27: I'm little bit confused with this notation without using `this`.
I think the code below should work too, but shouldn't ; isn't it ?
```
$('#dummy_id').keeptab({ localeItemName: 'document-tab'});
```
Why not simply use `$(this).find(opts.selector).on....` ?

rero_ils/static/js/rero_ils/rero_keep_tab.js
```diff
@@ -0,0 +1,46 @@
...
+      var elementId = undefined;
+      if ($(this).attr('id')) {
+        elementId = '#' + $(this).attr('id');
+      } else if ($(this).attr('href')) {
+        elementId = $(this).attr('href')
+      } else {
+        alert('Missing element id or href.');
```
jma, line 34: alert is a bit strong, no?

rero_ils/static/js/rero_ils/rero_keep_tab.js
Garfield-fr: If you have two sections of tabs on same page, you can limit the search area for storage

PR #1051, opened 2020-06-18
permissions: refactoring organisation permissions

zannkukai:
* Refactoring the permissions process using a permission factory and
permission classes for any resource instead of resources shared
functions. This way is the same way used by SONAR & Zenodo project
* Creates permission factory method
* Creates basic permission classes
* Create Organisation permission class
* Fixes unitests

## How to test?

- `poetry run ./run-tests.sh`

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?


rero_ils/modules/organisations/api.py
```diff
@@ -121,6 +90,28 @@ def get_record_by_viewcode(cls, viewcode):
...
 
+    @classmethod
+    def get_record_by_user(cls, user):
+        """Return organisation associated with patron.
+
+        :param user: the user to check.
+        :return: Organisation record or None.
+        """
+        from ..patrons.api import current_patron, Patron
+        patron = Patron.get_patron_by_user(user) or current_patron
```
jma, line 101: you can also write:
```python
return Patron.get_patron_by_user(user) or current_patron or None
```

rero_ils/modules/organisations/api.py
```diff
@@ -121,6 +90,28 @@ def get_record_by_viewcode(cls, viewcode):
...
+        if patron:
+            return patron.get_organisation()
+
+    @classmethod
+    def get_record_by_online_harvested_source(cls, source):
+        """Get record by online harvested source."""
```
jma, line 107: `source` is not described in the docstring.

rero_ils/modules/organisations/api.py
```diff
@@ -136,21 +127,48 @@ def online_circulation_category(self):
...
+            .filter('term', organisation__pid=self.pid)\
+            .source()\
+            .count()
+
+    def get_links_to_me(self):
+        """Get number of links."""
+        links = {}
+        libraries = self.get_number_of_libraries()
+        if libraries:
+            links['libraries'] = libraries
```
jma, line 161: `links` is useless this can be rewritten as:
```python
libraries = self.get_number_of_libraries()
if libraries:
    return dict(libraries=libraries)
```

rero_ils/modules/organisations/api.py
```diff
@@ -136,21 +127,48 @@ def online_circulation_category(self):
...
+        return links
+
+    def reasons_not_to_delete(self):
+        """Get reasons not to delete record."""
+        cannot_delete = {}
```
jma, line 166: same as previous

rero_ils/modules/organisations/permissions.py
```diff
@@ -0,0 +1,93 @@
...
+        # For staff users, they can read only their own organisation.
+        return current_organisation['pid'] == record['pid']
+
+    @classmethod
+    def create(cls, user, record=None):
+        """Create permission check.
+
+        :param user: Logged user.
+        :param record: Record to check.
+        :returns: True is action can be done.
```
jma, line 62: Please rename returns to return and check it on the entire project. Thanks.

rero_ils/modules/organisations/api.py
zannkukai: I don't think so because I have to return the organisation not the patron.

rero_ils/modules/organisations/api.py
zannkukai: In all cases, to not break the process, I need to return a dict (even i empty).
So the current solution, with only one return statement, seems as good as your proposal.

PR #1059, opened 2020-06-22
document: adapt the importation REST API to the UI

jma:
* Splits the document dumps function to be usable to the serializers.
* Renames `/api/import/bnf` to `/api/import_bnf` to be usable by the UI.
* Fixes year aggregation filter by forcing the terms to be a string.
* Returns empty result for empty query.
* Sorts aggregations by term counts.
* Creates a post_process function in JSON serializers.
* Creates two serializers: one without extra data for the editor one
  with extra data for the document views.
* Adds a new not found result error handler.
* Makes the tests more robust to BNF changes.

- https://tree.taiga.io/project/rero21-reroils/us/1274

## Dependencies

* rero/rero-ils-ui#286
* rero/ng-core#194

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/imports/serializers.py
+++ b/rero_ils/modules/imports/serializers.py
@@ -26,6 +26,10 @@ from invenio_records_rest.serializers.response import add_link_header, \
     search_responsify
 from marshmallow import fields
 
+from ..documents.api import Document
+from ..documents.dojson.contrib.unimarctojson import unimarc
+from ..documents.utils import title_format_text_head
+
 
 def record_responsify(serializer, mimetype):
     """Create a Records-REST response serializer.
@@ -74,14 +78,28 @@ class ImportsSearchSerializer(JSONSerializer):
         :param search_result: Elasticsearch search result.
         :param links: Dictionary of links to add to response.
         """
+        for hit in search_result['hits']['hits']:
+            hit['metadata'] = self.post_process(hit['metadata'])
+            hit['metadata']['pid'] = hit['id']
         results = dict(
             hits=dict(
                 hits=search_result['hits']['hits'],
                 total=search_result['hits']['total'],
+                remote_total=search_result['hits']['remote_total'],
             ),
             aggregations=search_result.get('aggregations', dict()),
         )
         return json.dumps(results, **self._format_args())
+
+    def post_process(self, metadata):
+        """Post process the data.
+
+        hook to be redefined in a child class.
+
+        :param metadata: dictionary version of a record
+        :return: the modified dictionary
+        """
+        return metadata
 
     def serialize(self, pid, record, links_factory=None, **kwargs):
         """Serialize a single record.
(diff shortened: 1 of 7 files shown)


rero_ils/modules/imports/serializers.py
```diff
@@ -83,14 +90,48 @@ def serialize_search(self, pid_fetcher, search_result, links=None,
         )
         return json.dumps(results, **self._format_args())
 
+    def post_process(self, metadata):
```
rerowep, line 93: Do we need this function, it is doing nothing?

JoelleDosimont:
Some remarks :
- Facet "date" : doesn’t include all the type of date (for example : "impr. 2008" -> 2008 will not be in the facet.
- The organization is mapped to a person in RERO ILS.
- Results for titles in non-roman alphabet doesn't show properly (search for "Vivliografia" for example).
- Some fuzzy results if you search "Mendeleiev". Some titles only present "@". But, it's seems to be the same in BnF (so I guess it's not a problem with the import).
- When I add an item to the newly create record, system goes crazy (but still creates the item if I refresh the page).
- To me, it's ok to have only the first 10 results, but maybe we can write "10 premiers résultats" in top of the results, and add "Afficher les 10 prochains résultats" ?
- Now that I see the results, I'm not sure the MARC tab will be fully use… The Intermarc is mostly unknown to the staff. When they will be used to JSON, many of them will lose there knowledge of MARC21 format. So I don't know what to do with it (maybe it's a good thing to have it, for the beginning).

jma:
> * [detail] the "MARC" tab should not appear just after import
> * it would be interesting to have on top of the brief view the total number or results instead of the number of results of the page. What is more, there is no pager in the bottom, so the user cannot deduce that there are more results.
> * The facet "date" should be sorted in alphabetical order
> * Some unknown errors happen sometimes, with message "Something bad happened; please try again later." Ex: request "peter weber"

All has been addresed.

jma:
> Some remarks :
> 
> * Facet "date" : doesn’t include all the type of date (for example : "impr. 2008" -> 2008 will not be in the facet.
Probably it is due to the transformation which is not a part of this PR.
> * The organization is mapped to a person in RERO ILS.
I do not understand.
> * Results for titles in non-roman alphabet doesn't show properly (search for "Vivliografia" for example).
Probably it is due to the transformation which is not a part of this PR.
> * Some fuzzy results if you search "Mendeleiev". Some titles only present "@". But, it's seems to be the same in BnF (so I guess it's not a problem with the import).
This is in the original data.
> * When I add an item to the newly create record, system goes crazy (but still creates the item if I refresh the page).
I was unable to reproduce this.
> * To me, it's ok to have only the first 10 results, but maybe we can write "10 premiers résultats" in top of the results, and add "Afficher les 10 prochains résultats" ?
Some enhancement has been done.
> * Now that I see the results, I'm not sure the MARC tab will be fully use… The Intermarc is mostly unknown to the staff. When they will be used to JSON, many of them will lose there knowledge of MARC21 format. So I don't know what to do with it (maybe it's a good thing to have it, for the beginning).
Take a decision with the PO group.

jma:
> You need to increase the code coverage

I tried to increase the coverage. It is not easy as it depends on the BNF data.

PR #1065, opened 2020-06-24
documentation: update the circulation actions file

pronguen:

- Simplifications according to improvement made by @BadrAly : items with pending loan is always IN_TRANSIT_FOR_PICKUP, instead of being sometimes IN_TRANSIT_TO_HOUSE


doc/circulation/actions.md
```diff
@@ -20,11 +20,7 @@
 1. :100: __ADD_REQUEST_4__: item in_transit (IN_TRANSIT_FOR_PICKUP)
     1. __ADD_REQUEST_4_1__: request patron = current loan patron → request denied
     1. __ADD_REQUEST_4_2__: request patron != current loan patron →  (add loan PENDING)
-1. :100: __ADD_REQUEST_5__: item in_transit (IN_TRANSIT_TO_HOUSE)
-    1. __ADD_REQUEST_5_1__: PENDING loan does not exist →  (add loan PENDING)
-    1. __ADD_REQUEST_5_2__: PENDING loan exists
-        1. __ADD_REQUEST_5_2_1__: request patron = current loan patron → request denied
-        1. __ADD_REQUEST_5_2_2__: request patron != current loan patron →  (add loan PENDING)
+1. :100: __ADD_REQUEST_5__: item in_transit (IN_TRANSIT_TO_HOUSE) →  (add loan IN_TRANSIT_FOR_PICKUP, item is: in_transit) [cancel previous loan]
```
BadrAly, line 23: Maybe it is simpler to have here `add loan pending` instead of `cancel loan + add loan IN_TRANSIT_FOR_PICKUP`.
because normally the item is in transit once it arrives at the library , the action `checkin` does the work for you.

doc/circulation/actions.md
```diff
@@ -72,17 +64,9 @@
...
-			1. __CHECKIN_5_2_1_1__: pickup location of first PENDING loan = item library →  (house_receive current loan, item is: at_desk(ITEM_AT_DESK))[automatic validate first PENDING loan]
-			1. __CHECKIN_5_2_1_2__: pickup location of first PENDING loan != item library →  (cancel current loan, item is: at_desk(ITEM_AT_DESK))[automatic validate first PENDING loan] *It is the case if the first PENDING loan has been deleted or unpriorised since last checkin*
-		1. __CHECKIN_5_2_2__: pickup location of first PENDING loan != transaction library
-			1. :white_check_mark: __CHECKIN_5_2_2_1__: pickup location of first PENDING loan = item library →  (no action, item is: in_transit (IN_TRANSIT_TO_HOUSE)) *Future: would be useful to track that an intermediary checkin has been made*
-			1. __CHECKIN_5_2_2_2__: pickup location of first PENDING loan != item library →  (checkin current loan, item is: in_transit (IN_TRANSIT_FOR_PICKUP))[automatic cancel current loan, automatic validate first PENDING loan] *It is the case if the first PENDING loan has been deleted or unpriorised since last checkin*
+1. __CHECKIN_5__: item in_transit (IN_TRANSIT_TO_HOUSE)
```
BadrAly, line 67: I do not think you need to remove the `IN_TRANSIT_TO_HOUSE` here. This is the main method that actually do the job for you. So it better to have it detailed here. Maybe you can have two texts:
- the details one to explain how it is coded
- the new one is actually for the librarian

pronguen: I thought that your code doesn't take into account the case of items `IN_TRANSIT_TO_HOUSE` with pending loans, since it is a case we want to avoid (see my previous response).

blankoworld: I think we need to talk about that. Because on my side even if specification doesn't talk about pending loan, the system can be in a state where pending loan exist during a IN_TRANSIT_xx exists and someone make a "checkout" of the item. It generates problems that I didn't resolve yet.

PR #1068, opened 2020-06-26
circulation: add tests on CHECKOUT action

blankoworld:
* Adds few tests for CHECKOUT on every item status
* Sets CHECKOUT actions as implemented in actions.md description file
* Fixes CHECKOUT_1_2_2 (checkout denied for loan with different patrons
on existing pending loan)
* Deactivates test_item_multiple_transit as it would be checked later

Because of [US1505, implement checkout actions](https://tree.taiga.io/project/rero21-reroils/task/1505?kanban-status=1224894).

## How to test?

```bash
poetry run bootstrap
poetry run run-tests
```

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/items/api/circulation.py
+++ b/rero_ils/modules/items/api/circulation.py
@@ -22,8 +22,8 @@ from functools import wraps
 
 from flask import current_app
 from invenio_circulation.api import get_loan_for_item
-from invenio_circulation.errors import MissingRequiredParameterError, \
-    NoValidTransitionAvailableError
+from invenio_circulation.errors import ItemNotAvailableError, \
+    MissingRequiredParameterError, NoValidTransitionAvailableError
 from invenio_circulation.proxies import current_circulation
 from invenio_circulation.search.api import search_by_patron_item_or_document, \
     search_by_pid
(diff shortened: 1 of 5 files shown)


rero_ils/modules/items/api/circulation.py
```diff
@@ -793,15 +793,34 @@ def prior_checkout_actions(self, action_params):
...
             loan = get_loan_for_item(item_pid_to_object(self.pid))
             if (loan and loan['state'] != LoanState.ITEM_AT_DESK):
                 item, cancel_actions = self.cancel_loan(pid=loan.get('pid'))
                 actions.update(cancel_actions)
+        # CHECKOUT_1_2_2: checkout denied if some pending loan are linked to it
+        # with different patrons
+        # Except while coming from an ITEM_IN_TRANSIT_TO_HOUSE loan because
+        # the loan is cancelled and then came up in ON_SHELF to be checkout
+        # by the second patron.
+        if self.status == ItemStatus.ON_SHELF and \
```
BadrAly, line 810: I think here, you do not need to check:
`loan['state'] != LoanState.ITEM_IN_TRANSIT_TO_HOUSE` because you will never have an item `in_transit` and `on_shelf` at same time.

If I were you, I replace it with `loan['state'] == LoanState.PENDING`

rero_ils/modules/items/api/circulation.py
```diff
@@ -793,15 +793,34 @@ def prior_checkout_actions(self, action_params):
...
+                loan['state'] != LoanState.ITEM_IN_TRANSIT_TO_HOUSE:
+            item_pid = item_pid_to_object(self.pid)
```
BadrAly, line 812: you can replace the following lines by using the method:
`self.get_item_loans_by_state(state= LoanState.PENDING)`

rero_ils/modules/items/api/circulation.py
```diff
@@ -793,15 +793,34 @@ def prior_checkout_actions(self, action_params):
...
+            search = search_by_pid(
+                item_pid=item_pid,
+                filter_states=[LoanState.PENDING])
+            search_result = search.execute()
+            for result in search_result:
+                if result.patron_pid != loan.get('patron_pid'):
+                    pending = Loan.get_record_by_pid(result.pid)
```
BadrAly, line 819: You do not need to fetch the loan, because `get_item_loans_by_state` will return the loans  for you

rero_ils/modules/items/api/circulation.py
blankoworld: During the process of a checkout of a ITEM_IN_TRANSIT_TO_HOUSE, a loan is canceled. And the ITEM become ON_SHELF. That's why I add this condition.
I will try what you suggest and will come back to you about the result.

blankoworld: Tests CHECKOUT_1_2_2 didn't pass with `loan['state'] == LoanState.PENDING` :disappointed:

BadrAly: I got it. 
if the loan is cancelled, then you do not need to check its state at all:
`if self.status == ItemStatus.ON_SHELF:`
and then you continue to compare  the  `pending.patron_pid` with `loan.pid`
we can take a look at it together if you want

blankoworld: After the loan is cancelled, the variable is not updated as it becomes `None`. And then the checkout cannot pass. And so the item didn't becomes ON_LOAN.
I will add a comment `TODO` to check this problem later.

rero_ils/modules/items/api/circulation.py
```diff
@@ -849,15 +849,33 @@ def prior_checkout_actions(self, action_params):
...
+        # CHECKOUT_1_2_2: checkout denied if some pending loan are linked to it
+        # with different patrons
+        # Except while coming from an ITEM_IN_TRANSIT_TO_HOUSE loan because
+        # the loan is cancelled and then came up in ON_SHELF to be checkout
+        # by the second patron.
+        if self.status == ItemStatus.ON_SHELF and \
+                loan['state'] != LoanState.ITEM_IN_TRANSIT_TO_HOUSE:
+            for res in self.get_item_loans_by_state(state=LoanState.PENDING):
+                if res.patron_pid != loan.get('patron_pid'):
+                    pending = Loan.get_record_by_pid(res.pid)
```
BadrAly, line 873: since `res` is a `loan`, then you can remove the line 873:

`pending = Loan.get_record_by_pid(res.pid)`

and replace 
`pending.patron_pid` by `res.patron_pid` in line 876

Issue #1003, opened 2020-05-22
editor : multiple provision activity lost when editing a document


zannkukai:
**Describe the issue**
When editing a document with multiple provision activity, only the last one will be appears into the editor; this cause a metadata lost.

**To Reproduce**
1. Log as librarian
1. Import or create a document with multiple provision activity (ex : import EAN 9782298133103)
1. Correct editor (removing "place" part of provisionActivity[@type=Manufacture])
1. Save it
1. The created document should have 3 provisions activities : two "bf:Publication" parts and one "bf:Manufacture" part
1. Edit this document : both "bf:Publication" parts are disappear.

**Expected behavior**
All existing provisionActivity parts should be displayed.

**Software version**
Tell us which version you're using:

  * rero-ils : fe0a8c3
  * rero-ils-ui : e3248a9

AoNoOokami:
I saw the same with subjects while working on the UX of editor.

zannkukai:
Note to solve (to myself or anybody): 

The problem seems to be present into the `array_type.component.ts`. When a "hidden" change is detected on a field, then all children of this field will be removed (except the lastOne if `minItems=1` into JSON Schema). 
https://github.com/rero/ng-core/blob/dev/projects/rero/ng-core/src/lib/record/editor/array-type/array-type.component.ts#L58
The question is now : "Why an 'hide' event is send for these elements"

PR #1069, opened 2020-06-29
branch for the data model series user story

AoNoOokami:

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?


pronguen:
Here the comment of my review and the review of @PascalRepond 
* Title of host document must be indexed in the child document; it is not.
* Possible values of `issuance.subtype` must depend on `issuance.main_type`; they are independent.
* `seriesStatement.seriesTitle` should have no description; it is useless
* Placeholders for `seriesEnumeration` and `subseriesEnumeration` are missing.
* When editing a document with a series statement, the editor opens by default the subseries statement, which makes it impossible to save the document.
	* The subseries statement should always be added manually (except when it already contains data), as it is quite rare.
	* By default, fields `seriesTitle` and `seriesEnumeration` should be displayed inside `seriesStatement`.
* Format validation for `partOf`
	* when the data entered in `numbering.pages` does not respects the pattern, there should be a validation message: "Should be in the following format: 135 or 5-27". Same for `year`
	* `numbering.volume` has an integer format, but it is possible to write a string without receiving any validation message. When we save, the value is not saved. There should be a validation message and it should be impossible to save the form.
	* `numbering.issue` has no validation; any value can be entered and is saved. There should be a validation message and it should be impossible to save the form.
* The main type of `modeOfIssuance` is displayed as a code in the pro detailed view; there should be a human translation.

Can be done later:
* If various series statement exist, only one is kept when editing the document (same than for provision activity, https://github.com/rero/rero-ils/issues/1003). This can be done in the next *editor* US.
* Ideally, the editor proposes a small widget, for field `partOf` to look for another document and link it (as for the authors). This can be done in another US.

AoNoOokami:
Here is the issue opened for the validation of integer-type fields: https://github.com/ngx-formly/ngx-formly/issues/2360

AoNoOokami:
>     * Is it possible to set a default value the subtype of modeOfIssuance? Ideally one for each main type.

Normally, the editor HAS default values, in create and edit mode.
> 
>     * Regarding the creation of link to host document (773):
>       
>       * all the articles that I tested don't have this link. Example: R222472560
>       * we should be able to deduce this link in most cases. See link in [Explore](http://data.rero.ch/01-R222472560/html): there is only one candidate in this case; did we link it in RERO ILS? In ilsdev.test.rero.ch, it is not the case.

I need more info. Is it absent from data, view (public, admin)? @reropag, @rerowep, you may be involved in the resolution of this point.

> 
>     * The point _"fields `seriesTitle` and `seriesEnumeration` should be displayed inside `seriesStatement`"_ has not been addressed. It can be done in the US 1542.
> 
>     * Besides the validation problems (issue opened at ngx-formly), the validation messages are wrong (for year) and not well displayed (for pages), see below.
>       image
>       
>       * year format should be AAAA
> 
>     * The search input to add a new field doesn't work anymore.

PR #1079, opened 2020-07-02
circulation: ensure request validation from ui

BadrAly:
Replaces the api call /validate by /validate_request.

Users have now the choice to give one of the transaction_location_pid
or transaction_library_pid when placing a request.

The validate action of item request is now possible from the item views.
The list of required parameters are clearly noted for frontend calls in
the actions.md and api_views.

* Creates additional units testing for frontend calls.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/items/api/circulation.py
+++ b/rero_ils/modules/items/api/circulation.py
@@ -47,33 +47,33 @@ from ....filter import format_date_filter
 
 
 def add_action_parameters_and_flush_indexes(function):
-    """Add missing action parameters and validate parameters.
+    """Add missing action and validate parameters.
 
-    This method will replce add_loans_parameters_and_flush_indexes
+    This method will replace add_loans_parameters_and_flush_indexes
     """
     @wraps(function)
     def wrapper(item, *args, **kwargs):
         """Executed before loan action."""
         # TODO: this code will be enhanced while adding the other actions.
-        from . import ItemsSearch
-        # the smart checkin requires extra checks and actions before a checkin
-        if function.__name__ == 'checkin':
+        checkin_loan = None
+        if function.__name__ == 'validate_request':
+            # checks if the given loan pid can be validated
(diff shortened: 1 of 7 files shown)


rero_ils/modules/items/api/circulation.py
```diff
@@ -56,6 +56,10 @@ def wrapper(item, *args, **kwargs):
         """Executed before loan action."""
         # TODO: this code will be enhanced while adding the other actions.
         from . import ItemsSearch
+        if function.__name__ == 'validate_request':
+            # checks if the given loan pid can be validated
+            item.is_action_validate_request_possible(**kwargs)
+
         # the smart checkin requires extra checks and actions before a checkin
         if function.__name__ == 'checkin':
             loan, kwargs = item.prior_checkin_actions(item, **kwargs)
```
blankoworld, line 65: Why not using `elif`? I think there is more optimization in code from Python. And we cannot be in `validate_request` and `checkin` at the same time. So `elif` is acceptable.

rero_ils/modules/items/api/circulation.py
```diff
@@ -185,6 +185,30 @@ class ItemCirculation(IlsRecord):
         LoanState.ITEM_IN_TRANSIT_TO_HOUSE: 'in_transit',
     }
 
+    def process_item_and_loans(self, item):
+        """Commit and reindex the item and its loans."""
```
blankoworld, line 189: L.73 it's not easy to understand what you do with this method unless you read the code.
The title of method just say "I do something". And the comment say: "I commit changes and reindex items and loans".

What about something like `commit_and_reindex` or `commit_and_reindex_item_and_loan`? Or `commit_and_reindex_with_loans`?

PR #1084, opened 2020-07-03
circulation: ensure the item extend loan action for frontend calls

blankoworld:
* adds new prior_extend_loan_actions method used before extend_loan
action
* adds new test_actions_views_extend_loan_request test file
* uses do_jsonify_action for extend_loan action

Because of [US1580 about extend_loan action in API views](https://tree.taiga.io/project/rero21-reroils/task/1580)

## How to test?

Just check that Travis tests are OK

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/items/api/circulation.py
+++ b/rero_ils/modules/items/api/circulation.py
@@ -61,7 +61,7 @@ def add_action_parameters_and_flush_indexes(function):
             item.is_action_validate_request_possible(**kwargs)
         elif function.__name__ == 'checkin':
             # the smart checkin requires extra checks/actions before a checkin
-            loan, kwargs = item.prior_checkin_actions(item, **kwargs)
+            loan, kwargs = item.prior_checkin_actions(**kwargs)
             checkin_loan = loan
         # CHECKOUT: Case where no loan PID
         elif function.__name__ == 'checkout' and not kwargs.get('pid'):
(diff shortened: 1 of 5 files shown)


rero_ils/modules/items/api/circulation.py
```diff
@@ -215,31 +218,57 @@ def is_action_validate_request_possible(self, **kwargs):
...
+        loan_pid = kwargs.get('pid')
+        if not loan_pid:
+            loan = self.get_first_loan_by_state(LoanState.ITEM_ON_LOAN)
+        else:
+            loan = Loan.get_record_by_pid(loan_pid)
+
+        # Check extend is allowed
+        checked_out = True  # we consider loan as checked-out
+        have_request = False
+        if not loan_pid and not loan:
```
BadrAly, line 232: since the `loan_pid` is checked before then only:
`if not loan`

rero_ils/modules/items/api/circulation.py
```diff
@@ -215,31 +218,57 @@ def is_action_validate_request_possible(self, **kwargs):
...
+            # no loand_pid, so we check first_loan_by_state, without result:
+            # item was not checked out
+            checked_out = False
+        if LoanState.PENDING in self.get_loan_states_for_an_item():
```
BadrAly, line 236: `elif` ?

rero_ils/modules/items/api/circulation.py
blankoworld: I want to be in the only case where loan_pid was not given *and* that `get_first_loan_by_state()` returns None.

You're right, I need to change a little bit the code.

rero_ils/modules/items/api/circulation.py
blankoworld: Not needed anymore as I moved previous `if` in another block.

PR #1098, opened 2020-07-17
check_license: accept Triple Slash Directive

blankoworld:
run-tests.sh checks that Licenses are OK for some files.
This scripts fails when JS files use a Triple Slash Directive.
This commit change this.

* Improves check_license method to include Triple-Slash directives for
.js files (Cf.
https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html)
* Avoids checking screenshots directory in Cypress
* Adds a triple slash directive on 2 JS files (from Cypress).

Because Cypress tests needs to be done on RERO ILS and I need to extend developer environment to do so.

## How to test?

`poetry run run-tests` shouldn't give any error on Cypress JS files (that contains Triple Slash Directive)

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/cli.py
+++ b/rero_ils/modules/cli.py
@@ -26,6 +26,7 @@ import json
 import logging
 import multiprocessing
 import os
+import re
 import shutil
 import sys
 import traceback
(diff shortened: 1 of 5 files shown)


rero_ils/modules/cli.py
```diff
@@ -448,6 +449,12 @@ def test_file(file_name, extensions, extension, license_lines,
         # No error found
         return 0
 
+    def is_slash_directive(file, line):
+        is_js_file = file.name.split('.')[-1] == 'js'
+        if is_js_file and re.search(triple_slash, line):
+            return True
+        return False
```
lauren-d, line 456: cosmetic: can be replaced by:
```
if file.name.split('.')[-1] == 'js' and re.search(triple_slash, line):
   return True
```

PR #1105, opened 2020-07-20
loans: item deletion allowed with CREATED loan

blankoworld:
Sometimes items have Loan in CREATED state. Mainly due to circulation
action that are interrupted.
These Loan are not considered as important for the Circulation. But the
delete button on Item cannot be active if CREATED Loan exists.
This commit fixes the problem by bypassing CREATED Loan in item delete
button.

* Fixes bug about Item that have Loan in state CREATED and cannot be
deleted.

Because you cannot delete an item that have CREATED Loan.

## How to test?

* Attempt a checkout on an item not allowed by policy (an error should appear)
* Check that you have CREATED Loan here: http://localhost:9200/loans/_search/?q=CREATED
* Go to the item and check that the "delete" button is possible
* Click on it.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.
- [ ] Extracted translations?

### DIFF
--- a/rero_ils/modules/items/api/circulation.py
+++ b/rero_ils/modules/items/api/circulation.py
@@ -1151,10 +1151,14 @@ class ItemCirculation(IlsRecord):
         }
 
     def get_number_of_loans(self):
-        """Get number of loans."""
+        """Get number of loans.
+
+        Exclude CREATED Loan as it can block Item deletion.
+        """
         search = search_by_pid(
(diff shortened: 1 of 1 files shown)


blankoworld:
> For the commit title : `loans: item deletion allowd with CREATED loan`

@iGormilhit it's now fixed!

PR #1111, opened 2020-07-23
reports: create inventory list based on item resource

rerowep:
This part implements backend to perform inventory list reports in CSV format.

* Adds CSVResponse with filename.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/items/serializers/response.py
+++ b/rero_ils/modules/items/serializers/response.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+#
+# RERO ILS
+# Copyright (C) 2019 RERO
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, version 3 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


rero_ils/modules/items/serializers/response.py
```diff
@@ -0,0 +1,63 @@
...
+        response = current_app.response_class(
+            serializer.serialize_search(pid_fetcher, search_result,
+                                        links=links,
+                                        item_links_factory=item_links_factory),
+            mimetype=mimetype)
+        response.status_code = code
+        if headers is not None:
+            response.headers.extend(headers)
+        file_name = '{date}-inventory.csv'.format(
+            date=datetime.now().strftime('%Y%m%d')
```
blankoworld, line 50: In RERO ILS I think we talk about using aware date with timezone (in case we work in a different timezone).

Do you think it's possible to use the timezone from the library in which we do the inventory?

For now we use this method for the timezone date: https://github.com/rero/rero-ils/blob/dev/rero_ils/modules/libraries/api.py#L317

Info: the timezone can change the date. For example if the server is in Switzerland, and you make an inventory in New York at 6'30PM, the date wouldn't be the same.

PR #1194, opened 2020-09-02
document: replace role label content with values

iGormilhit:

To fix the agent relation role of the contribution field, in order to get the
correct translations

## How to test?

- Check the changed files, verify that the codes are consistent. 

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/documents/jsonschemas/documents/document_contribution_role-v0.0.1.json
+++ b/rero_ils/modules/documents/jsonschemas/documents/document_contribution_role-v0.0.1.json
@@ -161,595 +161,595 @@
     "options": [
       {
         "value": "aut",
-        "label": "author"
+        "label": "aut"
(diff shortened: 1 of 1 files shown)


rerowep (approved):
Would it be better to prefix the values with `contributor_`?

zannkukai (approved):

iGormilhit:
@rerowep 

> Would it be better to prefix the values with `contributor_`?

1. If we prefix the value, then we'll have to prefix the label too, and the items of the enumeration. Otherwise, label and value won't be the same anymore, and we will have to translate both.
1. We already discussed it with @Garfield-fr and @pronguen and decided to wait until a prefix is necessary.

So, I'm going to merge it like it is.

PR #1204, opened 2020-09-08
fixtures: add an organisation for cypress

AoNoOokami:
* Adds fixtures for cypress tests (organisation, libraries, librarians,
patrons, locations, patron type, item type and circ policy). This leverages
 the creation of resources needed to test circulation scenarios

Add fixtures dedicated to cypress tests.

## Dependencies

N/A

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/organisations/api.py
+++ b/rero_ils/modules/organisations/api.py
@@ -173,6 +173,12 @@ class Organisation(IlsRecord):
             cannot_delete['links'] = links
         return cannot_delete
 
+    def is_test_organisation(self):
+        """Check if this is a test organisation."""
+        if self.get('code') == 'cypress':
+            return True
+        return False
+
 
 class OrganisationsIndexer(IlsRecordsIndexer):
     """Holdings indexing class."""
--- a/rero_ils/templates/rero_ils/frontpage.html
+++ b/rero_ils/templates/rero_ils/frontpage.html
@@ -125,7 +125,9 @@
           <h3>{{ _('Organisations\' catalogues') }}</h3>
           <ul class="list-unstyled">
             {% for org in organisations %}
-            <li><a href="{{ url_for('rero_ils.index_with_view_code', viewcode=org.code) }}">{{ org.name }}</a></li>
+              {% if  not org.is_test_organisation() %}
+                <li><a href="{{ url_for('rero_ils.index_with_view_code', viewcode=org.code) }}">{{ org.name }}</a></li>
+              {% endif %}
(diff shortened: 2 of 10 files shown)


rero_ils/templates/rero_ils/frontpage.html
```diff
@@ -125,7 +125,9 @@ <h3>Il progetto</h3>
           <h3>{{ _('Organisations\' catalogues') }}</h3>
           <ul class="list-unstyled">
             {% for org in organisations %}
-            <li><a href="{{ url_for('rero_ils.index_with_view_code', viewcode=org.code) }}">{{ org.name }}</a></li>
+              {% if org.code != 'cypress' %}
```
zannkukai, line 128: I'm not ultra fan about this string check here. Maybe it will be better to call a function on `organisation` to know if the organisation is a test one. Something like this in the template :
``` {% if !org.is_test_organisation %} ```
and in `Organisation` class :
```
...
@property
def is_test_organisation(self):
    if self.code == 'cypress':
        return False
    if self.other_test_could_be_placed_here():
        return False
    return True
```

**Review by zannkukai — APPROVED** [review]

**Review by iGormilhit — APPROVED** [review]
Commit message approved.

File: rero_ils/templates/rero_ils/frontpage.html (hunk 1 above)
Comment (AoNoOokami) [reply rero_ils/templates/rero_ils/frontpage.html]: Done

File: rero_ils/modules/organisations/api.py [hunk 2]
```diff
@@ -173,6 +173,12 @@ def reasons_not_to_delete(self):
             cannot_delete['links'] = links
         return cannot_delete
 
+    def is_test_organisation(self):
+        """Check if this is a test organisation."""
+        if self.code == 'cypress':
+            return False
+        return True
+
```
zannkukai, line 181: It's just the opposite :-) "Cypress" is a test organisation

zannkukai (changes requested):

rero_ils/modules/organisations/api.py
BadrAly: +1

BadrAly (approved):

rero_ils/modules/organisations/api.py
AoNoOokami: When you follow suggestions without thinking.... Thanks.

zannkukai (approved):

PR #1218, opened 2020-09-14
user: sign in with a username

jma:
* Adds a username to all fixtures.
* Adds a new invenio accounts login view REST API.
* Adds a js code to the login view to be able to be logged using the
  username or the email.
* Renames the login (/signin) and logout (/signout) url to avoid
  conflict with the accounts REST API.
* Changes the link between the patron and invenio user link from email
  to id.
* Synchronize the data between the patron and user profile in both
  direction.
* Sets the default user password as the birth date.

- https://tree.taiga.io/project/rero21-reroils/us/1590

## How to test?

- Bootstrap/Setup
- Login with a librarian and create a new patron.
- The new patron can be logged using a username.
- Change the data in the patron and professional view the data should be modified in both direction.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


rero_ils/accounts_views.py
```diff
@@ -0,0 +1,68 @@
...
+    try:
+        # print('======================>', email)
+        profile = UserProfile.get_by_username(email)
+        # email = profile.user.email
+    except NoResultFound:
+        if not current_datastore.get_user(email):
+            raise ValidationError('USER_DOES_NOT_EXIST')
+
+current_datastore = LocalProxy(
+    lambda: current_app.extensions['security'].datastore)
```
zannkukai, line 42: As other LocalProxy of the poject, these lines could be moved at the top of the file, just after imports

rero_ils/accounts_views.py
```diff
@@ -0,0 +1,68 @@
...
+from werkzeug.local import LocalProxy
+
+
+#
+# Field validators
+#
+def user_exists(email):
+    """Validate that a user exists."""
+    try:
+        # print('======================>', email)
```
zannkukai, line 34: COSMETIC : can be removed

rero_ils/accounts_views.py
```diff
@@ -0,0 +1,68 @@
...
+        profile = UserProfile.get_by_username(email)
+        # email = profile.user.email
```
zannkukai, line 36: email is never reuse --> can be removed

rero_ils/modules/ext.py
```diff
@@ -71,7 +72,7 @@ def __init__(self, app=None):
             app.add_template_filter(can_edit, name='can_edit')
             app.add_template_filter(
                 can_access_item,
-                name='can_access_item'
+                name='can_access_itetest_patron_organisation_pidm'
```
zannkukai, line 75: it seems `test_patron_organisation` string has been placed into 'ite' and 'm'. Bad copy/paste ?

rero_ils/modules/patrons/api.py
```diff
@@ -190,20 +227,65 @@ def delete(self, force=False, delindex=False):
...
+        # retrieve the user
+        patron = Patron.get_patron_by_user(profile.user)
+        if patron:
+            for field in cls.profile_fields:
+                # date field requires conversion
+                if field == 'birth_date':
+                    patron[field] = getattr(
+                        profile, field).strftime('%Y-%m-%d')
+                else:
+                    patron[field] = getattr(profile, field)
```
zannkukai, line 245: Suggestion/Idea (optional) :: Is it possible that `cls.profile_fields` return a dict where key are the field name and value could be the function to get the patron.field value ? I will easy to add a new field to synchronize in the future (if needed)

rero_ils/modules/patrons/api.py
```diff
@@ -190,20 +227,65 @@ def delete(self, force=False, delindex=False):
...
+                user = User.query.filter_by(id=data.get('user_id')).first()
+            # update all common fields
+            for field in cls.profile_fields:
+                # date field need conversion
+                if field == 'birth_date':
+                    setattr(
+                        user.profile, field,
+                        datetime.strptime(data.get(field), '%Y-%m-%d'))
+                else:
+                    setattr(user.profile, field, data.get(field, ''))
```
zannkukai, line 279: same as previous

rero_ils/views.py
```diff
@@ -368,3 +387,46 @@ def schemaform(document_type):
...
+            pass
+        return current_datastore.get_user(email)
+
+    @use_kwargs(post_args)
+    def post(self, **kwargs):
+        """Verify and login a user."""
+        user = self.get_user(**kwargs)
+        self.verify_login(user, **kwargs)
+        self.login_user(user)
+        return self.success_response(user)
```
zannkukai, line 432: Same code in `rero_ils/account_views.py`. Keep only one

zannkukai (changes requested):
2 request changes : first for the `test_patron...` string mistake. Second for the duplicate code. Other are more cosmetic

rero_ils/modules/patrons/api.py
jma: Thanks for the suggestion. This would be a good idea, but seems a bit overkill as we need to define a set and get function for each field. An other solution is to create a new children Profile class with a getter and setter. Again it seems a bit overkill for only one field.

zannkukai (approved):

lauren-d (approved):

Issue #1305, opened 2020-10-15
Labels of the circulation policy editor shoud be improved.


pronguen:
**How it works**

The labels of the circulation policy are not understandable:
* First reminder days (below *Send first reminder*).
* Days before due date (below *Sending how many days after due date*).
* Fees (EUR) (below *Fees fo this notification [currency]*).

**Improvement suggestion**

We should have

A section "Send a loan expiry notice" (with toggle). If activated:
* "Sending how many days before the due date"

A section "Send a first reminder" (with toggle). If activated:
* "Sending how many days after the due date"
* Fees for this notification ([currency])

See mockup on https://tree.taiga.io/project/rero21-reroils/us/1759

The following sections will be added through a user story:

A section "Send a second reminder" (with toggle). If activated:
* "Sending how many days after the due date"
* Fees for this notification ([currency])

A section "Send a third reminder" (with toggle). If activated:
* "Sending how many days after the due date"
* Fees for this notification ([currency])

A section "Send an invoice" (with toggle). If activated:
* "Sending how many days after the due date"

**Context**

Proposed by pilot libraries

PR #1354, opened 2020-11-02
circulation: hide buttons if action isn't available.

zannkukai:
If the patron can't operate an action due to circulation restrictions
(patron blocked, patron type limits, ...), circulation buttons should be
hide. In the patron profile view, adds warning/error messages depending
of patron restrictions.

Closes rero/rero-ils#1357
Closes rero/rero-ils#1356

## How to test?

image

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/patrons/templates/rero_ils/_patron_profile_requests.html
+++ b/rero_ils/modules/patrons/templates/rero_ils/_patron_profile_requests.html
@@ -49,7 +49,7 @@
         <div class="col-lg-2{% if loan.state == 'ITEM_AT_DESK' %} text-success{% endif %}">
           {% if loan.state == 'ITEM_AT_DESK' %}
           <i class="fa fa-check"  title="{{ _('item at desk') }}" aria-hidden="true"></i>
-          {{ _('ready') }}
+          {{ _('to pick up') }}
           {% elif loan.state in ['PENDING', 'ITEM_IN_TRANSIT_FOR_PICKUP'] %}
           <i class="fa fa-bullseye"  title="{{ _('pending') }}" aria-hidden="true"></i>
           <!-- TODO: Add expired date and remove "waiting" text
--- a/rero_ils/modules/documents/templates/rero_ils/_documents_get.html
+++ b/rero_ils/modules/documents/templates/rero_ils/_documents_get.html
@@ -207,27 +207,38 @@
             {%- endfor %}
 
             <!-- action button : Should be at the end to be render above and be clickabke -->
-            {%- set locations = item|item_library_pickup_locations %}
-            {%- if item|can_request and locations %}
-              <div class="action-buttons">
-                <a href="#" type="button" class="btn btn-primary btn-sm" data-toggle="dropdown" aria-haspopup="true"
-                  aria-expanded="false" id="{{ item.barcode }}-dropdownMenu">
-                  {{ _('Request') }}
-                  <i class="fa fa-caret-down fa-fw"></i>
-                </a>
-                <!-- TODO: Style the dropdown header -->
-                <div class="dropdown-menu dropdown-menu-right" aria-labelledby="dropdownMenu">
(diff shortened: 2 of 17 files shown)


Garfield-fr (approved):

rero_ils/modules/patrons/templates/rero_ils/_patron_profile_requests.html
```diff
@@ -49,7 +49,7 @@
         <div class="col-lg-2{% if loan.state == 'ITEM_AT_DESK' %} text-success{% endif %}">
           {% if loan.state == 'ITEM_AT_DESK' %}
           <i class="fa fa-check"  title="{{ _('item at desk') }}" aria-hidden="true"></i>
-          {{ _('ready') }}
+          {{ _('to pick uup') }}
```
lauren-d, line 52: change `to pick uup` to `to pick up` 🙃?

lauren-d (changes requested):

ManaDeweerdt:
When I try to checkout an item I get an error 'item not found'.
But I can checkin and checkout an item already borrowed.

ManaDeweerdt:
Like the https://github.com/rero/rero-ils/issues/1357 the button 'request' should be visible even he has the limits reached, followed by an explanation.

benerken:
Great job Renaud ! 
some comments for now:

- If we define a general checkout limit only => the system displays a message when you arrive on the patron account in the professional view  AND  It displays a flash message when you do a checkout  => OK
(A permanent message is displayed in the patron public view) => OK

- If we define a general checkout limit AND a library checkout limit => the system doesn't display a permanent message when you arrive on the patron account in the professional view. It displays a flash message when you do a checkout. Why ? 
(Same behaviour in the patron public view => No permanent message)

iGormilhit (changes requested):
Commit message body: replace twice *disable* by *disabled*. Thanks.

zannkukai:
> Like the #1357 the button 'request' should be visible even he has the limits reached, followed by an explanation. [enhancement]

It's more difficult :: One of the main reason why an item is not requestable is because the item is not in the same organisation than patron. In this case, it's not user friendly to get the button disabled with this message.

lauren-d (approved):

pronguen:
* Wording
  * The unit of the field for each limit should be clearer in the field label (same problem than in cipo, https://github.com/rero/rero-ils/issues/1305).
  * I don't understand the word *default* in some of the labels.
  * The label should correspond to the description given below and does not need to repeat the info given in the toggle.
  * Suggestions:
    * Global checkout limit > Number of checkouts in the organisation
    * Default checkout limit by library > Number of checkouts by library
    * Default fee amount limit > Fee amount
    * Default overdue items limit > Number of overdue items
image
* Message in public interface (for patrons)
  * Don't display any message above, in the patron account (simplification), as they are note user-friendlyI think it is enough to display the message on the disabled buttons as tooltip.
    * Potential problem: the user does not know that he cannot request document. But if we display the request button in any case (sometimes disabled) it will solve this. See also Mana's comment [can be done later, issue can be opened].
  * Tooltips on buttons (request/renew): adapt if possible to end user, as specified in the US: "Renewal impossible: you have too many overdue items." / "Renewal impossible: you have too many fees." / etc.
  * Tootips in Google Chrome (Version 86.0.4240.183 (Build officiel) (64 bits)) is problematic: no tooltip for the first item to renew, other tooltip permanent (see screenshot)
  * This works on Firefox
image
* Circulation interface
  * Simplification of the flash message: the first part could be used as first line. In example below, use "Checkout denied" as first line.
image
  * [enhancement] Ideally, the messages display above in user checkout account are adapted. Now they are answer messages after an action tried.
* General
  * Limit by fee amount includes only overdue fees? Or it just does not include subscription fees? To decide, we may need to adapt the labels and messages.

Issue #1373, opened 2020-11-04
In the pro patron account, some paid fees still appear.


pronguen:
**Describe the bug**

In the pro interface, patron account, tab Fees, some paid fees still appear.

**To Reproduce**
1. [Diagnostic wanted, it was impossible to reproduce]

**Expected behavior**

The paid fees should appear only in the fee history.

**Context**
* server: ilspilot.test.rero.ch
* version: v0.12.0

Proposed by the pilot libraries

**Screenshots**

![image](https://user-images.githubusercontent.com/8154915/98135031-acb40d80-1ebf-11eb-9d07-425947942056.png)

iGormilhit:
@zannkukai It reminds me something, such as a PR that you proposed. Does it ring a bell to you? :bell:

zannkukai:
@iGormilhit it seems not be linked. Subscriptions creation and transactions management come from 2 distinct modules.

zannkukai:
For dark reasons some payment are not corresponding to the total amount. 
So the total due is not equals to 0 and the system consider this transaction as Open (no problem on backend).
For this transaction the total due is still "4.996003610813204e-16" CHF (you can see it, clicking on "pay all" action button

![image](https://user-images.githubusercontent.com/10031585/98222347-ff89d580-1f50-11eb-9b69-54a1ef057a5e.png)

The problem is now : how to know how this strange amount is has been placed ?

![image](https://user-images.githubusercontent.com/10031585/98222508-35c75500-1f51-11eb-9ca1-cbe42e05f1eb.png)

Simple solution : click on "pay all" to create an new payment event and closed the transaction.

pronguen:
@zannkukai
The problem was probably that a payment was done partially
* Example: fee CHF 0.20
* Payment: CHF 0.199
* Remaining fee: CHF 0.00 (rounded)

Solution: prevent to pay an amount with 3 digits after comma? Would it be possible?

zannkukai:
Problem found ! It's due to bit reprsentation of real number in python 
```
$ python
Python 2.7.12 (default, Jul 21 2020, 15:19:50) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> 10 - 9.54
0.46000000000000085
>>> 
```

PR #1404, opened 2020-11-05
transactions: fix paid fees still opened

zannkukai:
The bit representation of real number
(https://en.wikipedia.org/wiki/IEEE_754) cause problem with arithmetic
operation on them (10-9.54 = 0.460000000000085). This commit fixes this
problem by multiplying transaction amount by 100 and cast them into
integer when arithmetic operations are done.

Closes rero/rero-ils#1373

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/patron_transaction_events/api.py
+++ b/rero_ils/modules/patron_transaction_events/api.py
@@ -128,16 +128,25 @@ class PatronTransactionEvent(IlsRecord):
 
     def update_parent_patron_transaction(self):
         """Update parent patron transaction amount and status."""
+        # NOTE :
+        #   due to bit representation of float number
+        #   (https://en.wikipedia.org/wiki/IEEE_754), the arithmetic operation
+        #   with float can cause some strange behavior
+        #     >>> 10 - 9.54
+        #     0.46000000000000085
+        #   To solve this problem in our case, as we keep only 2 decimal
+        #   digits, we can multiply amounts by 100, cast result as integer,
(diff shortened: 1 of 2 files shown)


rero_ils/modules/patron_transaction_events/api.py
```diff
@@ -128,16 +128,25 @@ def create_event_from_patron_transaction(
...
+        #   with float can cause some strange behavior
+        #     >>> 10 - 9.54
+        #     0.46000000000000085
+        #   To solve this problem in our case, as we keep only 2 decimal
+        #   digits, we can multiply amounts by 100, cast result as integer,
+        #   do operation with these values, and (at the end) divide the result
+        #   by 100.
         patron_transaction = self.patron_transaction()
-        total_amount = patron_transaction.get('total_amount')
+        total_amount = int(patron_transaction.get('total_amount') * 100)
```
BadrAly, line 142: My understanding is the `int` removes the decimal part of the value. is this working as expected?

zannkukai: exactly ! According with PO, we only keep 2 decimals. So multiply by 100 and cast to int do the works (and divide at the end)

BadrAly (approved):

iGormilhit (approved):
Commit message approved.

lauren-d (approved):

PR #1538, opened 2020-12-02
items: inherit holdings first call_number

### LINKED ISSUE #1288: The issue call number should be generated according to the holding call number.
**How it works**

A call number (1st and 2nd) exists in the holding. But the issues are created without a call number.

**Improvement suggestion**

If a call number (1st and 2nd) exists in the holding, it is used
* as is, when receiving the issue
* as is, when editing the issue, so that it can be completed
* this should be the case for the 1st and 2nd call numbers

**Context**

Proposed by pilot libraries

BadrAly:
For items of type `issue`, the parent holdings first call_number is
displayed when the item has no first call_number. This call number is
inherited in the following screens in the public interface:

* Document details view.
* Patron profile: loans, requests, history tabs.
* Collection details view.
*  Late issues and inventory CSV exports.
* Generated notifications.

* Closes #1288.

https://tree.taiga.io/project/rero21-reroils/task/1896?kanban-status=1224894

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/collections/api.py
+++ b/rero_ils/modules/collections/api.py
@@ -70,8 +70,15 @@ class Collection(IlsRecord):
         :param self: self
         :return: list of items linked to collection
         """
-        return [Item.get_record_by_pid(item['pid']) for
-                item in self.replace_refs().get('items', [])]
+        items = []
+        for item in self.replace_refs().get('items', []):
+            item = Item.get_record_by_pid(item['pid'])
+            # inherit holdings call number when possible
+            issue_call_number = item.issue_inherited_first_call_number
+            if issue_call_number:
+                item['call_number'] = issue_call_number
+            items.append(item)
+        return items
 
 
 class CollectionsIndexer(IlsRecordsIndexer):
--- a/rero_ils/modules/items/api/record.py
+++ b/rero_ils/modules/items/api/record.py
@@ -361,6 +361,18 @@ class ItemRecord(IlsRecord):
         return circulation_category_pid
 
     @property
+    def issue_inherited_first_call_number(self):
+        """Get issue inherited first call number.
+
+        For item of type issue, when the issue first call number is missing,
+        it returns the parent holdings first call number if exists.
+        """
+        from ...holdings.api import Holding
+        if self.get('type') == 'issue' and not self.get('call_number'):
+            return Holding.get_record_by_pid(
+                self.holding_pid).get('call_number')
+
+    @property
     def location_pid(self):
         """Shortcut for item location pid."""
         location_pid = None
--- a/rero_ils/modules/holdings/api.py
+++ b/rero_ils/modules/holdings/api.py
@@ -361,6 +361,10 @@ class Holding(IlsRecord):
             item = Item.get_record_by_pid(item_pid)
             if not item.issue_status or \
                     item.issue_status == ItemIssueStatus.RECEIVED:
+                # inherit holdings first call# for issues with no 1st call#.
+                issue_call_number = item.issue_inherited_first_call_number
+                if issue_call_number:
+                    item['call_number'] = issue_call_number
                 yield item
 
     def get_number_of_items(self):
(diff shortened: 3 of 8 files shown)


iGormilhit (commented):
For the commit message title, *items* is better than *item*. But it's not so important.

Garfield-fr (approved):

rero_ils/modules/items/api/record.py
```diff
@@ -353,6 +353,17 @@ def holding_circulation_category_pid(self):
...
+    def item_inherited_first_call_number(self):
+        """Return holdings first call number if item 1st call# is missing."""
+        from ...holdings.api import Holding
+        holding = Holding.get_record_by_pid(self.holding_pid)
+        if (
+            self.get('type') == 'issue' and
+            not self.get('call_number') and
+            holding.get('call_number')
+        ):
+            return holding.get('call_number')
```
zannkukai, line 365: In this case, I suggest to split the `if` statement in 2 to avoid loading the holding if it's not necessary (best perf).
```
    if (self.get('type') == 'issue' and not self.get('call_number')):
        return Holding.get_record_by_pid(self.holding_pid).get('call_number')
```
the `if holding.get('call_number'):` can be skipped because in both case the function will return 'None' (a function without specific return statement return None by default)

More I read this function, more I think it could be a `call_number item @properety`. If it could then previous change could be removed.
```
# inherit holdings first call# for issues with no 1st call#.
inherited_call_number = item.item_inherited_first_call_number()
if inherited_call_number:
    item['call_number'] = inherited_call_number
```
==> a call number for an issue without specific call_number is :
  - None if parent holding doesn't have a call_number
  - The parent holding call_number

zannkukai (changes requested):

rero_ils/modules/collections/api.py
```diff
@@ -70,8 +70,14 @@ def get_items(self):
...
         :return: list of items linked to collection
         """
-        return [Item.get_record_by_pid(item['pid']) for
-                item in self.replace_refs().get('items', [])]
+        items = []
+        for item in self.replace_refs().get('items', []):
+            item = Item.get_record_by_pid(item['pid'])
+            # inherit holdings call number when possible
+            if item.issue_inherited_first_call_number:
+                item['call_number'] = item.issue_inherited_first_call_number
```
zannkukai, line 78: Using this, you call two times the property and so potentially load two times the holdings (perf).
But, I don't think you need to check if `item.issue_inherit_first_call_number` is none, you can remove the if statement and directly assign `item.issue_inherit_first_call_number` to `item['call_number']`.

Same remark for others files.

zannkukai (changes requested):
Sorry for this second "requrest changes" :-)

iGormilhit (changes requested):
Commit message proposition:

```
items: inherit holdings first call_number

For items of type `issue`, the parent holdings first call_number is
displayed when the item has no first call_number. This call number is
inherited in the following screens in the public interface:

* Document details view.
* Patron profile: loans, requests, history tabs.
* Collection details view.
* Generated notifications.

* Closes #1288.
```

zannkukai (approved):

rero_ils/modules/items/api/record.py
```diff
@@ -355,6 +355,18 @@ def holding_circulation_category_pid(self):
                     self.holding_pid).circulation_category_pid
         return circulation_category_pid
 
+    @property
+    def issue_inherited_first_call_number(self):
+        """Get issue inherited first call number.
+
+        For item of type issue, when the issue first call number is missing,
+        it returns the parent holdings first call number. if exists.
```
AoNoOokami, line 363: Suggestion: ` it returns the parent holdings first call number if exists.` (remove `.` after number)

AoNoOokami (approved):

iGormilhit (approved):
`* Closes #1288.` With a dot. It's a sentence.

lauren-d (approved):

zannkukai (approved):

PR #1593, opened 2021-01-08
documents: implement document type

rerowep:

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


benerken:
In the professional view, logged as Jean:

- when I try to edit the document 815 (docsubtype_other_book), I get an error 500. I don't know if it's linked to this US (?)
image

- Same problem with an article : document 540.
- When I creates a new document and I want to save it "As a template", I get an error 500.

- Inventory List: When I export an list in CSV, the column "Document Type" contains main type and subtype at the same time. It could be improved to split value in 2 columns (not blocking but enhancement). 
For now, it's boring to use the list with values like this
`[{'subtype': 'docsubtype_other_book', 'main_type': 'docmaintype_book'}, {'subtype': 'docsubtype_manga', 'main_type': 'docmaintype_comic'}]`
--

pronguen:
- [ ] For Cantook e-books, a tab "Online access" is now displayed. It was not the case before: why? I think it is not necessary.
- [ ] In the editor:
   - [ ] The subtype is optional. It should be required (if the corresponding main type has subtypes).
   - [ ] If it is easy to set: put main type and subtype on the same line to save vertical space.
- [ ] Is it normal that the temporary circulation category is already deployed? The field is mandatory in the item editor...
- [ ] I cannot save templates anymore. The button works but has no effect.

See also Benoit's comments

AoNoOokami:
>     * [ ]  Is it normal that the temporary circulation category is already deployed? The field is mandatory in the item editor...

Ping @zannkukai  I think.

ThierryClavel:
the maintype comic should have 2 subtypes : comic and manga : in the following exemple there is manga and other book https://ilsdev.test.rero.ch/global/search/documents?q=&page=1&size=10

PR #1605, opened 2021-01-13
gh actions: set the stale workflow

iGormilhit:
* Adds a workflow for GitHub actions that mark issues and PR as stale
  when they are open with no activites for 6 months. No PR or issue will
  be closed (maybe in a future move).
* Some labels are excepted.

- To implement a github action to marke issue and PR as stale. 

## How to test?

- Check the changed files
- **Question: should we try it with the debug, dry run, function?**

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,26 @@
+name: "Mark stale issues and pull requests"
+on:
+  workflow_dispatch:
+  schedule:
+  - cron: "30 1 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        # Make issue/PR as stale after 6 months
+        days-before-stale: 180
+        # Do not close issue/PR marked as stale (will be changed later)
+        days-before-close: 0
+        stale-issue-message: 'This issue is stale because it has been open 6 months with no activity.'
+        stale-pr-message: 'This PR is stale because it has been open 6 months with no activity.'
+        # Set the label added to issue marked as stale (leave default value)
+        # stale-issue-label: 'stale'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,28 @@
...
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        # Make issue/PR as stale after 6 months
+        days-before-stale: 180
+        # Do not close issue/PR marked as stale (will be changed later)
+        days-before-close: 0
+        stale-issue-message: 'This issue is stale because it has been open 6 months with no activity.'
+        stale-pr-message: 'This PR is stale because it has been open 6 months with no activity.'
+        # Set the label added to issue marked as stale (leave default value)
+        # stale-issue-label: 'stale'
```
sebastiendeleze, line 23: Why not setting a label to the issues?

sebastiendeleze (approved):

.github/workflows/stale.yml
iGormilhit: A label is set by default to the issue, that's why it's commented here. But there's no default label for PR.

jma (approved):

.github/workflows/stale.yml
```diff
@@ -0,0 +1,28 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "30 1 * * *"
```
rerowep, line 5: Should we add `- workflow_dispatch` for the possibility to manually trigger the execution ?

iGormilhit: @rerowep Why not. Can we have both? Through a cron *and* manually triggered?

iGormilhit: I've made the change, base on this example: https://github.com/actions/stale/blob/main/.github/workflows/stale.yml

PR #1638, opened 2021-01-25
document: document type

rerowep:
* Adds `cached` decorator and uses it to cache covers.

* Adds subtype subfacet for documents.

* Adapts SIP2 types.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


iGormilhit (approved):
Commit message approved.

rerowep:
I fixed the templates and the csv export.

AoNoOokami:
Comments made by @benerken :
In the professional view, logged as Jean:

- [ ] when I try to VIEW the document 815 (docsubtype_other_book), I get an error 500. I don't know if it's linked to this US (?)
image
- [ ] Same problem with an article : document 540.
- [ ] - [ ] When I list the requests, some lines are not loaded. The little spinner loops and the console returns an error 500.
This is related to error 500 mentioned above.
- [x] When I create a new document (with type/ subtype) and try to save it "As a template", I get an error 500.

- [ ] Inventory List: When I export a list in CSV, the column "Document Type" contains main type and subtype at the same time. It could be improved to split value in 2 columns (not blocking but enhancement). 
For now, it's boring to use the list with values like this
`[{'subtype': 'docsubtype_other_book', 'main_type': 'docmaintype_book'}, {'subtype': 'docsubtype_manga', 'main_type': 'docmaintype_comic'}]`
--

- [ ] Inventory List: (depending of the amount of work needed) is it possible to add a new facet for DocType/SubType ?

AoNoOokami:
Comments by @pronguen :
- [x] For Cantook e-books, a tab "Online access" is now displayed. It was not the case before: why? I think it is not necessary.
- [ ] In the editor:
   - [ ] The subtype is optional. It should be required (if the corresponding main type has subtypes).
   - [ ] If it is easy to set: put main type and subtype on the same line to save vertical space.
- [x] Is it normal that the temporary circulation category is already deployed? The field is mandatory in the item editor...
- [x] I cannot save templates anymore. The button works but has no effect.

See also Benoit's comments

JoelleDosimont:
There is a problem with the facets when assigning more than one subtype. All the subtypes appears on the maintypes (even if the subtype is not linked to the maintype) : 
image
The subtype "documentary" is linked to "movies/series", but appears also in the "book" maintype. 
Don't know if it's normal ?

pronguen:
@JoelleDosimont concerning the facet problem, we know it but it would require quite an effort to solve it properly. This is apparently the standard functioning of Elasticsearch. So we decided to start like this and correct it later, if it is set as a priority. Now it is already operational.

rero_ils/modules/documents/utils.py
```diff
@@ -507,14 +507,58 @@ def create_contributions(contributions):
...
+            '&value={isbn}'.format(
+                cover_service=cover_service,
+                isbn=isbn
+            )
+        try:
+            host_url = flask_request.host_url
+        except:
+            host_url = current_app.config.get('RERO_ILS_APP_URL', '??')
+            if host_url[-1] != '/':
+                host_url = '{host_url}/'.format(host_url=host_url)
```
zannkukai, line 529: https://www.python.org/dev/peps/pep-0498/#:~:text=F%2Dstrings%20provide%20a%20way,which%20contains%20expressions%20inside%20braces.

With python 3.6, we can skip the ".format" and use simply `f'{host_url}/'`

rero_ils/modules/documents/views.py
```diff
@@ -393,24 +395,41 @@ def create_publication_statement(provision_activity):
...
+        isbn_cover = get_remote_cover(isbn)
+        if isbn_cover and isbn_cover.get('success'):
+            url = isbn_cover.get('image')
+            if save_cover_url:
+                pid = record.get('pid')
+                record_db = Document.get_record_by_pid(pid)
+                record_db.add_cover_url(url=url, dbcommit=True, reindex=True)
+                msg = 'Add cover art url: {url} do document: {pid}'.format(
+                    url=url,
+                    pid=pid
```
zannkukai, line 428: same remark concerning string format (use simply `f()`)

rero_ils/modules/documents/utils.py
zannkukai, line 529: https://www.python.org/dev/peps/pep-0498/#:~:text=F%2Dstrings%20provide%20a%20way,which%20contains%20expressions%20inside%20braces.

With python 3.6, we can skip the ".format" and use simply `f'{host_url}/'`

rero_ils/modules/documents/views.py
```diff
@@ -574,3 +593,50 @@ def in_collection(item_pid):
...
+
+    :param record: record
+    :param translate: translate document type
+    :return: dictonary list of document types
+    """
+    doc_types = record.document_types
+    if translate:
+        for idx, doc_type in enumerate(doc_types):
+            doc_types[idx] = _(doc_type)
+    return doc_types
```
zannkukai, line 610:
```python
if translate:
  return [_(doc_type) for doc_type in record.document_types]
else:
  return record.document_types 
```
If only Python has an easy-readable ternary instruction :'(

rero_ils/modules/imports/api.py
```diff
@@ -119,9 +119,29 @@ def calculate_aggregations_add(self, type, data, id):
...
+
+    def calculate_aggregations_add_sub(self, type, data, sub_type, sub_data,
+                                       id):
+        """Add data to aggregations_creation.
+
+        :param type: type of the aggregation
+        :param data: data for the aggregation
+        :param sub_type: type of the aggregation
+        :param sub_data: data for the aggregation
+
```
zannkukai, line 135: unnecessary blank line

zannkukai (approved):

rero_ils/modules/documents/views.py
rerowep: Maybe we can change all the format() into f() ?
I will no change it here, because it is writen mostly like this.

rero_ils/modules/documents/views.py
rerowep: `return [_(doc_type) for doc_type in record.document_types] if translate else record.document_types`

rero_ils/modules/imports/api.py
rerowep: changed

lauren-d (approved):

pronguen:
New problem - not blocking:
1. Select a main type in the facet
2. Select another main type in the facet
3. See that the results are not updated/filtered according to the 2nd main type. Only the 1st is considered.

Why?

PR #1657, opened 2021-02-01
documents: fix form options label/value

iGormilhit:
* Fixes the JSON schema, in the form options, when the label has a value
  in natural language instead of the code (as in the value of the value
  key) as specified in the guidelines¹.
- Replaces the title value that is in natural language by it's code in
  some oneOf sections of the document JSON schema, in order to avoid
  translating both the code and the natural language version. Removes
  the corresponding useless strings from the manual translations file.

¹ https://github.com/rero/developer-resources/blob/master/data/jsonschema.md#enum

- Some strings have to be translated twice, because the label value is not like
  the value value. :wink:
- In *some* JSON oneOf sections, the title value should be the code.

## How to test?

- Check changed files. 

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


rerowep (approved):

rero_ils/modules/documents/jsonschemas/documents/document_issuance-v0.0.1.json
```diff
@@ -189,7 +189,7 @@
         }
       },
       {
-        "title": "Integrating resource",
+        "title": "rdami:1004",
```
jma, line 192: I understand why but looks like ugly as the JSONSchema is also a data model documentation.

rero_ils/modules/documents/jsonschemas/documents/document_production_method-v0.0.1.json
```diff
@@ -33,83 +33,83 @@
         "type": "selectWithSort",
         "options": [
           {
-            "label": "Blueline process",
+            "label": "rdapm:1001",
```
jma, line 36: same here and the followers

jma (approved):

PR #1659, opened 2021-02-01
exceptions: correct raise statements

BadrAly:
* Adapts units testing.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/holdings/api_views.py
+++ b/rero_ils/modules/holdings/api_views.py
@@ -57,7 +57,7 @@ def jsonify_error(func):
             return jsonify({'status': 'error: {error}'.format(
                 error=error)}), 400
         except Exception as error:
-            raise(error)
+            raise Exception(error)
             current_app.logger.error(str(error))
             return jsonify({'status': 'error: {error}'.format(
                 error=error)}), 500
--- a/rero_ils/modules/items/api_views.py
+++ b/rero_ils/modules/items/api_views.py
@@ -71,7 +71,7 @@ def jsonify_error(func):
         except NotFound as error:
             raise(error)
         except Exception as error:
-            # raise(error)
+            # raise Exception(error)
             current_app.logger.error(str(error))
             return jsonify({'status': 'error: {error}'.format(
                 error=error)}), 500
--- a/tests/api/holdings/test_patterns.py
+++ b/tests/api/holdings/test_patterns.py
@@ -142,20 +142,22 @@ def test_receive_regular_issue_api(
     # librarian of another library are not authoritzed to receive issues
     # for another library.
     login_user_via_session(client, librarian_fully_no_email.user)
-    res, data = postdata(
-        client,
-        'api_holding.receive_regular_issue',
-        url_data=dict(holding_pid=holding.pid)
-    )
-    assert res.status_code == 401
+    with pytest.raises(Exception):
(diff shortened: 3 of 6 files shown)


rero_ils/modules/holdings/api_views.py
```diff
@@ -57,7 +57,7 @@ def decorated_view(*args, **kwargs):
             return jsonify({'status': 'error: {error}'.format(
                 error=error)}), 400
         except Exception as error:
-            raise(error)
+            raise Exception(error)
```
BadrAly, line 60: @zannkukai does this break something for the `quick recieve` in the frontend?

rero_ils/modules/items/api_views.py
```diff
@@ -71,7 +71,7 @@ def decorated_view(*args, **kwargs):
         except NotFound as error:
             raise(error)
         except Exception as error:
-            # raise(error)
+            # raise Exception(error)
```
iGormilhit, line 74: If it's commented, should it be removed?

iGormilhit (commented):
I would prefer `python: correct raise statements`.

rero_ils/modules/holdings/api_views.py
zannkukai: Nope, it shouldn't

rero_ils/modules/items/api_views.py
zannkukai: No :-) it's always the place for debugging

tests/api/holdings/test_patterns.py
```diff
@@ -142,20 +142,22 @@ def test_receive_regular_issue_api(
...
-        url_data=dict(holding_pid=holding.pid)
-    )
-    assert res.status_code == 401
+    with pytest.raises(Exception):
+        res, data = postdata(
+            client,
+            'api_holding.receive_regular_issue',
+            url_data=dict(holding_pid=holding.pid)
+        )
+        assert res.status_code == 401
```
zannkukai, line 151: Are you sure the function return à 401 ? 
As I read the test, the `with` statement is validated because `postdata` raise an `Exception` and the status_code isn't check. For me it's not the good. The API call to 'receive_regular_issue' should not raise any Exception and ony return 401 (So the previous test is correct)

tests/api/holdings/test_patterns.py
```diff
@@ -142,20 +142,22 @@ def test_receive_regular_issue_api(
...
```
zannkukai, line 160: same

tests/ui/patrons/test_patrons_api.py
```diff
@@ -47,7 +45,7 @@ def test_patron_create(app, roles, lib_martigny, librarian_martigny_data_tmp,
 
     wrong_librarian_martigny_data_tmp = deepcopy(librarian_martigny_data_tmp)
     wrong_librarian_martigny_data_tmp.pop('first_name')
-    with pytest.raises(ValidationError):
+    with pytest.raises(Exception):
```
zannkukai, line 48: Same here. If an Exception is raised during the `create` call (during validation process). It should be catched and re-raise a `ValidationError`

zannkukai (changes requested):

PR #1668, opened 2021-02-02
circulation: improve response time for circulation operations

### LINKED ISSUE #1600: Improve response time for circulation operations
In the professional view: since release **0.15**, circulation operations like checkout/checkin are too slow. It takes around 5 seconds to do a checkout. 
It's the same problem when you validate a request. It takes 3,5 sec. for each validation.

**How it works**

It's perhaps linked to new controls and validations added recently (limits, blocks, etc...)

**Improvement suggestion**

Try to improve the response time for circulation operations. 
We need to think about staff at the desk :+1: 

Add a spinner to display activity to the staff ( Linked to  issue https://github.com/rero/rero-ils/issues/1298 )

rerowep:
* closes #1600.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/api.py
+++ b/rero_ils/modules/api.py
@@ -21,6 +21,7 @@
 from copy import deepcopy
 from uuid import uuid4
 
+import click
 import pytz
 from celery import current_app as current_celery_app
 from elasticsearch import VERSION as ES_VERSION
@@ -44,6 +45,7 @@ from sqlalchemy import text
 from sqlalchemy.orm.exc import NoResultFound
 
 from .errors import RecordValidationError
+from .utils import extracted_data_from_ref
 
 
 class IlsRecordError:
@@ -109,7 +111,7 @@ class IlsRecord(Record):
                     cls.provider.pid_type
                 ]['indexer_class']
             )
-        except:
+        except Exception:
             # provide default indexer if no indexer is defined in config.
             indexer = IlsRecordsIndexer
         return indexer
@@ -186,23 +188,28 @@ class IlsRecord(Record):
         return record
 
     @classmethod
-    def get_record_by_pid(cls, pid, with_deleted=False):
+    def get_record_by_pid(cls, pid, with_deleted=False, verbose=False):
         """Get ils record by pid value."""
-        assert cls.provider
-        try:
-            persistent_identifier = PersistentIdentifier.get(
-                cls.provider.pid_type,
-                pid
-            )
-            return super().get_record(
-                persistent_identifier.object_uuid,
-                with_deleted=with_deleted
-            )
-        # TODO: is it better to raise a error or to return None?
-        except NoResultFound:
-            return None
-        except PIDDoesNotExistError:
-            return None
+        if verbose:
+            click.echo('\t\tget_record_by_pid: {name} {pid}'.format(
+                name=cls.__name__,
+                pid=pid
+            ))
+        if pid:
+            assert cls.provider
+            try:
+                persistent_identifier = PersistentIdentifier.get(
+                    cls.provider.pid_type,
+                    pid
+                )
+                record = super().get_record(
+                    persistent_identifier.object_uuid,
+                    with_deleted=with_deleted
+                )
+                return record
+            # TODO: is it better to raise a error or to return None?
+            except (NoResultFound, PIDDoesNotExistError):
+                return None
 
     @classmethod
     def record_pid_exists(cls, pid):
@@ -219,9 +226,7 @@ class IlsRecord(Record):
             )
             return True
 
-        except NoResultFound:
-            return False
-        except PIDDoesNotExistError:
+        except (NoResultFound, PIDDoesNotExistError):
             return False
 
     @classmethod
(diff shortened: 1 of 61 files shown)


rero_ils/modules/api.py
```diff
@@ -185,6 +186,7 @@ def create(cls, data, id_=None, delete_pid=False,
         return record
 
     @classmethod
+    @memoized(timeout=3)
```
zannkukai, line 189: I'm not fan about cached a record for 3 seconds. A lot of thing can happens during this delay on an object. But if
the cached is removed/replace if the object is update/delete/reindex, it could ba a good solution but with a lot of test. 

Probably we need to implement a performance test (using JMeter ?) to benchmark this "performance issue"

rero_ils/modules/circ_policies/api.py
```diff
@@ -265,6 +264,7 @@ def allow_request(cls, item, **kwargs):
         library_pid = kwargs['library'].pid if kwargs.get('library') \
             else item.library_pid
         cipo = cls.provide_circ_policy(
+            item.organisation_pid,
```
zannkukai, line 267: to be tested : Is the `replace_refs()` call (behind the `organisation_pid` property) has better performance than a single DB call if the object has multiple references inside.

rero_ils/modules/api.py
rerowep: Still a draft and not ready jet.

rero_ils/modules/vendors/api.py
```diff
@@ -76,9 +76,8 @@ def order_email(self):
         vendor = self.replace_refs()
         if vendor:
             return vendor.get(
-                'order_contact', vendor.get(
-                    'default_contact', {})).get('email')
-        return None
+                'order_contact', vendor.get('default_contact', {})
+            ).get('email')
```
zannkukai, line 80: what a strange construction... the `replace_refs` seems completely unnecessary.
```python
@property
def order_email(self):
    """Shortcut for vendor order email."""
    return self.get('order_contact', self.get('default_contact', {})).get('email')
```

zannkukai (approved):

rero_ils/modules/documents/api.py
```diff
@@ -207,17 +207,20 @@ def index_contributions(self, bulk=False):
...
                     contributions_ids.append(contrib.id)
                 else:
                     contrib.reindex()
         if contributions_ids:
-            IlsRecordsIndexer().bulk_index(
-                contributions_ids, doc_type=['cont'])
+            from ..contributions.api import ContributionsIndexer
+            indexer = ContributionsIndexer()
+            indexer.bulk_index(contributions_ids)
+            indexer.process_bulk_queue()
```
zannkukai, line 223: After some discuss with @lauren-d this afternoon, this syntax seems synchronous ; if we're right, there is no benefits using this syntax. But calling `process_bulk_queue.apply_async()` from `rero_ils.modules.tasks`, the process become async (but you need that Celery is up and running).

Will test again tomorrow using this syntax

lauren-d (approved):

rero_ils/modules/documents/api.py
rerowep: I think you are right. I have one problem with asynchronous indexing. Sometimes the indexing is not done and I have not yet found out why.

rero_ils/modules/api.py
```diff
@@ -185,23 +187,30 @@ def create(cls, data, id_=None, delete_pid=False,
...
+                record = super().get_record(
+                    persistent_identifier.object_uuid,
+                    with_deleted=with_deleted
+                )
+                return record
+            # TODO: is it better to raise a error or to return None?
+            except NoResultFound:
+                return None
+            except PIDDoesNotExistError:
+                return None
```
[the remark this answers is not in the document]
rerowep: The correct syntax is:
```
except (NoResultFound, PIDDoesNotExistError):
    return None
```

rero_ils/modules/items/api/api.py
```diff
@@ -229,33 +230,35 @@ class ItemsIndexer(IlsRecordsIndexer):
...
-        current_search.flush_and_refresh(DocumentsSearch.Meta.index)
+        # reindex document in background
+        document_pid = extracted_data_from_ref(record.get('document'))
+        uid = Document.get_id_by_pid(document_pid)
+        from ...documents.api import DocumentsIndexer
+        DocumentsIndexer().bulk_index([uid])
+        from rero_ils.modules.tasks import process_bulk_queue
+        process_bulk_queue.delay()
+        # document = Document.get_record_by_pid(document_pid)
+        # document.reindex()
```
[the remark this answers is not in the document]
rerowep: Changed to:
```python
        uid = Document.get_id_by_pid(document_pid)
        DocumentsIndexer().index_by_id(uid)
```

rero_ils/modules/vendors/api.py
rerowep: Yes indeed, very strange.

rero_ils/modules/documents/api.py
rerowep: Rewritten.

PR #1680, opened 2021-02-05
public ui: better document type facet

rerowep:

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


rero_ils/modules/documents/serializers.py
```diff
@@ -30,6 +30,92 @@
...
+    },
+    'docmaintype_other': {},
+    'docmaintype_score': {
+        'docsubtype_printed_score': True,
+        'docsubtype_music_method': True,
+        'docsubtype_manuscript_score': True
+    },
+    'docmaintype_serial': {},
+    'docmaintype_series': {}
+}
```
zannkukai, line 98: Cosmetic and philosophic : Is this kind of configuration shouldn't be placed into a `config.py` file inside the module ?

rero_ils/modules/documents/serializers.py
```diff
@@ -30,6 +30,92 @@
...
+        main_type = type_bucket['key']
+        new_subtype_buckets = []
+        subtype_buckets = type_bucket['document_subtype']['buckets']
+        for subtype_bucket in subtype_buckets:
+            if DOCUMENT_TYPES.get(main_type, {}).get(subtype_bucket['key']):
+                new_subtype_buckets.append(subtype_bucket)
+        new_type_bucket[
+            'document_subtype'
+        ]['buckets'] = new_subtype_buckets
+        print(main_type, new_subtype_buckets)
```
zannkukai, line 115: remove this line

rero_ils/modules/documents/serializers.py
```diff
@@ -30,6 +30,92 @@
...
+        new_type_buckets.append(new_type_bucket)
+    return new_type_buckets
+
```
zannkukai, line 118: Maybe interesting to think how to "remove" unwanted key from `buckets` arguments instead of recreate a new complete buckets ? In this case the L110-L111 should be something like 
```
for subtype_idx, subtype_bucket in enumerate(subtype_buckets):
    if subtype_bucket['key'] not in DOCUMENT_TYPES.get(main_type, {}):
        del buckets[bucket_idx]['document_subtype']['buckets'][subtype_idx]
```
I'm not sure about the bucket structure, you know it better than me.

For me it will increase the function readability.

rero_ils/modules/documents/serializers.py
```diff
@@ -30,6 +30,92 @@
...
+        'docsubtype_printed_score': True,
+        'docsubtype_music_method': True,
+        'docsubtype_manuscript_score': True
+    },
+    'docmaintype_serial': {},
+    'docmaintype_series': {}
+}
+
+
+def filter_document_type_buckets(buckets):
```
zannkukai, line 101: It seems a kind of private function. Python best practice are to prepend function name with "_" in this case

rero_ils/modules/documents/serializers.py
```diff
@@ -133,6 +219,27 @@ def post_process_serialize_search(self, results, pid_fetcher):
...
+        #     subtype_buckets = type_bucket['document_subtype']['buckets']
+        #     for subtype_bucket in subtype_buckets:
+        #         if DOCUMENT_TYPES.get(maintype, {}).get(subtype_bucket['key']):
+        #             new_subtype_buckets.append(subtype_bucket)
+        #     new_type_bucket[
+        #         'document_subtype'
+        #     ]['buckets'] = new_subtype_buckets
+        #     print(maintype, new_subtype_buckets)
+        #     new_type_buckets.append(new_type_bucket)
+        # results['aggregations']['document_type']['buckets'] = new_type_buckets
```
zannkukai, line 238: Can be removed

rero_ils/modules/documents/serializers.py
rerowep: Yes, but I don't know where.

zannkukai: My thinking (should be discuss with others) : create a `config.py` file into the `rero_ils.modules.documents`. Use an `from .config import DOCUMENT_TYPES` where you need this configuration.

BadrAly (approved):

iGormilhit (changes requested):
What does it fix? What was broken?

rero_ils/modules/documents/serializers.py
rerowep: It also used in item serializer and maybe in import serializer. I have put everythin into utils.py.

rero_ils/modules/documents/serializers.py
rerowep: I put it into utils.py.

rerowep:
> What does it fix? What was broken?

Actually nothing was broken, the function of the document type was just not understandable.

zannkukai (approved):

rero_ils/modules/documents/utils.py
```diff
@@ -30,6 +30,91 @@
 from .dojson.contrib.marc21tojson.model import remove_trailing_punctuation
 from ...utils import get_i18n_supported_languages
 
+DOCUMENT_TYPES = {
```
jma, line 33: This can be added to the config.py file, no?

jma (commented):
the commit message is quite poor. Can you please add more details?

rero_ils/modules/documents/utils.py
rerowep: Could be, but do we really like to put document facet correction specific definitions in the config?

rero_ils/modules/utils.py
```diff
@@ -50,6 +50,17 @@ def wrapper(*args, **kwargs):
     return caching
 
 
+def memoized(timeout=50):
```
jma, line 53: Please correct the doc string. Can you add more details on it and mention this in the commit message. Thanks.

jma (commented):
Do you need to add some tests? Perhaps it is already covered by the current tests?

jma (approved):

rero_ils/modules/utils.py
rerowep: ok

lauren-d (approved):

PR #1702, opened 2021-02-16
circulation policies: update json schema for Formly

Garfield-fr:

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/circ_policies/jsonschemas/circ_policies/circ_policy-v0.0.1.json
+++ b/rero_ils/modules/circ_policies/jsonschemas/circ_policies/circ_policy-v0.0.1.json
@@ -9,7 +9,23 @@
     "pid",
     "name",
     "organisation",
-    "is_default"
+    "is_default",
+    "allow_requests",
+    "policy_library_level"
+  ],
+  "propertiesOrder": [
+    "name",
+    "description",
+    "checkout_duration",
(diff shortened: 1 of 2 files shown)


rero_ils/modules/circ_policies/jsonschemas/circ_policies/circ_policy-v0.0.1.json
```diff
@@ -9,7 +9,23 @@
     "pid",
     "name",
     "organisation",
-    "is_default"
+    "is_default",
+    "allow_requests",
+    "policy_library_level"
```
zannkukai, line 14: why these field are now required ? 
I can understand about `allow_request`. But for `policy_library_level`, I think that if there is no information, this field will not be present into the json schema.

Additionally, I think that the EditorComponent from ng-core, clear the model before sending data to the API. If it detects that an object/field/array is empty, it removes this entry from the model (maybe same way for boolean field ?). So the API will not received any data and as field is marked as required, the json validation will failed.

**EDIT:**
If mix up `setting` and `policy_library_level`. But It should be test if EditorComponent doesn't remove entry for False boolean value. I have a big doubt..

zannkukai (approved):

AoNoOokami (approved):

PR #1711, opened 2021-02-19
transactions: optimise incremental creation

zannkukai:
Creating an event for each incremental step was a poor idea (regarding
performance issues). Indeed, for old loan with a lot of incremental fee,
we created a PatronTransactionEvent and reindex it. This process take a
lot of time. To Solve this problem, we create only one event with a set
of step. So only one PatronTransactionEvent is created ; and only one
reindex is neceesary.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/babel.ini
+++ b/babel.ini
@@ -35,4 +35,4 @@ extract_messages = $._, jQuery._
 
 # Extraction from json files (schema, form)
 [json: **.json]
-keys_to_translate = ['^title$', '^label$', 'description', 'placeholder', 'name', 'add', '403', '.*Message']
+keys_to_translate = ['^title$', '^label$', 'description', 'placeholder', 'name', 'add', '403', '.*Message', '^text$']
--- a/data/circulation_policies.json
+++ b/data/circulation_policies.json
@@ -12,14 +12,14 @@
       {
         "type": "due_soon",
         "days_delay": 5,
-        "communication_channel": "email",
+        "communication_channel": "patron_setting",
         "template": "email/due_soon"
       },
       {
         "type": "overdue",
         "days_delay": 5,
         "fee_amount": 2.0,
-        "communication_channel": "email",
+        "communication_channel": "patron_setting",
         "template": "email/overdue"
       }
     ],
@@ -41,14 +41,14 @@
       {
         "type": "due_soon",
         "days_delay": 5,
-        "communication_channel": "email",
+        "communication_channel": "patron_setting",
         "template": "email/due_soon"
       },
       {
         "type": "overdue",
         "days_delay": 5,
         "fee_amount": 2.0,
-        "communication_channel": "email",
+        "communication_channel": "patron_setting",
         "template": "email/overdue"
       }
     ],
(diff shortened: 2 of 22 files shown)


benerken (approved):

BadrAly (approved):

pronguen (commented):
- [x] In the previous editor, some fields were conditional
   - [x] `number_renewals` should be displayed only if `checkout_duration`>0
   - [x] `renewal_duration` should be displayed only if `number_renewals`>0
- [x] `is_default` should not be editable; this allows to set more than one default policy. Only one policy per organisation is set as default.
- [ ] For a library level policy, I should be able to add more than one library.
- [x] The communication channel for each reminder should have these two choices: "According to the patron's settings" (default value), "Always per mail". Indeed, by default we want to use the settings in each patron records.
- [x] For some fields in the editor (Delay in days, Renewal duration, Checkout duration), can we use the same display "/day" as for field `reminders.fee_amount`? See image:
image
- [ ] The editor becomes complex and documentation is needed. The JSON schema descriptions are not displayed, or they don't exist? We should add a lot of them. Here are some descriptions to be added:
   - [ ] `checkout_duration`: Enter 0 if checkout is not allowed.
   - [ ] `number_renewals`: Enter 0 if renewal is not allowed.
   - [ ] `reminders`: They include the loan expiry notices, sent before the due date, and the reminders, sent after the due date.
   - [ ] `days_delay`: Number of days before the due date (for loan expiry notices) or after the due date (for reminders). This number defines when the notification is sent.
   - [ ] `overdue_fees.intervals`: During an interval, the specified amount will be incremented for each overdue open day.
   - [ ] `maximum_total_amount`: This defines the maximum fee amount per document, without considering the reminder fees.
   - [ ] `settings`: This defines which item categories and patron types this circulation policy applies to.
- [ ] When an overdue item is renewed, it should generate the "previewed fees", as if it would have been checked in (to be confirmed by @ManaDeweerdt )
- [x] I would not use the small caps for the vertical tabs. They are never used in RERO ILS.
- [ ] I was not able to modify the previewed fees according to a cipo. Example for [this item (type=Short)](https://ilsdev.test.rero.ch/professional/records/items/detail/1150) on loan for [John Smith (type=children)](https://ilsdev.test.rero.ch/professional/circulation/patron/2000000001). I modified the [corresponding cipo](https://ilsdev.test.rero.ch/professional/records/circ_policies/detail/4) but it has no effect on the previewed fees.
- [x] "Circulation policy name" and "Circulation policy description" to be renamed into "Name" and "Description"

lauren-d (approved):

zannkukai:
@pronguen I'm not able to reproduce the problem where the preview isn't updated when you change the cipo. Seems Ok on localhost. Should be retested on next deploy

iGormilhit (changes requested):
Commit message proposition:

```
transactions: optimise incremental creation

Creating an event for each incremental step is a poor idea (regarding to
performance issues). Indeed, for an old loan with a lot of incremental
fees, the system have to create a PatronTransactionEvent and reindex it
for each increment. This process takes a lot of time.

Creating only one event with a set of steps, solves the issue. So, only
one PatronTransactionEvent is created and only one reindex is necessary.
```

Issue #1495, opened 2020-11-24
circulation: block checkout/renew/request when patron expiration date is reached


benerken:
Since release `v0.14.0`, a patron has an expiration date. 
Today, the system doesn't check this date during circulation operation.
Once the patron expiration date is reached, the patron shouldn't be able to checkout/renew/request items.

**Improvement suggestion**

The system must block checkout/renew/request operations when patron expiration date is reached.

*The way of updating the expiration date is not addressed in this issue.*

iGormilhit:
See [US 1550][1] *patron: add expiration date*.

[1]: https://tree.taiga.io/project/rero21-reroils/us/1550

iGormilhit:
See also rero/rero-ils#1530.

PR #1743, opened 2021-03-03
circulation: block operations for expired patron

zannkukai:
* Checkout, request and renew circulation operations are blocked if the
  patron expiration date is reached.
* Introduces new utils fixtures and adapts unit tests.
* Closes rero/rero-ils#1495.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/patrons/api.py
+++ b/rero_ils/modules/patrons/api.py
@@ -380,11 +380,16 @@ class Patron(IlsRecord):
             # 'patron' argument are present into kwargs. This check can't
             # be relevant --> return True by default
             return True, []
-        # a blocked patron can't request any item
-        if patron.is_blocked:
-            return False, [patron.get_blocked_message()]
 
-        return True, []
+        messages = []
+        # 1) a blocked patron can't request any item
+        # 2) a patron with obsolete expiration_date can't request any item
+        if patron.is_blocked:
+            messages.append(patron.get_blocked_message())
+        if patron.is_expired:
+            messages.append(_('Patron rights expired.'))
+
+        return len(messages) == 0, messages
 
     @classmethod
     def can_checkout(cls, item, **kwargs):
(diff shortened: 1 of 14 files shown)


iGormilhit (approved):
Commit message approved.

benerken (approved):

rero_ils/modules/patrons/api.py
```diff
@@ -434,9 +439,22 @@ def remove_role(self, role_name):
...
+        """Shortcut to find the patron expiration date."""
+        date_string = self.patron.get('expiration_date')
+        if date_string:
+            return datetime.strptime(date_string, '%Y-%m-%d')
+
+    @property
+    def is_expired(self):
+        """Check if patron expiration date is obsolete."""
+        expiration_date = self.expiration_date
+        return expiration_date and datetime.now() > expiration_date
```
rerowep, line 456: Do we have to work with UTC times?

tests/fixtures/basics.py
```diff
@@ -0,0 +1,34 @@
...
+@pytest.fixture(scope="module")
+def yesterday():
+    """Get yesterday timestamp."""
+    return datetime.now(timezone.utc) - timedelta(days=1)
+
+
+@pytest.fixture(scope="module")
+def tomorrow():
+    """Get tomorrow timestamp."""
+    return datetime.now(timezone.utc) + timedelta(days=1)
```
rerowep, line 34: Shall we use `datetime.utcnow()` everywhere?

rero_ils/modules/patrons/api.py
zannkukai: `expiration_date` stored into the patron resource doesn't have any time (it's just YYYY-MM-DD).
Using `datetime.now()`, I'm sure tu use timestamp with server localized timezone.
So I think check is consistant and correct here

tests/fixtures/basics.py
zannkukai: This notation is a legacy from old unit test. See here : https://github.com/rero/rero-ils/blob/dev/tests/ui/holdings/test_serial_claims.py#L61

lauren-d (approved):

PR #1746, opened 2021-03-04
masking: enable masking for document, holdings and items

BadrAly:
When a librarian masks all items of a standard holdings,
the system masks the standard holdings automatically.
The standard holdings will be un masked when there is
at least one unmasked item attached to it.

* Migrates bib and item masking from Virtua.
* Creates dojson transformation for document masking.
* Hides masked records from public interface.
* Sets the _masked fields as optional for concerned records.
* Fixes problem of the item url validation.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


rero_ils/modules/items/api/api.py
```diff
@@ -247,15 +248,30 @@ def index(self, record):
...
+            update_holdings = False
+            if not holding.get('_masked') and not number_of_unmasked_items:
+                holding['_masked'] = True
+                update_holdings = True
+            elif holding.get('_masked') and number_of_unmasked_items:
+                holding['_masked'] = False
+                update_holdings = True
+            if update_holdings:
+                holding = holding.update(
+                    data=holding, dbcommit=True, reindex=True)
```
zannkukai, line 266: Just like me, you don't have many hairs. So maybe, you can simplify the code fr next read/update :
```python
# masked holding if all items are masked
if number_of_unmasked_items == 0 && not holding.get('_masked'):
    holding['_masked'] = True
    holding.update(data=holding, dbcommit=True, reindex=True)
# unmask holding if some related items are visible
elif number_of_unmasked_items > 0 && holding.get('_masked'):
    holding['_masked'] = False
    holding.update(data=holding, dbcommit=True, reindex=True)
```
1) you skip the third IF statement
2) no need of return value from update call
3) check if reindex is necessary at this step or if another index call is performed after this operation (because we update item or document)

rero_ils/modules/documents/views.py
```diff
@@ -593,8 +593,8 @@ def online_holdings(document_pid, viewcode='global'):
     if viewcode != current_app.config.get('RERO_ILS_SEARCH_GLOBAL_VIEW_CODE'):
         organisation = Organisation.get_record_by_viewcode(viewcode)
     query = HoldingsSearch()\
-        .filter('term', document__pid=document_pid) \
-        .filter('term', _masked=False)
+        .filter('term', document__pid=document_pid)
+    query = query.filter('bool', must_not=[Q('term', _masked=True)])
```
zannkukai, line 597:
```python
query = HoldingsSearch()\
    .filter('term', document__pid=document_pid) \
    .filter('bool', must_not=[Q('term', _masked=True)])
```

rero_ils/modules/items/api/record.py
```diff
@@ -510,3 +510,17 @@ def is_new_acquisition(self):
...
         return False
+
+    @classmethod
+    def get_number_masked_items_by_holdings_pid(cls, holding_pid):
+        """Returns the number of unmasked items and attached to a holding.
+
+        :param holding_pid: the pid of the holdings.
+        :return number of un masked items.
+        """
+        from . import ItemsSearch
```
zannkukai, line 521: ```from api import ItemsSearch``` ? Don't understand why `from .` is working :s

rero_ils/modules/items/api/api.py
BadrAly: `if not number_of_unmasked_items` = `number_of_unmasked_ite<ms == 0`
`if number_of_unmasked_items` = `number_of_unmasked_items == 0`

> check if reindex is necessary at this step or if another index call is performed after this operation (because we update item or document)

checking....

rero_ils/modules/items/api/record.py
BadrAly: because the `ItemsSearch` is in `..modules.items.api.__init__.py`

benerken (approved):

rero_ils/modules/items/api/record.py
```diff
@@ -510,3 +510,17 @@ def is_new_acquisition(self):
...
+        results = ItemsSearch() \
+            .filter('term', holding__pid=holding_pid) \
+            .filter('term', _masked=False) \
```
rerowep, line 524: Do we need `.filter('bool', must_not=[Q('term', _masked=True)])` here?

Garfield-fr (approved):

zannkukai (approved):

lauren-d (approved):

Issue #1584, opened 2021-01-04
Collection identifier should be optional


pronguen:
**How it works**

The field `collection_id` is required.

**How it should works**

It should be optional, as it is used for local purposes only.

Issue #1583, opened 2021-01-04
Collection title and description should be indexed in the document search


pronguen:
**How it works**

Currently, it is impossible, through the document search, to find a document by its collection.

**Improvement suggestion**

Collection title and description are indexed in the document search.

PR #1764, opened 2021-03-11
Document: index collection title and description.

zannkukai:
Currently, it is impossible, through the document search, to find a
document by its collection. With this PR, the collection title and
description are indexed into the document index.

Closes rero/rero-ils#1583.
Closes rero/rero-ils#1584.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/collections/api.py
+++ b/rero_ils/modules/collections/api.py
@@ -23,6 +23,7 @@ from rero_ils.modules.items.api import Item
 
 from .models import CollectionIdentifier
 from ..api import IlsRecord, IlsRecordsIndexer, IlsRecordsSearch
+from ..documents.api import Document, DocumentsIndexer
 from ..fetchers import id_fetcher
 from ..minters import id_minter
 from ..providers import Provider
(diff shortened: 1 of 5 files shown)


rerowep (approved):

rero_ils/modules/collections/api.py
```diff
@@ -88,6 +89,17 @@ class CollectionsIndexer(IlsRecordsIndexer):
 
     record_cls = Collection
 
+    def index(self, record):
```
BadrAly, line 92: The only problem here is that a simple update of the `collection` record triggers the indexing of all of its items `normally a lot` in the document record. There are many fields in the `collection` records that does not need an indexing at the document level (`subjects` for example).

zannkukai: @BadrAly I'm agree with you. Do I detect change only on title and description fields?
In this case, the only solution that I know is to create a listener on `before_record_update`. But this cause a call to this method for each update of any ressources (first operation of this listener should be to detect if the update is about a Collection).
What's you opinion about this solution ?

PR #1770, opened 2021-03-15
circulation category: adapt JSON schema

zannkukai:
* Adds 'disable_circulation' field into the JSON schema. When enable,
  all items related to this circulation category could not be relevant
  for any circulation operations.
* Adds 'displayed_status' field. Define the label to display in case of
  'disable_circulation' is set to True (multi_language field).
* Adds 'circulation_information' field. Define the label to display into
  the public interface (multi_language field).

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


rero_ils/modules/item_types/jsonschemas/item_types/item_type-v0.0.1.json
```diff
@@ -90,6 +94,97 @@
         }
       }
     },
+    "disable_circulation": {
```
pronguen, line 97: Not sure that the field name is the best one. The idea is not to block any circulation operations, but to always *display* an item with a red availability dot (instead of the green dot).

Name suggestion: `negative_availability`

Typical example: the item has as base item type "Standard" and as temporary item type "In acquisition". For "in acquisition", all items are displayed with the red dot. But they can be requested for example.

rero_ils/modules/item_types/jsonschemas/item_types/item_type-v0.0.1.json
```diff
@@ -90,6 +94,97 @@
...
+      "type": "boolean",
+      "title": "Mark the items of this category as not available",
+      "description": "If checked, none circulation operation can be performed on items related to this category.",
```
pronguen, line 100:
```suggestion
      "description": "If enabled, all items of this circulation category will be displayed as not available, with the red availability symbol.",
```

Note: I prefer "If enabled..." because it is already in use in the location editor.

rero_ils/modules/item_types/jsonschemas/item_types/item_type-v0.0.1.json
zannkukai: done

rero_ils/modules/item_types/jsonschemas/item_types/item_type-v0.0.1.json
zannkukai: 👍 done

data/item_types.json
```diff
@@ -24,6 +26,17 @@
...
+    "negative_availability": false,
+    "circulation_information": [
+      {
+        "language": "eng",
+        "label": "On-site consultation"
+      },
+      {
+        "language": "fre",
+        "label": "Consultation sur place"
+      }
```
rerowep, line 38: In the interface we also have Italian and German. Do we have to add Italian and German here?

data/item_types.json
```diff
@@ -33,6 +46,17 @@
...
+    "negative_availability": true,
+    "displayed_status": [
+      {
+        "language": "eng",
+        "label": "Not circulating"
+      },
+      {
+        "language": "fre",
+        "label": "Pas de pr\u00eat possible"
+      }
```
rerowep, line 58: See above

rerowep (approved):

data/item_types.json
```diff
@@ -6,6 +6,7 @@
     "organisation": {
       "$ref": "https://ils.rero.ch/api/organisations/1"
     },
+    "negative_availability": false,
```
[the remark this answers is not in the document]
zannkukai: It's defined by the PO. In practice, we can create a "Missing" circulation category (itty). Into the editor, we would ike to enable the toggle to specify than all related items should be displayed as "not available".

data/item_types.json
zannkukai: We could, add `de` and `it` translation into data but it's not required.
The UUI will try to use the correct language label, if none label are defined for the UI language, then the itty_name will be used as default value.

Issue #1872, opened 2021-05-04
Harvested e-books should be marked as available


pronguen:
**How it works**

Imported ebooks are always marked as not available.

![image](https://user-images.githubusercontent.com/8154915/117011118-ce1a7880-aced-11eb-88bd-8e80a47fd9d3.png)

**How it should works**

* They are always marked as available.

PR #1912, opened 2021-05-11
holdings: fix electronic holding availability

zannkukai:
* Mark an electronic holding as always available.
* Clean some tests from fixture `es_clear`.
* Closes rero/rero-ils#1872.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/documents/listener.py
+++ b/rero_ils/modules/documents/listener.py
@@ -41,12 +41,14 @@ def enrich_document_data(sender, json=None, record=None, index=None,
         # HOLDINGS
         holdings = []
         document_pid = record['pid']
-        es_holdings = HoldingsSearch().filter(
-            'term', document__pid=document_pid
-        ).scan()
+        es_holdings = HoldingsSearch()\
+            .filter('term', document__pid=document_pid)\
+            .scan()
         for holding in es_holdings:
-            data = {
-                'pid': holding.pid,
+            holding = holding.to_dict()
+            holding_obj = Holding.get_record_by_pid(holding['pid'])
+            hold_data = {
+                'pid': holding['pid'],
                 'location': {
                     'pid': holding['location']['pid'],
                 },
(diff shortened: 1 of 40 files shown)


iGormilhit (commented):
In the commit message: `s/holding/holdings/` (holdings with an s).

iGormilhit (approved):

lauren-d (approved):

rero_ils/modules/documents/listener.py
```diff
@@ -102,7 +102,7 @@ def enrich_document_data(sender, json=None, record=None, index=None,
                 #   'nested' structure.
                 acq_date = item.get('acquisition_date')
                 if acq_date:
```
rerowep, line 104: Could be written in one line: `if item.get('acquisition_date'):`

rero_ils/modules/documents/listener.py
```diff
@@ -116,22 +116,20 @@ def enrich_document_data(sender, json=None, record=None, index=None,
...
                     }
                     coll_data = {k: v for k, v in coll_data.items() if v}
-                    item_record.setdefault('collections', []).append(coll_data)
-
-                data.setdefault('items', []).append(item_record)
-            data['available'] = Holding.isAvailable(es_items)
-            holdings.append(data)
+                    item_data.setdefault('collections', []).append(coll_data)
+                hold_data.setdefault('items', []).append(item_data)
+            hold_data['available'] = holding_obj.available
```
rerowep, line 131: We have already all the items, it should be possible to calculate the availability without the Holding object directly in the items loop. Maybe we can get some speed improvements here.

rero_ils/modules/documents/listener.py
zannkukai: It's because we use `acq_date` two times : One for test, one for assignment (L109).
Using this method, we skip one dictionary search

rero_ils/modules/documents/listener.py
zannkukai: You're right. But `Holding.available` method has its own logic (really simple but centralized in one place).
I tried to pass items array to the `available` method, but we can ensure every time that this array contains all items for this holding ; in this case, this method will return a non-consistent value.

rerowep (changes requested):
How long do's this code needs to process a Holding with 7000 items attached. I think we need to minimize DB and ES request to a minimum in this functions.

rero_ils/modules/documents/serializers.py
```diff
@@ -74,15 +74,14 @@ def preprocess_record(self, pid, record, links_factory=None, **kwargs):
...
+            'RERO_ILS_SEARCH_GLOBAL_VIEW_CODE')
+        viewcode = request.args.get('view', global_view_code)
         records = results.get('hits', {}).get('hits', {})
         for record in records:
             metadata = record.get('metadata', {})
-            available = Document.get_record_by_pid(
-                metadata.get('pid')).is_available(viewcode)
-            metadata['available'] = available
+            document = Document.get_record_by_pid(metadata.get('pid'))
+            metadata['available'] = document.is_available(viewcode)
```
rerowep, line 84: This two lines are responsible for a lot of delay during display of the results. We have to find a faster or asynchron way to do it.

rerowep (changes requested):

rero_ils/modules/documents/serializers.py
zannkukai: You're right, potentially we read all holdings and all items to check availability.
But, if we found one related available holding/item, the loops breaks.
Don't' know how to improve this code, because it's needed by UI at this time.

I think it should be managed into a separate PR/issue (because it will have a huge impact on UI)

rerowep (approved):

iGormilhit (approved):
Commit message approved.

jma (approved):

PR #1973, opened 2021-06-02
resources: improve organisation ref link validation

BadrAly:
During the loading of ilspilot it was revealed that additional validation
is needed for the organisation field in all resources. System accepted
bad organisation data mainly for the library resource that prevented users
from logging in. With this commit, organisation field is validated correctly.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/acq_accounts/jsonschemas/acq_accounts/acq_account-v0.0.1.json
+++ b/rero_ils/modules/acq_accounts/jsonschemas/acq_accounts/acq_account-v0.0.1.json
@@ -75,6 +75,10 @@
     "organisation": {
       "title": "Organisation",
       "type": "object",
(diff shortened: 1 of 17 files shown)


iGormilhit (approved):

benerken (approved):

Garfield-fr (approved):

zannkukai (commented):
> Are you tested the editor after this change ?
> I remember than it may cause problem if a required field is defined into a not-visible field.

I tested the editing of all the resources and seems working fine. I imagine, testing in depth will be carried out by the PO when testing the release 1.3

rerowep (approved):

zannkukai (approved):

PR #1980, opened 2021-06-04
patron type: add `code` field

sebastiendeleze:
Adds a `code` field to the patron type. This field is mainly useful to store the code from Virtua and keep the OAuth server working.

Co-Authored-by: Sébastien Délèze <sebastien.deleze@rero.ch>

### DIFF
--- a/rero_ils/modules/patron_types/jsonschemas/patron_types/patron_type-v0.0.1.json
+++ b/rero_ils/modules/patron_types/jsonschemas/patron_types/patron_type-v0.0.1.json
@@ -8,7 +8,8 @@
     "name",
     "description",
     "subscription_amount",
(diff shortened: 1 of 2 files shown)


rerowep (approved):
Do we have to add `code` to the fixtures and do we have to update the tests?

Garfield-fr (approved):

sebastiendeleze:
> Do we have to add `code` to the fixtures and do we have to update the tests?

I don't think it's necessary, the field is not mandatory and only used in the context of OAuth.

benerken (approved):

iGormilhit (approved):

BadrAly (approved):

PR #2046, opened 2021-06-17
core: fix the wrong way to update a record

jma:
core: fix the worng way to update a record 

The way of `invenio-records` has been misunderstood. The `update` method
change only the `dict` value without any db transaction. To push the
transaction on the db, the `commit` method should be called, which is
not the case for the create. The previous implementation called the
validation twice and create two operation logs (create and update) when
a resource was created.

* Avoids multiple validations when a record is created.
* Adds a new `commit` parameter to the update, replace methods in
  `ILSRecord`.
* Fixes two operation logs creation when a resource a created, one for
  create one for update.

## How to test?

- As it is a core modification it is important to test a bit the resource creation, modification and update and also some circulation operations.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/items/api/circulation.py
+++ b/rero_ils/modules/items/api/circulation.py
@@ -153,6 +153,8 @@ class ItemCirculation(ItemRecord):
                     'item_pid': item_pid_to_object(item.pid),
                     'patron_pid': patron_pid
                 }
+                data.setdefault(
+                    'transaction_date', datetime.utcnow().isoformat())
                 loan = Loan.create(data, dbcommit=True, reindex=True)
             if not patron_pid and loan:
                 kwargs.setdefault('patron_pid', loan.patron_pid)
@@ -1122,6 +1124,7 @@ class ItemCirculation(ItemRecord):
         else:
             if item['status'] != ItemStatus.MISSING and on_shelf:
                 item['status'] = ItemStatus.ON_SHELF
+        item.commit()
         if dbcommit:
             item.dbcommit(reindex=True, forceindex=True)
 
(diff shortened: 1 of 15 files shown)


rero_ils/modules/items/api/circulation.py
```diff
@@ -153,6 +153,8 @@ def complete_action_missing_params(
                     'item_pid': item_pid_to_object(item.pid),
                     'patron_pid': patron_pid
                 }
+                data.setdefault(
+                    'transaction_date', datetime.now(timezone.utc).isoformat())
```
rerowep, line 157: I think there is a ` utcnow` for this.

rero_ils/modules/loans/api.py
```diff
@@ -246,24 +246,25 @@ def create(cls, data, id_=None, delete_pid=True,
             reindex=reindex, **kwargs)
         return record
 
-    def update(self, data, dbcommit=False, reindex=False):
+    def update(self, data, commit=False, dbcommit=False, reindex=False):
```
rerowep, line 249: Normally default for `commit` is True, why it is different here?

sebastiendeleze (approved):
There's just a type in the commit message:

core: fix the **worng** way to update a record

rero_ils/modules/loans/api.py
jma: Because of other invenio modules i.e. invenio-circulation. The default behaviour in invenio is that after an update the commit is not effective. But if `dbcommit` is True then `commit` is done.

lauren-d (approved):

iGormilhit (approved):
Commit message approved.

PR #2157, opened 2021-07-13
circulation: improve request to validate list

Garfield-fr:
* Improves performance when displaying a lot of requests.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


iGormilhit (changes requested):
Commit message title proposition: `circulation: improve request to validate list`

Could you elaborate on why this PR is useful? Thanks.

iGormilhit (approved):

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+
+        # Proxy record
+        patrons = {}
+        locations = {}
+        libraries = {}
+        items = {}
+        item_types = {}
+
+        def loans_pending():
+            """Get loans pending."""
```
iGormilhit, line 307:
```suggestion
            """Get pending loans."""
```

iGormilhit (commented):
Docstrings improvements.

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+                .filter('term', library_pid=library_pid)\
+                .sort({'_created': {"order": 'asc'}})\
+                .source(includes=[
+                    'pid', 'transaction_date', 'item_pid', 'patron_pid',
+                    'document_pid', 'library_pid', 'state',
+                    'transaction_location_pid', 'pickup_location_pid'
+                ])\
+                .scan()
+
+        def patron_by_pid(patron_pid):
```
zannkukai, line 320: You need to pay attention to memory usage. Because you keep all patrons, items, ... you could load a lot of data and cause memory overflow.

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+            """Get patron by pid."""
+            if patron_pid not in patrons:
+                patrons[patron_pid] = PatronsSearch()\
+                    .filter('term', pid=patron_pid)\
+                    .source(includes=[
+                        'pid', 'first_name', 'last_name', 'patron.barcode'])\
+                    .execute()[0].to_dict()
```
zannkukai, line 327: It should not happen in the best world .... but this will cause Exception if patron_pid doesn't exists. By example if we delete patron/item/location and not the corresponding request.

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
             if field in self:
                 self[field] = self[field].isoformat()
 
+    @classmethod
+    def requested_loans_to_validate(cls, library_pid):
+        """Get Requests to be validated."""
```
sebastiendeleze, line 295: doc for parameters is missing

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+        from ..item_types.api import ItemTypesSearch
+        from ..items.api import ItemsSearch
```
sebastiendeleze, line 297: Cannot be imported on top?

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+            return item_types[item_type_pid]
+
+        metadata = []
+        item_pids = []
+
+        loans = loans_pending()
+        for loan in loans:
+            item_pid = loan['item_pid']['value']
+            item = item_by_pid(item_pid)
+            if (len(item) > 0):
```
zannkukai, line 376: `item_by_pid` seems return an Item not a list. This test seems invalid. Try to use `if item:`
Not necessary to enclose condition between parentheses in python (best practice).

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+                    item_data['library']['name'] = library_name_by_pid(
+                        item_data['library']['pid'])
+                    item_data['location']['name'] = location_by_pid(
+                        item_data['location']['pid'])['name']
+                    patron_data = patron_by_pid(loan_data['patron_pid'])
+                    patron_first_name = patron_data['first_name']
+                    patron_last_name = patron_data['last_name']
+                    loan_data['patron'] = {
+                        'barcode': patron_data['patron']['barcode'],
+                        'name': f'{patron_first_name} {patron_last_name}'
```
sebastiendeleze, line 396: You can directly do:
```suggestion
                        'name': f'{patron_data["first_name"]} {patron_data["last_name"]}'
```
Instead of storing in a variable

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+        for loan in loans:
+            item_pid = loan['item_pid']['value']
+            item = item_by_pid(item_pid)
+            if (len(item) > 0):
+                add = True
+                item = item[0]
+                if 'temporary_item_type' in item:
+                    item_type_pid = item['temporary_item_type']['pid']
+                    itemType = item_type_by_pid(item_type_pid)
+                    add = True if len(itemType) == 0 else False
```
zannkukai, line 382: same here. `add` is already True. Just use 
```
if item_type is None:
   add = false
```

rero_ils/modules/loans/api.py
```diff
@@ -290,6 +290,122 @@ def date_fields2str(self):
...
+        loans = loans_pending()
+        for loan in loans:
+            item_pid = loan['item_pid']['value']
+            item = item_by_pid(item_pid)
+            if (len(item) > 0):
+                add = True
+                item = item[0]
+                if 'temporary_item_type' in item:
+                    item_type_pid = item['temporary_item_type']['pid']
+                    itemType = item_type_by_pid(item_type_pid)
```
zannkukai, line 381: Use snake_case --> `item_type`

sebastiendeleze (approved):

zannkukai (changes requested):

BadrAly (approved):

rero_ils/modules/loans/api.py
Garfield-fr: Because of the circular references

jma (commented):
I think it is not usefull to keep the items in cache as we do not have more than one request to validate by item.

rero_ils/modules/loans/api.py
jma: Keeping items is useless as we can not have more than one request to validate by item.

zannkukai:
> I think it is not usefull to keep the items in cache as we do not have more than one request to validate by item.

@jma you shouldn't .... but it can occurs.
It happens that just after a lesson a group of student (ex: 20) asked for the same document but we have only 2 items. So in few times, we can have multiple request on same item.

Issue #2282, opened 2021-08-10
The link to the patron account should redirect to a login page if the user is not authentified


pronguen:
**Describe the bug**

In the notifications, the link to the patron account `https://[instance]/[view]/patrons/profile` open a page *Unauthorized. You need to be authenticated to view this page*.

**To Reproduce**

Open the link https://ils.test.rero.ch/global/patrons/profile when you are not connected.

**Expected behavior**

The link redirects to the login page.

**Context**

* server: bib.rero.ch
* version: `v1.4.6`

PR #2302, opened 2021-08-17
notification: fix patron profile URL

zannkukai:
Closes rero/rero-ils#2282.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.


Garfield-fr (approved):

lauren-d (approved):

rero_ils/modules/notifications/api.py
```diff
@@ -187,12 +188,12 @@ def replace_pids_and_refs(self):
...
             base_url = current_app.config.get('RERO_ILS_APP_URL')
-            profile_url = f'{base_url}/{view_code}/patrons/profile'
+            next_page_url = urllib.parse.quote(f'{view_code}/patrons/profile')
+            profile_url = f'{base_url}/signin/?next={next_page_url}'
             data['loan']['profile_url'] = profile_url
-
             return data
         except Exception as error:
-            raise(error)
+            raise error
```
rerowep, line 196: I think you have to write something like `raise Exception(error)`.

rero_ils/modules/notifications/api.py
```diff
@@ -187,12 +188,12 @@ def replace_pids_and_refs(self):
             patron = Patron.get_record_by_pid(data['loan']['patron']['pid'])
             view_code = patron.get_organisation().get('code')
             base_url = current_app.config.get('RERO_ILS_APP_URL')
-            profile_url = f'{base_url}/{view_code}/patrons/profile'
+            next_page_url = urllib.parse.quote(f'{view_code}/patrons/profile')
+            profile_url = f'{base_url}/signin/?next={next_page_url}'
             data['loan']['profile_url'] = profile_url
```
rerowep, line 193: I don't think this is the right profile url. The url is correct as it was before, but we have to get redirected to the signin if we are not logged in.

rerowep (changes requested):

rero_ils/modules/notifications/api.py
rerowep: I think we have to change this:
```
def login_and_patron():
    """Patron is logged in."""
    if current_user and not current_user.is_authenticated:
        abort(401)
    if len(current_patrons) == 0:
        abort(403)
```

rero_ils/modules/notifications/api.py
zannkukai: `error` is already an `Exception`subclass. Using parentheses to raise an exception isn't a best practice.
Tested on localhost and works as expected.

rerowep (approved):

PR #2342, opened 2021-08-28
notifications: do not send email in case of error

jma:
* Avoids to send an incomplete email in case of error during the
  notification generation.
* Fixes resend email every 15 minutes.
* Fixes bad logger messages.
* Adds traceback to the notification error logs.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/notifications/dispatcher.py
+++ b/rero_ils/modules/notifications/dispatcher.py
@@ -47,7 +47,6 @@ class Dispatcher:
         :param verbose: Verbose output.
         :returns: dictionary with processed and send count
         """
-
         def get_dispatcher_function(channel):
             try:
                 communication_switcher = current_app.config.get(
@@ -55,8 +54,8 @@ class Dispatcher:
                 return communication_switcher[channel]
             except KeyError:
                 current_app.logger.warning(
-                    'The communication channel of the patron'
-                    f' (pid: {patron["pid"]}) is not yet implemented')
+                    f'The communication channel: {channel}'
+                    ' is not yet implemented')
                 return Dispatcher.not_yet_implemented
 
         sent = not_sent = 0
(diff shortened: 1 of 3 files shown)


rero_ils/modules/notifications/dispatcher.py
```diff
@@ -47,39 +49,40 @@ def dispatch_notifications(cls, notification_pids=None, resend=False,
...
             try:
                 cls._process_notification(
                     notification, resend, aggregated)
             except Exception as error:
+                errors += 1
                 current_app.logger.error(
                     f'Notification has not be sent (pid: {notification.pid},'
                     f' type: {notification["notification_type"]}): '
                     f'{error}')
+                traceback.print_exc()
```
rerowep, line 85: I don't know where we can see the tracback print. Probably stderr, do we have to use the logger for this?

rerowep (approved):

rero_ils/modules/notifications/dispatcher.py
lauren-d: maybe we ca use `exc_info=True` and/or `stack_info=True`  arguments in logger.error like:
```
current_app.logger.error(
                    f'Notification has not be sent (pid: {notification.pid},'
                    f' type: {notification["notification_type"]}): '
                    f'{error}', exc_info=True, stack_info=True)
```
see https://docs.python.org/3/library/logging.html#logging.Logger.debug

jma: Thanks, I learnt something.

lauren-d (approved):

zannkukai (approved):

PR #2357, opened 2021-09-02
notifications: add due soon date

jma:
* Adds a new `due_soon_date` property.
* Computes the due soon loan from elasticsearch.
* Creates a loan extension to put the new due soon property at each
  update.

__Note__: need to update the es mapping and do a mass correction for the loan ON_LOAN to update the due_soon_date.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/loans/api.py
+++ b/rero_ils/modules/loans/api.py
@@ -37,6 +37,7 @@ from invenio_jsonschemas import current_jsonschemas
 from rero_ils.modules.circ_policies.api import DUE_SOON_REMINDER_TYPE, \
     OVERDUE_REMINDER_TYPE, CircPolicy
 
+from .extensions import SetDueSoonDate
 from ..api import IlsRecord, IlsRecordError, IlsRecordsIndexer, \
     IlsRecordsSearch
 from ..errors import NoCirculationActionIsPermitted
@@ -118,6 +119,8 @@ class Loan(IlsRecord):
         "start_date",
         "transaction_date"
     ]
+    # Invenio Records extensions
+    _extensions = [SetDueSoonDate()]
 
     def __init__(self, data, model=None):
         """Loan init."""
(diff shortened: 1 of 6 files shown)


rero_ils/modules/loans/api.py
```diff
@@ -470,20 +473,17 @@ def is_loan_overdue(self):
...
-        due_date = ciso8601.parse_datetime(self.end_date).replace(
-            tzinfo=timezone.utc)
-        days_before = circ_policy.due_soon_interval_days
-        if days_before:
-            start_date = ciso8601.parse_datetime(self.get('start_date'))
-            due_soon_date = due_date - timedelta(days=days_before)
-            return start_date < due_soon_date <= date < due_date
+        due_soon_date = self.get('due_soon_date')
+        """Check if a loan is due soon."""
+        if due_soon_date:
```
zannkukai, line 485: Could you specify that this check if for legacy loan without 'due_soon_date' attribute ? 
Maybe a `TODO` comment to remember to remove this statement when legacy loan will be obsolete

rero_ils/modules/loans/api.py
```diff
@@ -783,7 +783,11 @@ def create_notification(self, trigger=None, _type=None, counter=0):
             # Internal notification (library destination) should be directly
             # dispatched. Other notifications types could be asynchronously
             # processed (to save server response time).
-            dispatch = n_type in NotificationType.INTERNAL_NOTIFICATIONS
+            dispatch = n_type in NotificationType.INTERNAL_NOTIFICATIONS + \
+                [
+                    NotificationType.OVERDUE,
+                    NotificationType.DUE_SOON
+                ]
```
zannkukai, line 790: You should use `NotificationType.REMINDERS_NOTIFICATIONS`

rero_ils/modules/loans/extensions.py
```diff
@@ -0,0 +1,46 @@
...
+from datetime import timedelta, timezone
+
+import ciso8601
+from invenio_records.extensions import RecordExtension
+
+
+class SetDueSoonDate(RecordExtension):
+    """Set the due soon date when the loan is updated."""
+
+    def pre_commit(self, record):
```
zannkukai, line 30: From my memory : the `pre_commit` hook is only used when you **update** a record. 
If the `due_soon_date` should be created anytime, you also need to use the **pre_create** hook.
Something like 
```python

def _create_due_soon(self, record):
    ...
    
pre_commit = _create_due_soon
pre_create = _create_due_soon
```

zannkukai (changes requested):

rero_ils/modules/loans/extensions.py
jma: I do not think so. A loan can becomes ON_LOAN only at update, not during the creation.

zannkukai: You're right. I hope it will be always the case :-)

iGormilhit (approved):
Commit message approved.

zannkukai (approved):

benerken (approved):

lauren-d (approved):

PR #2372, opened 2021-09-08
acquisition: order notification preview

zannkukai:
* Adds dumpers class on `Vendor` resource.
* Adds API endpoint to preview an acquisition order.
* Creates vendor order notification templates.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/eng.tpl.txt
+++ b/rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/eng.tpl.txt
@@ -1,0 +1,39 @@
+{{ order.library.name }} - Order [{{ order.reference }}]
+
+Our reference: {{ order.reference }}
+Date : {{ order.order_date }}
+
+Dear Sir,
+
+We would like to order the following list of books:
+
+------------------------------------------------------------
+{%- for line in order.order_lines %}
+Title : {{ line.document.title }}
+{%- if line.document.identifiers %}
+ISBN : {{ line.document.identifiers | join(', ') }}
+{%- endif %}
+{%- if line.note %}
+Note : {{ line.note }}
+{%- endif %}
+Quantity : {{ line.quantity }}
+------------------------------------------------------------
+{%- endfor -%}
+{%- if order.note %}
+
+We would like to draw your attention to the following remark:
+
+{{ order.note }}
+{%- endif %}
+
+Best regards,
+
+{{ order.library.shipping_informations.name }}
+
+Shipping address:
+{{ order.library.shipping_informations | address_block(language=order.vendor.language) }}
+
+{% if order.billing_informations -%}
+Billing address:
+{{ order.library.billing_informations | address_block(language=order.vendor.language) }}
+{%- endif %}
--- a/rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/fre.tpl.txt
+++ b/rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/fre.tpl.txt
@@ -1,0 +1,39 @@
+{{ order.library.name }} - Commande [{{ order.reference }}]
+
+Notre référence: {{ order.reference }}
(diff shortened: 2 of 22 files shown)


rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/eng.tpl.txt
```diff
@@ -0,0 +1,39 @@
...
+We would like to order the following list of books:
+
+------------------------------------------------------------
+{%- for line in order.order_lines %}
+Title : {{ line.document.title }}
+{%- if line.document.identifiers %}
+ISBN : {{ line.document.identifiers | join(', ') }}
+{%- endif %}
+{%- if line.note %}
+Note : {{ line.note }}
```
BadrAly, line 17: filter by `vendor_note`, we do not want to send the `staff_note` to the vendor

rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/eng.tpl.txt
```diff
@@ -0,0 +1,39 @@
...
+{%- endif %}
+Quantity : {{ line.quantity }}
+------------------------------------------------------------
+{%- endfor -%}
+{%- if order.note %}
```
BadrAly, line 22: same here, filter by `vendor_note`

rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/fre.tpl.txt
```diff
@@ -0,0 +1,39 @@
...
+
+Nous souhaitons vous commander la liste d'ouvrages suivants:
+
+------------------------------------------------------------
+{%- for line in order.order_lines %}
+Titre : {{ line.document.title }}
+{%- if line.document.identifiers %}
+ISBN : {{ line.document.identifiers | join(', ') }}
+{%- endif %}
+{%- if line.note %}
```
BadrAly, line 16: the remark for the ` vendor_note` applies to all templates

rero_ils/modules/vendors/api.py
```diff
@@ -72,11 +72,13 @@ class Vendor(IlsRecord):
...
+        :return the best possible email to use for this vendor. If the specific
+                order contact information does not exist, the default contact
+                information will be used.
         """
-        return self.get(
-            'order_contact', self.get('default_contact', {})
-        ).get('email')
+        return self\
+            .get('order_contact', self.get('default_contact', {}))\
+            .get('email')
```
BadrAly, line 81: in this case, we need to validate that each vendor has at least one email. I did not find this check in the vendor class

BadrAly (changes requested):

rero_ils/modules/acq_orders/templates/rero_ils/vendor_order_mail/eng.tpl.txt
zannkukai: This is the beauty of custom dumper :-)
into the context i build to render the template, I take only vendor note. So `line.note` and `order.note` are filtered vendor note (check the acq_order dumper and acq_order_line dumper classes)

rero_ils/modules/vendors/api.py
zannkukai: Not really :-) 
First this will be check about an `order_contact` or an `default_contact` dict attribute. If none are found, we will use an empty dict as default value.
On this dict, we get the `email` attribute. If this attribute isn't present, then `None` is returned.
So, for me, this method is correct.

But, in the UI, I need to check if vendor has an email (order or default). If not, then we should not select 'send to vendor' option.

BadrAly: clear now, thanks.
Does it mean the `send_order` will fail if we use the `console` instead of the UI for vendors without email?

BadrAly (approved):

PR #2386, opened 2021-09-15
templates: prevent resolving $ref from data field

zannkukai:
the `$ref` from the template data doesn't need to be resolved when the
record is indexed into ES. This should be cause indexing errors if the
related resource doesn't exists anymore and we will not consider this
spacial case.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/templates/api.py
+++ b/rero_ils/modules/templates/api.py
@@ -17,7 +17,6 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 """API for manipulating Templates."""
-
 from functools import partial
 
 from .extensions import RemoveDataPidExtension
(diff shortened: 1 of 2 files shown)


iGormilhit (changes requested):
Commit message proposition:

```
templates: prevent resolving $ref from data field

The `$ref` of the template data doesn't need to be resolved when the
record is indexed by ES, as it causes indexing errors when the related
resource doesn't exist anymore.
```

iGormilhit (approved):

lauren-d (approved):

lauren-d:
be careful, tests failed

rerowep (approved):

PR #2392, opened 2021-09-17
footer: fix missing version

Garfield-fr:

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/ext.py
+++ b/rero_ils/modules/ext.py
@@ -61,6 +61,7 @@ from .users.views import UsersCreateResource, UsersResource
 from .utils import set_user_name
 from ..filter import empty_data, format_date_filter, get_record_by_ref, \
     jsondumps, node_assets, text_to_id, to_pretty_json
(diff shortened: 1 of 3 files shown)


zannkukai (commented):
I don't understand why the `version` wasn't print into footer.
`version` was a context argument for frontpage.html --> extending page.html  --> including footer.html.

Garfield-fr:
> I don't understand why the `version` wasn't print into footer.
> `version` was a context argument for frontpage.html --> extending page.html --> including footer.html.

In the current version there is an error. The version number is displayed only on the home page, because the parameter must be passed to each template. With the "Global" version, it is no longer necessary to pass this parameter, because it is accessible through a global variable of jinja.

zannkukai (approved):

lauren-d (approved):

iGormilhit (approved):

PR #2403, opened 2021-09-23
acquisition: delete order and its order lines

BadrAly:
* Makes cascade deleting of order lines possible when parent order
   status is still PENDING.

https://tree.taiga.io/project/rero21-reroils/task/2281

## Dependencies

My PR depends on the US `US_acquisition`

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/acq_orders/api.py
+++ b/rero_ils/modules/acq_orders/api.py
@@ -23,7 +23,7 @@ from functools import partial
 from flask_babelex import gettext as _
 
 from .extensions import AcquisitionOrderCompleteDataExtension, \
-    AcquisitionOrderDynamicFieldsExtension
+    AcquisitionOrderExtension
 from .models import AcqOrderIdentifier, AcqOrderMetadata, AcqOrderStatus
 from ..acq_order_lines.api import AcqOrderLine, AcqOrderLinesSearch
 from ..acq_order_lines.models import AcqOrderLineStatus
@@ -63,7 +63,7 @@ class AcqOrder(IlsRecord):
     """AcqOrder class."""
 
     _extensions = [
-        AcquisitionOrderDynamicFieldsExtension(),
+        AcquisitionOrderExtension(),
         AcquisitionOrderCompleteDataExtension()
     ]
 
@@ -184,6 +184,15 @@ class AcqOrder(IlsRecord):
         """Get reasons not to delete record."""
         cannot_delete = {}
         links = self.get_links_to_me()
+        # The link with AcqOrderLine ressources isn't a reason to not delete
+        # an AcqOrder. Indeed, when we delete an AcqOrder, we also delete all
+        # related AcqOrderLines (cascade delete). Check the extension
+        # ``pre_delete`` hook.
+        links.pop('acq_order_lines', None)
+        if self.status != AcqOrderStatus.PENDING:
+            cannot_delete['others'] = {
+                _(f'Order status is {self.status}'): True
+            }
         if links:
             cannot_delete['links'] = links
         return cannot_delete
@@ -200,3 +209,15 @@ class AcqOrdersIndexer(IlsRecordsIndexer):
         :param record_id_iterator: Iterator yielding record UUIDs.
         """
         super().bulk_index(record_id_iterator, doc_type='acor')
+
+    def delete(self, record):
+        """Delete a record from indexer.
+
+        First delete order lines from the ES index, then delete the order.
+        """
+        es_query = AcqOrderLinesSearch()\
+            .filter('term', acq_order__pid=record.pid)
+        if es_query.count():
+            es_query.delete()
+            AcqOrderLinesSearch.flush_and_refresh()
+        super().delete(record)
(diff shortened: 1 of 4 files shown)


rero_ils/modules/acq_orders/api.py
```diff
@@ -184,6 +184,9 @@ def reasons_not_to_delete(self):
         """Get reasons not to delete record."""
         cannot_delete = {}
         links = self.get_links_to_me()
+        # allow deletion of order and its line when its status is still PENDONG
```
iGormilhit, line 187:
```suggestion
        # allow deletion of order and its line when its status is still PENDING
```

rero_ils/modules/acq_orders/api.py
```diff
@@ -200,3 +203,13 @@ def bulk_index(self, record_id_iterator):
         :param record_id_iterator: Iterator yielding record UUIDs.
         """
         super().bulk_index(record_id_iterator, doc_type='acor')
+
+    def delete(self, record):
+        """Delete a record from indexer."""
+        # delete order lines from the ES index first before deleting the order.
```
iGormilhit, line 209:
```suggestion
        # First delete order lines from the ES index, then delete the order.
```
Furthermore, could/should this comment be part of the docstring? In fact, I have this question on a much broader level: what is docstring, what is comments?

iGormilhit (changes requested):
Commit message proposition:

```
acquisitions: delete order and its order lines

* Makes cascade deleting of order lines possible when parent order
  status is still PENDING.
```

rero_ils/modules/acq_orders/extensions.py
```diff
@@ -43,6 +43,20 @@ def pre_load(self, data, loader=None):
...
+        """Called before a record is deleted.
+
+        :param record: the record metadata.
+        """
+        # For pending orders, we are allowed to delete all of its
+        # line orders without futher checks.
+        # there is no need to check if it is a pending order or not because
+        # the can_delete is already execute in the method self.delete
+        order_record = record.__class__.get_record_by_pid(record.pid)
+        if order_record:
```
zannkukai, line 56: the `record` method attribute is already the AcqOrder record. No need to reload it (2 DB calls).
Just use 
```
for line in record.get_order_lines():
   line.delete(delindex=True)
```

rero_ils/modules/acq_orders/extensions.py
```diff
@@ -43,6 +43,20 @@ def pre_load(self, data, loader=None):
         data.pop('total_amount', None)
         data.pop('status', None)
 
+    def pre_delete(self, record, force=False):
```
zannkukai, line 46: It should be interesting to either change the extension name, either create an extension with your `pre_delete` hook. Solution#1 seems great (AcquisitionOrderExtension ?)

rero_ils/modules/acq_orders/api.py
```diff
@@ -200,3 +203,15 @@ def bulk_index(self, record_id_iterator):
...
+    def delete(self, record):
+        """Delete a record from indexer.
+
+        First delete order lines from the ES index, then delete the order.
+        """
+        es_query = AcqOrderLinesSearch()\
+            .filter('term', acq_order__pid=record.pid)
+        if es_query.count():
+            es_query.delete()
+            AcqOrderLinesSearch.flush_and_refresh()
```
zannkukai, line 216: Could you put everything into the `pre_delete` function. Seems confused to have multiple codes on different places.

rero_ils/modules/acq_orders/api.py
```diff
@@ -184,6 +184,9 @@ def reasons_not_to_delete(self):
         """Get reasons not to delete record."""
         cannot_delete = {}
         links = self.get_links_to_me()
+        # allow deletion of order and its line when its status is still PENDING
```
zannkukai, line 187: I will more verbose to well understand why we do that.
```
The link with AcqOrderLine ressources isn't a reason to not delete an AcqOrder.
Indeed, when we delete an AcqOrder, we also delete all related AcqOrderLines (cascade delete).
Check the extension ``pre_delete`` hook.
```

rero_ils/modules/acq_orders/api.py
```diff
@@ -184,6 +184,9 @@ def reasons_not_to_delete(self):
...
+        if self.status == AcqOrderStatus.PENDING:
+            links.pop('acq_order_lines', None)
```
zannkukai, line 189: As mentionned, AcqOrderLine is never a reason why not to delete. But status is a valid reason :
```
links.pop('acq_order_lines', None)
if self.status != AcqOrderStatus.PENDING:
    cannot_delete['others'] = {
        _(f'Order status is {self.status}'): true
    }
```

zannkukai (changes requested):

rero_ils/modules/acq_orders/api.py
BadrAly: as discussed, we separate the deletion from DB and ES.

zannkukai (approved):

lauren-d (approved):

iGormilhit (approved):

rerowep (approved):

PR #2417, opened 2021-09-28
notification: fix loan notification candidates

zannkukai:
When we check about notification candidates on a loan, we need to
exclude the current loan from the current request list on the related
item.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/items/api/circulation.py
+++ b/rero_ils/modules/items/api/circulation.py
@@ -1164,11 +1164,12 @@ class ItemCirculation(ItemRecord):
                 LoanState.ITEM_RETURNED,
             ]).source().count()
 
-    def get_requests(self, sort_by=None, count=False):
+    def get_requests(self, sort_by=None, count=False, pids=False):
         """Return sorted pending, item_on_transit, item_at_desk loans.
 
-        :param sort_by: the sort to appy. default sort is _created.
+        :param sort_by: the sort to result. default sort is _created.
         :param count: if True, return the number of request.
+        :param pids: if True, return the only the pids.
         :return a generator of corresponding request or a request counter.
         """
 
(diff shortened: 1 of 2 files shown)


Garfield-fr (approved):

BadrAly (approved):

rero_ils/modules/loans/api.py
```diff
@@ -761,7 +761,11 @@ def get_notification_candidates(self, trigger):
 
         candidates = []
         item = Item.get_record_by_pid(self.item_pid)
-        has_request = item.number_of_requests() > 0
+        # Get the number of requests on the related item and exclude myself
+        # from the result list.
+        requests = item.get_requests(pids=True)
+        requests = [loan_pid for loan_pid in requests if loan_pid != self.pid]
```
jma, line 767: You can also use filer if you want;
```python
requests = filter(lambda x: x != self.pid, requests)
```

rero_ils/modules/items/api/circulation.py
```diff
@@ -1191,7 +1192,12 @@ def _list_obj():
                 LoanState.ITEM_AT_DESK,
                 LoanState.ITEM_IN_TRANSIT_FOR_PICKUP
             ]).source(['pid'])
-        return query.count() if count else _list_obj()
+        if pids:
+            return [hit.pid for hit in query.execute()]
```
jma, line 1196: you can use `query.scan()` instead of `execute()`. It is more efficient.

jma: Can you also remove the two `print` line 1180 even if it is not a part of you modifications? Thanks.

jma (approved):

iGormilhit (approved):

rero_ils/modules/loans/api.py
zannkukai: list comprehension seems more efficient than filter + lambda. More : filter return a genetator and you need to cast it to a list again. I read somewhere than Python guru (Guido) should have liked to remove `filter` from python3

rero_ils/modules/items/api/circulation.py
zannkukai: done

rerowep (approved):

PR #2528, opened 2021-11-16
holdings/items: remove query filter and add a sort on the holdings

Garfield-fr:
* Adds a sort for holdings.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -2282,6 +2282,11 @@ RECORDS_REST_SORT_OPTIONS['holdings']['library_location'] = dict(
     title='Holdings library location sort',
     default_order='asc'
 )
+RECORDS_REST_SORT_OPTIONS['holdings']['organisation_library_location'] = dict(
+    fields=['organisation.pid', 'library.pid', 'location.pid'],
+    title='Holdings Organisation library location sort',
+    default_order='asc'
+)
 RECORDS_REST_DEFAULT_SORT['holdings'] = dict(
     query='bestmatch', noquery='library_location')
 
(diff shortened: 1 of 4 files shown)


benerken:
- [ ] In the prof. view: when there is no holding/item for my organisation, BUT there is an item for other organisation, could you open these information by default ? 
So, if there is an item in my organisation, display/open it by default
If there is no item in my org but there are items in other org, display/open these one by default  

image

rero_ils/query.py
```diff
@@ -203,11 +203,6 @@ def search_factory_for_all_interfaces(view, search):
             search = search.filter('term', organisation__pid=org['pid'])
         # masked records are hidden for all public interfaces
         search = search.filter('bool', must_not=[Q('term', _masked=True)])
-    # Logic for admin interface
```
jma, line 206: Can you change de function name that suggest to be used by all resources. Something like `search_factory_for_items_and_holdings`. Thanks.

rero_ils/config.py
```diff
@@ -2263,6 +2263,11 @@ def _(x):
     title='Holdings library location sort',
     default_order='asc'
 )
+RECORDS_REST_SORT_OPTIONS['holdings']['organisation_library_location'] = dict(
```
jma, line 2266: Can you mention this in the commit message?

jma (approved):

zannkukai (approved):

BadrAly (approved):

benerken (approved):

PR #2589, opened 2021-12-09
loans: dumping related circulation policy

zannkukai:
Creates an API endpoint to allow dumping the circulation
policies informations related to a loan.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/loans/api_views.py
+++ b/rero_ils/modules/loans/api_views.py
@@ -20,12 +20,13 @@
 from flask import Blueprint, abort, jsonify
 from flask_login import login_required
 
+from rero_ils.modules.decorators import check_logged_as_librarian
 from rero_ils.modules.items.api import Item
 from rero_ils.modules.items.models import ItemCirculationAction
 from rero_ils.modules.items.views.api_views import \
     check_logged_user_authentication, jsonify_error
 from rero_ils.modules.loans.api import Loan
-from rero_ils.modules.loans.utils import sum_for_fees
+from rero_ils.modules.loans.utils import get_circ_policy, sum_for_fees
 
 api_blueprint = Blueprint(
     'api_loan',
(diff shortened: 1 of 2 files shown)


rero_ils/modules/loans/api_views.py
```diff
@@ -34,13 +35,23 @@
 )
 
 
+@api_blueprint.route('/<loan_pid>/circulation_policy', methods=['GET'])
+@check_logged_as_librarian
+def dump_loan_circulation_policies(loan_pid):
```
BadrAly, line 40: I imagine we can only have one policy for each loan: `dump_loan_circulation_policies` -> `dump_loan_circulation_policy`

rero_ils/modules/circ_policies/api.py
```diff
@@ -70,6 +70,7 @@ class CircPolicy(IlsRecord):
     fetcher = circ_policy_id_fetcher
     provider = CircPolicyProvider
     model_cls = CircPolicyMetadata
+    _schema = 'circ_policies/circ_policy-v0.0.1.json'
```
BadrAly, line 73: do we need this? specially it is used only by the `tests`
it is already available by 
`current_app.config.get('RECORDS_JSON_SCHEMA')[provider.pid_type]`

BadrAly (changes requested):
I should also note that this `circulation policy` is not, all the time, the policy used for the `loan`. 
If after a `checkout`, the item has changed its `itty`, `temp_itty` or `loc` then the `api` will return a `new` circ policy which is not really the one used for the `loan`. same applies if `patron` changes its `ptty`.
so this `api` is most of the time correct except the two cases above.

future enhancement: would be nice to have the used `circ_policy` at the time of the `checkout` inside the `loan` object.

BadrAly (approved):

benerken (approved):

PR #2595, opened 2021-12-13
library: add `extra` field for acquisition setting

zannkukai:
Adds a field to allow extra setting (VAT number, special note, ...) for
any informations related to acquisition setting into the `Library`
resource.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/libraries/jsonschemas/libraries/library-v0.0.1.json
+++ b/rero_ils/modules/libraries/jsonschemas/libraries/library-v0.0.1.json
@@ -439,6 +439,12 @@
             },
             "address": {
               "$ref": "#/definitions/address_block"
+            },
+            "extra": {
+              "title": "Extra information",
+              "description": "VAT number, note, ...",
+              "type": "string",
+              "minLength": 3
             }
           }
         },
@@ -470,6 +476,12 @@
             },
             "address": {
               "$ref": "#/definitions/address_block"
+            },
+            "extra": {
+              "title": "Extra information",
+              "description": "VAT number, note, ...",
+              "type": "string",
+              "minLength": 3
             }
           }
         }
--- a/rero_ils/modules/libraries/mappings/v7/libraries/library-v0.0.1.json
+++ b/rero_ils/modules/libraries/mappings/v7/libraries/library-v0.0.1.json
@@ -122,6 +122,9 @@
               "phone": {
                 "type": "text"
               },
+              "extra": {
+                "type": "text"
+              },
               "address": {
                 "properties": {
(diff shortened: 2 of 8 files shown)


vgranata (approved):

rero_ils/modules/libraries/jsonschemas/libraries/library-v0.0.1.json
```diff
@@ -439,6 +439,12 @@
             },
             "address": {
               "$ref": "#/definitions/address_block"
+            },
+            "extra": {
```
BadrAly, line 443: if it is only used for the `notifications`, I think it is better to have some thing like `notification_additional_information` instead of `extra`? 
or simply `additional_information`

rero_ils/modules/libraries/mappings/v7/libraries/library-v0.0.1.json
```diff
@@ -157,6 +160,9 @@
               "phone": {
                 "type": "text"
               },
+              "extra": {
```
BadrAly, line 163: I think the field `extra`  is added `twice` in this file

rero_ils/modules/libraries/jsonschemas/libraries/library-v0.0.1.json
```diff
@@ -439,6 +439,12 @@
...
+              "title": "Extra information",
+              "description": "VAT number, note, ...",
```
BadrAly, line 445: Possibly need to mention that this field is used only in the notifications  and not somewhere else. but not really sure.

BadrAly (changes requested):

rero_ils/modules/libraries/mappings/v7/libraries/library-v0.0.1.json
zannkukai: Yes : one for billing address, one for shipping address. I tried to use a "definition" for both address but as specifications is not perfectly the same I didn't success.

rero_ils/modules/libraries/jsonschemas/libraries/library-v0.0.1.json
zannkukai: At this time, all data from shipping address and billing address sections are only use into notifications.
But could be use somewhere else if needed. I use "extra" as field_id because it's smaller as variable name (and we are limited to 80 chars).

BadrAly (approved):

benerken (approved):

PR #2612, opened 2021-12-20
search: add facets for inventory list

lauren-d:
* Adds temporary item type facet.
* Adds temporary location facet.
* Adds the checkout_date and the due_date fields to the export.
* Closes rero#2364.
* Closes rero#2611.

Co-Authored-by: Laurent Dubois <laurent.dubois@itld-solutions.be>

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#773

## How to test?

1. New added facets
- Login as librarian and go to the Inventory List on professionnal view
- The facet temporary_item_type and temporay_location will being displayed.
- Try and check filtering

2 New fields into the export
- In the Inventory List, try to find items on loan and export the results as csv.
- The checkout_date and the due_date columns must be filled

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/serializers.py
+++ b/rero_ils/modules/serializers.py
@@ -122,6 +122,40 @@ class JSONSerializer(_JSONSerializer):
                     if attr in resource:
                         term[attr] = resource.get(attr)
 
+    @staticmethod
+    def enrich_bucket_with_data(results, bucket_name, search_cls,
+                                attributes_name):
+        """Complete a bucket by adding new keys based on resource attributes.
+
+        :param results: the ES results data (containing aggregations).
+        :param bucket_name: the bucket name to perform.
+        :param search_cls: the related search class.
+        :param attributes_name: attributes to load from search data.
+        """
+        attributes_name = attributes_name or []
+        if not isinstance(attributes_name, list):
+            attributes_name = [attributes_name]
+        buckets = results.get('aggregations', {}) \
+            .get(bucket_name, {}).get('buckets', [])
+
+        # extract pids from buckets
+        bucket_pids = [term['key'] for term in buckets]
+
+        # search all pids with search class
+        query = search_cls() \
+            .filter('terms', pid=list(bucket_pids)) \
+            .source(['pid'] + attributes_name)
+
+        # preprocess results
+        search_data = {result.pid: result.to_dict() for result in
+                       query.scan()}
+
+        # complete bukets with data
+        for term in buckets:
+            for attr in attributes_name:
+                if attr in search_data[term['key']]:
+                    term[attr] = search_data[term['key']].get(attr)
+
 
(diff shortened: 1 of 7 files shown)


rero_ils/modules/serializers.py
```diff
@@ -122,6 +122,41 @@ def complete_bucket_with_attribute(results, bucket_name, resource_cls,
...
+
+        # search all pids with search class
+        results = search_cls() \
+            .filter('terms', pid=list(bucket_pids)) \
+            .source(['pid'] + attributes_name)
+
+        # preprocess results
+        search_data = {}
+        for result in results.scan():
+            search_data[result.pid] = result.to_dict()
```
zannkukai, line 152: more pythonic way
```python
search_data = {result.pid: result.to_dict() for result in results.scan()}
```
maybe the `results` variable could be renamed as `query` ?

zannkukai (approved):

rero_ils/modules/serializers.py
```diff
@@ -122,6 +122,41 @@ def complete_bucket_with_attribute(results, bucket_name, resource_cls,
...
+    def enrich_bucket_with_data(results, bucket_name, search_cls,
+                                attributes_name):
+        """Complete a bucket by adding new keys based on resource attributes.
+
+        :param results: the ES results data (containing aggregations).
+        :param bucket_name: the bucket name to perform.
+        :param search_cls: the related search class.
+        :param attributes_name: attributes to load from search data.
+        """
+        attributes_name = attributes_name or []
```
jma, line 135: this can be done with a default value in params, no?

jma (changes requested):
Tests are missing.

rero_ils/modules/serializers.py
zannkukai: @jma no. You explained a couple month ago that use empty array as default function arguments isn't a good idea because it's only compute once, when the code is 'compile' ;-)

rero_ils/modules/serializers.py
lauren-d: done.

rero_ils/modules/serializers.py
lauren-d: I think such as @zannkukai.
https://docs.python.org/3/reference/compound_stmts.html#:~:text=Default%20parameter%20values%20are,the%20function%2C%20e.g.%3A
but maybe, we are wrong

jma: @zannkukai you are right, but in this case it should be ok as the default value is a constant. The problem happens when for example you want to set the default value as the current date.

jma (approved):

ManaDeweerdt:
It seems that the due date and the checkout date are most of the times empty.
Even for items filtered with the facet 'on loan'.

benerken (approved):

Issue #1530, opened 2020-12-01
circulation: block checkout/renew/request when patron has unpaid subscription


pronguen:
**How it works**

In case of a subscription, there is no blocking.

**Improvement suggestion**

A patron with unpaid subscription fees is blocked (checkout/renew/request). The corresponding message is displayed in the public patron account, circulation UI after trying to do a transaction, pro patron account.

To clarify: have a setting in the patron type to choose to block or not the patrons with an unpaid subscription?

*Note: the end of the subscription is not linked to the expiration date.*

iGormilhit:
I my opinion:

- X days after the end of the subscription, circulation should be blocked. This should be set in a circulation policy.
- When a patron account reach the expiry date, the account will be reactivate
    - maybe the next time the patron interact with the online acount
    - or when the patron goes to the circulation desk.
    - X months after the deactivation of the patron account, if the patron didn't do anything, the account is deleted.

Of course some notifications should be sent.

zannkukai:
@pronguen is this new restrictions should be activated by default or not ?

pronguen:
@zannkukai I would say yes: it is activated by default when creating a patron type.

PR #2644, opened 2022-01-19
patron_types: add a limit for unpaid subscription

zannkukai:
Add a circulation operation restriction if a patron has pending unpaid
subscriptions. If this limit is enabled, any output circulation
operation should be denied (checkout/renew/request).
Closes rero/rero-ils#1530.

## Migration

An alembic script is available to migrate existing PatronTypes : `90d857fb5c23_unpaid_subscription_limit`

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/items/views/api_views.py
+++ b/rero_ils/modules/items/views/api_views.py
@@ -379,10 +379,11 @@ def item(item_barcode):
         if circ_policy.can_checkout:
             for action in item_dumps.get('actions', []):
                 if action == 'checkout':
-                    if item.number_of_requests() > 0:
-                        if item.patron_request_rank(patron) == 1:
-                            new_actions.append(action)
-                    else:
+                    if (
+                        item.number_of_requests() > 0
+                        and item.patron_request_rank(patron) == 1
+                        or item.number_of_requests() <= 0
+                    ):
                         new_actions.append(action)
                 elif action == 'receive' and item.number_of_requests() == 0:
                     new_actions.append('checkout')
--- a/tests/data/data.json
+++ b/tests/data/data.json
@@ -1192,6 +1192,9 @@
     "description": "children patrons",
     "organisation": {
       "$ref": "https://bib.rero.ch/api/organisations/org1"
+    },
+    "limits": {
+      "unpaid_subscription": false
     }
   },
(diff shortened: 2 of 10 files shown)


rero_ils/modules/items/views/api_views.py
```diff
@@ -379,10 +379,11 @@ def item(item_barcode):
...
             for action in item_dumps.get('actions', []):
                 if action == 'checkout':
-                    if item.number_of_requests() > 0:
-                        if item.patron_request_rank(patron) == 1:
-                            new_actions.append(action)
-                    else:
+                    if (
+                        item.number_of_requests() > 0
+                        and item.patron_request_rank(patron) == 1
+                        or item.number_of_requests() <= 0
```
jma, line 385: This seems not related to the commit message: looks like to be a fix, no? If yes can you add an entry in the commit message?

tests/data/data.json
```diff
@@ -1192,6 +1192,9 @@
     "description": "children patrons",
     "organisation": {
       "$ref": "https://bib.rero.ch/api/organisations/org1"
+    },
+    "limits": {
+      "unpaid_subscription": false
```
jma, line 1197: Is it required? If yes this needs data migration steps.

jma (approved):
You can add in the commit message: several refactoring. Thanks for all.

tests/data/data.json
zannkukai: @jma I include a alembic script to migrate existing PatronTypes ;-)

rero_ils/modules/items/views/api_views.py
zannkukai: OK

rerowep (approved):

lauren-d (approved):

ManaDeweerdt:
- [ ] Editing the patron type, I can see that the limit is activated but it is displayed like this.
image

zannkukai:
> * The option should be displayed closed to the subscription activation, and ideally should be displayed only if subscription are activated...

I can't move the options outside of the limits block ; but I placed it at first position and only visible if the `subscription_amount` is > 0

PR #2645, opened 2022-01-19
notification: fix `reply-to` recipient

zannkukai:
The `reply-to` recipients should be an array but CirculationNotification
classes returned only a string. This caused bad `reply-to` message header
only containing the first letter of the email.

## How to test?

before
image

after
image

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/notifications/dispatcher.py
+++ b/rero_ils/modules/notifications/dispatcher.py
@@ -155,7 +155,7 @@ class Dispatcher:
             template_body=template,
             sender=current_app.config.get('DEFAULT_SENDER_EMAIL',
                                           'noreply@rero.ch'),
-            reply_to=reply_to,
+            reply_to=','.join(reply_to),  # the client is unable to manage list
             recipients=recipients,
             ctx=ctx_data
         )
@@ -236,7 +236,7 @@ class Dispatcher:
         # 3. Send the message
         msg = Dispatcher._create_email(
             recipients=[recipient],
-            reply_to=reply_to,
+            reply_to=[reply_to],
             ctx_data=context,
             template=notification.get_template_path()
         )
@@ -255,7 +255,7 @@ class Dispatcher:
             return True
 
         notification = notifications[0]
-        reply_to = notification.get_recipients(RecipientType.REPLY_TO)[0]
+        reply_to = notification.get_recipients(RecipientType.REPLY_TO)
         recipients = notification.get_recipients(RecipientType.TO)
 
         error_reasons = []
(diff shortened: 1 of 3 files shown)


benerken (approved):

lauren-d (approved):

rero_ils/modules/notifications/dispatcher.py
```diff
@@ -155,7 +155,7 @@ def _create_email(recipients, reply_to, ctx_data, template):
             template_body=template,
             sender=current_app.config.get('DEFAULT_SENDER_EMAIL',
                                           'noreply@rero.ch'),
-            reply_to=reply_to,
+            reply_to=','.join(reply_to),  # the client is unable to manage list
```
rerowep, line 158: I'am not sure this is correct. Excample:
`['person1@gmail.com', 'person2@gmail.com']` after the join we will have a string with `'person1@gmail.com, person2@gmail.com'`

rero_ils/modules/notifications/dispatcher.py
```diff
@@ -236,7 +236,7 @@ def send_mail_for_printing(notifications=None):
         # 3. Send the message
         msg = Dispatcher._create_email(
             recipients=[recipient],
-            reply_to=reply_to,
+            reply_to=[reply_to],
```
rerowep, line 239: Here we will put the string into a list and get:
`['person1@gmail.com, person2@gmail.com']`
A list with one string. Should we not use ` isinstance(reply_to, list)` or always put a list into `reply_to?

rerowep (changes requested):

rero_ils/modules/notifications/dispatcher.py
zannkukai: @rerowep 
According to RFC5322 : "In either case, an optional reply-to field MAY also be included, which contains the field name "Reply-To" and a comma-separated list of one or more addresses."

So it seems correct for me. It should better to use list, but client (flask-mail) works fine with list for recipients, but not for reply-to :(

rero_ils/modules/notifications/dispatcher.py
zannkukai: in this method, I was sure than `reply_to = library.get('email')` return a string (Check library JSON schema). To work same between "send_by_mail" and "send_for_printing", I force "reply_to" to be a list containing only one address.

rerowep (approved):

PR #2706, opened 2022-02-16
holding: fix delete all items for issue

rerowep:

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/holdings/api.py
+++ b/rero_ils/modules/holdings/api.py
@@ -183,7 +183,7 @@ class Holding(IlsRecord):
         if can:
             if self.is_serial:
                 # Delete all attached items
-                for item in self.get_items:
+                for item in self.get_all_items():
                     item.delete(
                         force=force, dbcommit=dbcommit, delindex=False)
             return super().delete(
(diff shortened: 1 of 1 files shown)


rero_ils/modules/holdings/api.py
```diff
@@ -183,7 +183,8 @@ def delete(self, force=False, dbcommit=False, delindex=False):
         if can:
             if self.is_serial:
                 # Delete all attached items
-                for item in self.get_items:
+                for item_pid in Item.get_items_pid_by_holding_pid(self.pid):
+                    item = Item.get_record_by_pid(item_pid)
```
zannkukai, line 187: I understand why you change this code.
But my feeling is that `get_items()` functions  should return all related items and not just standard or received items. Should be better to create a property `def items()` returning all items related to this holdings and a function `def get_items(filters)` that will return only items matching criteria.

So in your code only change `for item in self.get_items` to `for item in self.items`.

Garfield-fr (approved):

rero_ils/modules/holdings/api.py
rerowep: Actually the command of the `get_items` function says: """Return standard items and received issues for a holding record.""". I will implement a function `items` as demanded.

zannkukai (approved):

Issue #2838, opened 2022-04-11
E-mail validation must be stricter


pronguen:
### How it works
It is possible to enter wrong formatted e-mails into RERO ILS for users/patrons.

### Improvement suggestion

In the user editor, professional view (user e-mail + patron additional communication e-mail) + in the public view (user e-mail), the validation should be stricter:
- no spaces in the beginning or the end (or they are automatically trimmed)
- no spaces in the e-mail
- no dot at the end, or other punctuation character
- only one `@`
- eventually the system checks if the domain name exists

PascalRepond:
Linked to https://github.com/rero/rero-ils/issues/2765

Issue #2765, opened 2022-03-15
Add user form validations


jma:
### Bug description:
The form to create or edit a user do not contains enough validation: this can cause and database error. I.e. long postal code. See sentry: https://sentryils.rero.ch:8443/share/issue/527d34f744a94555b152418b3b14b5e5/ for more details.

### Expected behavior:
It should not be possible to push non valid data.
__Note:__ other validation should be fixed: 

- submit a patron without a barcode, 
- submit a form with a communication email without email defined either in the user profile or the patron metadata.
- Validate email format ! I can currently save a completely wrong format : 
![image](https://user-images.githubusercontent.com/50177009/162693749-494eb972-fc1e-4fde-98f6-ea37ff85167a.png)

### Steps to Reproduce:

**Server:**
* [ x ] production: [bib.rero.ch](https://bib.rero.ch) or [ils.bib.uclouvain.be](https://ils.bib.uclouvain.be)
* [ x ] test: [bib.test.rero.ch](https://bib.test.rero.ch)
* [ x ] dev: [ilsdev.test.rero.ch](https://ilsdev.test.rero.ch)
* [ x ] dev test : [ils.test.rero.ch](https://ils.test.rero.ch)

**Version:**

`v1.9.0` ; commit hash : `XXXX`

**Browser:**

Chrome, Safari, Firefox; version XX

PR #2861, opened 2022-04-27
resource: improve data validation against JSONSchema

zannkukai:
* Adds a better 'email' format validation for all RERO-ILS resources.
* Simplify JSON schema defining fields with 'email' format.
* Adds length constraints for some `User` resource fields.
* Closes rero/rero-ils#2765.
* Closes rero/rero-ils#2838.

## Code review check list

- [ ] File names.
- [ ] Functions names.
- [ ] Functions docstrings.

### DIFF
--- a/rero_ils/modules/api.py
+++ b/rero_ils/modules/api.py
@@ -17,7 +17,7 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 """API for manipulating records."""
-
+import re
 from copy import deepcopy
 from uuid import uuid4
 
@@ -39,12 +39,31 @@ from invenio_records.api import Record
 from invenio_records_rest.utils import obj_or_import_string
 from invenio_search import current_search
 from invenio_search.api import RecordsSearch
+from jsonschema import FormatChecker
+from jsonschema.compat import str_types
 from jsonschema.exceptions import ValidationError
 from kombu.compat import Consumer
 from sqlalchemy import text
 from sqlalchemy.orm.exc import NoResultFound
 
 from .utils import extracted_data_from_ref
+
+"""Custom ILS record JSON schema format validator."""
+ils_record_format_checker = FormatChecker()
+
+
+@ils_record_format_checker.checks('email')
+@ils_record_format_checker.checks('idn-email')
+def _strong_email_validation(instance) -> bool:
+    """Allow to validate an email address (only email format, not DNS)."""
+    if not isinstance(instance, str_types):
+        return False
+    # DEV NOTE : If this regexp changes, you also need to alter the regexp
+    # into `email.validator.ts` file into @rero/ng-core. The best solution
+    # should be to use a configuration setting available through an API, but
+    # it should be a little overkill.
+    email_regexp = r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'
+    return bool(re.fullmatch(email_regexp, instance))
 
 
 class IlsRecordError:
@@ -111,6 +130,8 @@ class IlsRecord(Record):
     pids_exist_check = None
     pid_check = True
 
+    format_checker = ils_record_format_checker
+
     @classmethod
     def get_indexer_class(cls):
         """Get the indexer from config."""
@@ -134,6 +155,7 @@ class IlsRecord(Record):
         if self.get('_draft'):
             # No validation is needed for draft records
             return self
+
         json = super()._validate(**kwargs)
         validation_message = self.extended_validation(**kwargs)
         # We only like to run pids_exist_check if validation_message is True
@@ -168,24 +190,22 @@ class IlsRecord(Record):
         assert cls.provider
         if '$schema' not in data:
             pid_type = cls.provider.pid_type
-            schemas = current_app.config.get('RECORDS_JSON_SCHEMA')
+            schemas = current_app.config.get('RERO_ILS_DEFAULT_JSON_SCHEMA')
             if pid_type in schemas:
                 from .utils import get_schema_for_resource
                 data['$schema'] = get_schema_for_resource(pid_type)
         pid = data.get('pid')
         if delete_pid and pid:
             del data['pid']
-        else:
-            if pid:
-                test_rec = cls.get_record_by_pid(pid)
-                if test_rec is not None:
-                    raise IlsRecordError.PidAlreadyUsed(
-                        'PidAlreadyUsed {pid_type} {pid} {uuid}'.format(
-                            pid_type=cls.provider.pid_type,
-                            pid=test_rec.pid,
-                            uuid=test_rec.id
-                        )
+        elif pid:
+            if test_rec := cls.get_record_by_pid(pid):
+                raise IlsRecordError.PidAlreadyUsed(
+                    'PidAlreadyUsed {pid_type} {pid} {uuid}'.format(
+                        pid_type=cls.provider.pid_type,
+                        pid=test_rec.pid,
+                        uuid=test_rec.id
                     )
+                )
         if not id_:
             id_ = uuid4()
         persistent_identifier = cls.minter(id_, data)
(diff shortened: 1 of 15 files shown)


rero_ils/modules/users/api.py
```diff
@@ -59,6 +59,8 @@ def get_readonly_profile_fields():
 class User(object):
     """User API."""
 
+    _default_schema = 'https://bib.rero.ch/schemas/users/user-v0.0.1.json'
```
BadrAly, line 62: Any reason, why this is not in `config.py` like the other resources => `RECORDS_JSON_SCHEM`?

zannkukai: because this isn't a real Invenio resource.
But you're right, I will create a function to build this URL based on values from config
👍

rero_ils/modules/api.py
```diff
@@ -39,13 +39,32 @@
...
+
+
+@ils_record_format_checker.checks('email')
+@ils_record_format_checker.checks('idn-email')
+def _strong_email_validation(instance) -> bool:
+    """Allow to validate an email address (only email format, not DNS)."""
+    if not isinstance(instance, str_types):
+        return False
+    email_regexp = DEFAULT_EMAIL_REGEXP
+    if current_app:  # could be use outside of the application context
```
rerowep, line 63: Would it not be easier to import `EMAIL_VALIDATION_REGEXP` directly from the config.py file?

rero_ils/modules/patrons/api.py
```diff
@@ -180,8 +180,7 @@ def _validate_emails(self):
         Check if the user has at least one email if the communication channel
         is email.
         """
-        patron = self.get('patron')
-        user = self._get_user_by_user_id(self.get('user_id'))
+        patron, user = self.get('patron'), self.user
```
rerowep, line 183: Two lines would be more clear:
```python
patron = self.get('patron')
user = self.user
```

rero_ils/modules/patrons/api.py
```diff
@@ -180,8 +180,7 @@ def _validate_emails(self):
...
         if patron and patron.get('communication_channel') == 'email'\
                 and user.email is None\
```
rerowep, line 185: If we use `user` only on this place we can delete `user = self.user` above and write `self.user.email`

rero_ils/modules/api.py
zannkukai: If we put the DEFAULT_EMAIL_REGEXP into the `config.py` the app context should be required. Isn't it ?

jma: > If we put the DEFAULT_EMAIL_REGEXP into the `config.py` the app context should be required. Isn't it ?

Yes but you can have two different behaviour depending if you have the app context or not...

rero_ils/modules/api.py
```diff
@@ -175,17 +197,16 @@ def create(cls, data, id_=None, delete_pid=False,
...
-            if pid:
-                test_rec = cls.get_record_by_pid(pid)
-                if test_rec is not None:
-                    raise IlsRecordError.PidAlreadyUsed(
-                        'PidAlreadyUsed {pid_type} {pid} {uuid}'.format(
-                            pid_type=cls.provider.pid_type,
-                            pid=test_rec.pid,
-                            uuid=test_rec.id
-                        )
+        elif pid:
```
jma, line 200: this seems not be in the commit message

rero_ils/modules/libraries/jsonschemas/libraries/library-v0.0.1.json
```diff
@@ -465,9 +457,7 @@
             "email": {
               "title": "Contact email",
               "type": "string",
-              "format": "email",
-              "pattern": "^.*?@.*.*$",
-              "minLength": 6
+              "format": "email"
```
jma, line 460: great!

rero_ils/modules/patrons/extensions.py
```diff
@@ -35,5 +35,5 @@ def pre_dump(self, record, data, dumper=None):
         :return the future dumped data.
         """
         user = User.get_by_id(record.get('user_id'))
-        user_info = user.dumpsMetadata()
+        user_info = user.dumps_metadata()
```
jma, line 38: sorry for that, it was probably my own

rero_ils/modules/users/api.py
jma: You can access the config from the console without any problem.

rero_ils/modules/users/api.py
```diff
@@ -152,13 +154,12 @@ def update(self, data):
...
         """Validate user record against schema."""
-        format_checker = FormatChecker()
-        schema = data.pop('$schema', None)
+        schema = data.pop('$schema', User._default_schema)
         if schema:
             _records_state.validate(
                 data,
                 schema,
-                format_checker=format_checker,
+                format_checker=ils_record_format_checker,
```
jma, line 162: This can be a candidate to a small presentation: 5' after a daily.

rero_ils/modules/users/api.py
```diff
@@ -177,18 +178,23 @@ def get_by_id(cls, user_id):
...
+    def dumps_metadata(self, dump_patron: bool = False) -> dict:
+        """Dumps the profile, email, roles metadata.
+
+        :param dump_patron: is the patron metadata should be dumped.
+        :return a dictionary with all dump user metadata.
+        """
         from ..patrons.api import Patron
-        metadata = {}
+        metadata = {
+            'roles': [r.name for r in self.user.roles]
```
jma, line 196: Should it be in the commit message?

rero_ils/modules/users/jsonschemas/users/user-v0.0.1.json
```diff
@@ -37,10 +37,15 @@
...
     },
+    "pid": {
+      "title": "User ID",
+      "type": "number"
+    },
     "first_name": {
       "title": "First name",
       "type": "string",
       "minLength": 1,
+      "maxLength": 255,
```
jma, line 48: Be sure that all the common fields with invenio-userprofile has the same contraints.

jma (changes requested):

rero_ils/modules/users/jsonschemas/users/user-v0.0.1.json
zannkukai: I checked with the model used by SQLAlchemy (from invenio_userprofile)

rero_ils/modules/api.py
zannkukai: use use "elif" instead of "else" + "if"

rero_ils/modules/api.py
zannkukai: After concertation, I choose to place the email validation regexp into the function.
The reason is that it seems not relevant (not highly needed) to have a special configuration to validate email using a config setting. Placing this setting into the config forces to include `app` fixture to validate json schema using an 'email' field.
This solution is still much better than the current `jsonschema` library basic behavior : validate anything except if is a string. In this case, the validation only requires an "@" character.
```python
@_checks_drafts(name="idn-email")
@_checks_drafts(name="email")
def is_email(instance):
    if not isinstance(instance, str_types):
        return True
    return "@" in instance
 ```

**Review by jma — APPROVED** [review]

**Review by BadrAly — APPROVED** [review]

File: rero_ils/modules/utils.py [hunk 11]
```diff
@@ -409,18 +408,17 @@ def pid_exists(info, pid_type, pid, raise_on_error=False):
     :param raise_on_error: Raise PidDoesNotExist exception if enabled.
     :return: True if pid was found. Otherwise False.
     """
-    if PersistentIdentifier.query.filter_by(
-        pid_type=str(pid_type), pid_value=str(pid)
-    ).count() == 1:
+    if PersistentIdentifier.query\
```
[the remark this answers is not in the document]
zannkukai: I can't because if the query isn't true, we need to raise an `IlsRecordError` (according to method argument)

ManaDeweerdt:
- [ ] There is no warning message below the field Email if we enter a wrong formated email. [nice to have?]

PascalRepond (approved):

PR #2916, opened 2022-05-16
notifications: fix random unitest

zannkukai:
Some unitest are randomize to to the fact that AT_DESK and AVAILABILITY
notification could be operated during the same process cycle. So we
can't determine which notification is the last one received in the
mailbox. To solve this problem, we need to check into the two last
messages.

### DIFF
--- a/tests/api/notifications/test_notifications_rest.py
+++ b/tests/api/notifications/test_notifications_rest.py
@@ -918,8 +918,10 @@ def test_multiple_request_booking_notifications(
         )
     )
     assert res.status_code == 200
+    mailbox.clear()
 
-    # checkout for patron1
+    # CHECKOUT FOR REQUEST#1
+    #     After the checkout, no new notification will be sent.
     params = {
         'patron_pid': patron_martigny.pid,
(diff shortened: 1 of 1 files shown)


tests/api/notifications/test_notifications_rest.py
```diff
@@ -936,8 +936,13 @@ def test_multiple_request_booking_notifications(
     # checkin at the request pickup of patron2
     _, actions = item_lib_martigny.checkin(**params)
     assert actions.get(LoanAction.CHECKIN)
-    assert f'Lieu de retrait: {loc_public_sion.get("code")}'\
-        in mailbox[-1].body
+    # Checkin operation will generate 2 notifications : AT_DESK and
+    # AVAILABILITY. As both are generated during the same process cycle, we
+    # can't know if the availability is the last one or the second to last
+    # mail --> check in both !
+    notifications_content = [mailbox[-1].body, mailbox[-2].body]
```
rerowep, line 943: Do we have to test there are a minimum of two emails in the mailbox?

benerken (approved):

lauren-d (approved):

PR #2945, opened 2022-06-08
tests: fix test_loan_operation_log

rerowep:

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

### DIFF
--- a/rero_ils/modules/patrons/api.py
+++ b/rero_ils/modules/patrons/api.py
@@ -18,7 +18,7 @@
 
 """API for manipulating patrons."""
 from copy import deepcopy
-from datetime import datetime
+from datetime import date, datetime
 from functools import partial
 
 from elasticsearch_dsl import Q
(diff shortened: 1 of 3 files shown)


rero_ils/modules/patrons/api.py
```diff
@@ -825,6 +825,17 @@ def basic_query(channel):
             process_bulk_queue.apply_async()
 
 
+    def get_age(self):
```
zannkukai, line 828: This method could be a `@property` of patron.
```
@property
def age(self):
    ...
```

zannkukai (changes requested):

rero_ils/modules/patrons/api.py
rerowep: Ok

Garfield-fr (approved):

lauren-d (approved):

zannkukai (approved):

PR #2955, opened 2022-06-10
resources: refactoring serialization

zannkukai:
* Creates a post-processor mixin to simplify the search serialization
  process. Now a `ReroIlsJSONSerializer` could implements three specific
  method to alter a search result function :
    * one to alter a search result hit.
    * one to alter aggregations/facets.
    * one to alter search results links.
* Improves serializer classes code coverage.
* Updates existing JSONSerializer to use this new behavior.


rerowep (changes requested):

rero_ils/modules/acq_accounts/serializers.py
```diff
@@ -20,32 +20,29 @@
 from invenio_records_rest.serializers.response import record_responsify, \
     search_responsify
 
+from rero_ils.modules.libraries.api import LibrariesSearch
```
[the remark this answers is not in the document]
jma: For information invenio uses relative imports: https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py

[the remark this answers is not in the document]
zannkukai: My personal opinion is that above 2 parents relative import are difficult to read.
Also, when I use classes/object/function from other package, I think it's better to use absolute import.
Need to be discussed with all devs after a daily.

rero_ils/modules/documents/serializers/dc.py
```diff
@@ -16,21 +16,25 @@
...
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-"""DublinCore Document serialization."""
+"""RERO Document JSON DublinCore serialization."""
 
 from dcxml import simpledc
 from flask import current_app, request
 from invenio_records_rest.serializers.dc import \
-    DublinCoreSerializer as BaseDublinCoreSerializer
+    DublinCoreSerializer as _DublinCoreSerializer
```
lauren-d, line 24: cosmetic: in my opinion, it is less telling

rero_ils/modules/acq_orders/serializers.py
```diff
@@ -19,43 +19,43 @@
...
+        )
+        JSONSerializer.enrich_bucket_with_data(
+            aggregations.get('account', {}).get('buckets', []),
+            AcqAccountsSearch, 'name'
+        )
         # Add configuration for order_date and receipt_date buckets
-        for bucket in ['order_date', 'receipt_date']:
-            aggr = results['aggregations'].get(bucket)
+        for aggr_name in ['order_date', 'receipt_date']:
+            aggr = aggregations.get(aggr_name)
```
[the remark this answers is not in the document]
zannkukai: using your proposition, we add a tabulation/4_spaces on all lines into the statement.
Using the proposed solution, if "aggr" is missing from data, we just return. Seems ok and readable for me.

rero_ils/modules/serializers/base.py
```diff
@@ -0,0 +1,192 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""RERO ILS record serialization."""
+import inspect
+from abc import ABC
+
+from flask import json, request, url_for
+from invenio_records_rest.serializers.json import \
+    JSONSerializer as _JSONSerializer
```
lauren-d, line 24: cosmetic: same before `BaseJSONSerializer`?

rero_ils/modules/serializers/base.py
```diff
@@ -0,0 +1,192 @@
...
+                    term[attr] = data[term['key']].get(attr)
+
+
+class CachedDataSerializerMixin:
+    """Class to load and cached resources for serialization process."""
+
+    _resources = {}
+
+    def get_resource(self, loader, pid):
+        """Get a resource and store it into the cache if necessary.
```
lauren-d, line 178: Is the cache expired (or resets) after each serializations ?
If not, I think this is a problem when a resource is updated.

zannkukai: The cache is only apply for the object life cycle. As the object is only use for one serialization, it seems good.

rero_ils/modules/serializers/base.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def _postprocess_search_aggregations(self, aggregations: dict) -> None:
+        """Post-process aggregations from a search result.
+
+        :param aggregations: the dictionary representing ElasticSearch
+                             aggregations section.
+        """
+        if 'document_type' in aggregations:
+            aggr = aggregations['document_type'].get('buckets')
+            filter_document_type_buckets(aggr)
```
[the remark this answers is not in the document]
zannkukai: because, in any cases (for RERO-ILS serializing), if the search contains a `document_type` facets, this facets must be cleaned. I think it was often use with Virtua loaded records ; Now it's impossible to enter an invalid subtype against the JSONSchema validation.

rero_ils/modules/documents/serializers/dc.py
zannkukai: We use "_" syntax now for other resource that we could override. See: https://github.com/rero/rero-ils/blob/staging/rero_ils/modules/serializers.py#L21-L24 (staging branch).
I just try to harmonize imports... and it's an huge job !

lauren-d (approved):

BadrAly (approved):

PR #3087, opened 2022-09-12
fixtures: fix bulk_save

rerowep:
* Excludes operation logs from bulk save.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

### DIFF
--- a/rero_ils/modules/cli/fixtures.py
+++ b/rero_ils/modules/cli/fixtures.py
@@ -493,7 +493,7 @@ def bulk_save(pid_types, output_directory, deployment, verbose):
             )
             continue
         # TODO: do we have to save loanid and how we can save it?
-        if p_type == 'loanid':
+        if p_type in ['loanid', 'oplg']:
             continue
         click.secho(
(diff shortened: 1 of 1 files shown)


Garfield-fr (approved):

rero_ils/modules/cli/fixtures.py
```diff
@@ -493,7 +493,7 @@ def bulk_save(pid_types, output_directory, deployment, verbose):
             )
             continue
         # TODO: do we have to save loanid and how we can save it?
-        if p_type == 'loanid':
+        if p_type == 'loanid' or p_type == 'oplg':
```
zannkukai, line 496: more condensed code (and more readable ?)
```python
if p_type in ['loanid', 'oplg']:
```
Additionally I don't like use fixed-string in such case because it will be changed into config and we need to cascade this change in any code. In a better world, `pid_type` could be listed into an enum class and used into config (as rest_endpoint_key) and here. **But this it's not required in this PR ;-)**

zannkukai (approved):

PR #3102, opened 2022-09-20
loans: warns for loans not changing status

### LINKED ISSUE #3100: Some loan stay with `CREATED` temporary state.
### Bug description:
In some special cases, some loan are saved with `CREATED` state. This state should only be a temporary state when a `Loan` object is create, but must only transit to `PENDING` or `ITEM_ON_LOAN`. This should never happens

### Expected behavior:
No `CREATED` loan should be saved avec any circulation operation.

### Steps to Reproduce:
don't really how to reproduce.

### Context

**Version:**
`v1.13.0`

### Anything else?
No circulation operations are directly linked with `Loan` REST_API, all API calls are linked to `Item` API (see `items.views.api_views.py)`.
In some cases, the `add_action_parameters_and_flush_indexes` decorator create a new loan object (https://github.com/rero/rero-ils/blob/staging/rero_ils/modules/items/decorators.py#L61 , https://github.com/rero/rero-ils/blob/staging/rero_ils/modules/items/api/circulation.py#L159). But, if no operation are done on this loan due to `invenio_circulation` triggers, then this loan stay on `CREATED` state and should be removed.

I will be interesting to create a system to check if the loan state changed after some circulation operation (request, checkin,checkout, ... not extend [status stay `ITEM_ON_LOAN`]). If it's the case, an error will be logged in Sentry with request context (loan, patron, itty, ...) for future analyzes.

After any circulation operation, if a new loan status is `CREATED` an exception should be raised and this loan should be deleted.
_The best solution should be to not commit/reindex a loan during the circulation process but commit all changes at the end of action process (or rollback if any errors are detected) but this behavior change will be more challenging._

rerowep:
* Adds cli and task for deleting loans with status `CREATET`.
* Closes #3100.

* Adds your new feature.
* Fixes an existing issue, may closes #issue.
* Improves and existing feature.

Co-Authored-by: Peter weber <peter.weber@rero.ch>

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?


PascalRepond (changes requested):
Please fix your commit message, delete three last bullet points.

rero_ils/modules/loans/tasks.py
```diff
@@ -67,3 +70,29 @@ def cancel_expired_request_task(tstamp=None):
...
                   cancelled=total_cancelled_loans)
     return total_loans_counter, total_cancelled_loans
+
+
+@shared_task(ignore_result=True)
+def delete_loans_created(verbose=False, hours=1, dbcommit=True, delindex=True):
+    """Delete loans with state CREATED."""
+    now = datetime.now(timezone.utc)
+    if hours > 0:
+        now += timedelta(hours=hours)
```
zannkukai, line 80: are you sure about the "+=" ? It should be "-=" isn't it ?

rero_ils/modules/loans/tasks.py
```diff
@@ -67,3 +70,29 @@ def cancel_expired_request_task(tstamp=None):
...
+        .filter('term', state='CREATED').filter('range', _created={'lt': now})
+    if verbose:
+        click.echo(
+            f'TOTAL: {count} DELETE: {query.count()} HOURS: {-query.count()}'
+        )
+    idx = 0
+    for idx, hit in enumerate(query.source('pid').scan(), 1):
+        loan = Loan.get_record_by_pid(hit.pid)
+        state = loan.get('state')
+        if state == 'CREATED':
```
zannkukai, line 92: This IF statement seems useless because you're already filter your query on this state.

PascalRepond (approved):

lauren-d (commented):
there is a typo in commit message: `CREATET` => `CREATED`

rero_ils/modules/loans/tasks.py
zannkukai: You "resolved" but without answer. 
In my function comprehension (docstring is poor), you would delete all loans with status created AND creation date is from [* TO now-1h]. So for me your timedelta will be "-="

benerken (approved):

rero_ils/modules/loans/tasks.py
rerowep: Sorry, corrected it.

rero_ils/config.py
```diff
@@ -435,13 +435,18 @@ def _(x):
...
+    'delete-loans-created': {
+        'task': 'rero_ils.modules.loans.tasks.delete_loans_created',
+        'schedule': crontab(minute=0, hour=5), # Every day at 05:00 UTC,
+        'enabled': False,
+    },
     # 'mef-harvester': {
     #     'task': 'rero_ils.modules.apiharvester.tasks.harvest_records',
     #     'schedule': timedelta(minutes=60),
-    #     'kwargs': {'name': 'mef', 'enabled': False),
+    #     'kwargs': {'name': 'mef', 'enabled': False),setup
```
jma, line 446: Is it a mistake?

jma (approved):

rero_ils/config.py
rerowep: Mistake!

PR #3110, opened 2022-09-26
document: fix RIS export

jma:

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?


lauren-d (approved):
This will work, but it may be better to rename the property in the `__init__.py than to use a private property.

PascalRepond (approved):

benerken (approved):

PR #3155, opened 2022-11-09
document: document with item and or local field

rerowep:
* Cli for creation and validation for documents with items and or local fields.


rero_ils/modules/cli/utils.py
```diff
@@ -281,42 +297,66 @@ def add_org_lib_doc(item):
                 f'Error validate in document: {count} {pid} {trace}',
                 fg='red'
             )
+            from pprint import pprint
+            pprint(record)
```
zannkukai, line 301: Is it for debug ? If yes, you need to remove it

zannkukai (approved):

lauren-d (approved):

PR #3170, opened 2022-11-30
agents: add MEF synchro.

jma:

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?


lauren-d (approved):
Greats work !
I just have a general question. Could we use sync with long terms queries?
Maybe, it's better to write method with pids list in argument. 
The idea is to launch multiple sync in the same times using the new generic task from rero-invenio-base 😁

rero_ils/modules/contributions/api.py
```diff
@@ -77,12 +77,14 @@ def get_contribution(cls, ref_type, ref_pid):
...
         if ref_type == 'mef':
             return cls.get_record_by_pid(ref_pid)
+        # in case of multiple results get the more recent
+        query = ContributionsSearch() \
+            .params(preserve_order=True) \
+            .sort({'_created': {'order': 'desc'}})
         if ref_type == 'viaf':
-            query = ContributionsSearch() \
-                .filter('term', viaf_pid=ref_pid)
+            query = query.filter('term', viaf_pid=ref_pid)
```
zannkukai, line 85: you could use `&=` statement 😝
```query &= Q('term', viaf_pid=ref_pid)```

File: rero_ils/modules/contributions/sync.py [hunk 2]
```diff
@@ -0,0 +1,546 @@
...
+        if not log_dir:
+            log_dir = current_app.config.get(
+                'RERO_ILS_MEF_SYNC_LOG_DIR',
+                os.path.join(current_app.instance_path, 'logs')
+            )
+        # create the log directory if does not exists
+        if not os.path.exists(log_dir):
+            os.mkdir(log_dir)
+        verbose_level = ['ERROR', 'INFO', 'DEBUG']
+        logging_config = {
```
zannkukai, line 79: maybe it should be part of a `conf/logging.py` file into this module ? just a suggestion

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+        self.verbose = verbose
+        self.from_date = None
+        self.start_timestamp = None
+        self._init_logger(verbose, log_dir)
+        if from_last_date:
+            self._get_last_date()
+
+    def _get_last_date(self):
+        """Get the date of the last execution of the synchronization."""
+        self.from_date = None
```
zannkukai, line 62: should be moved into `__init__` method

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+                    'level': 'INFO',
+                    'propagate': False
+                }
+            }
+        }
+        logging.config.dictConfig(logging_config)
+        self.logger = logging.getLogger(__name__)
+        self.log_dir = log_dir
+
+    def agent_are_different(self, agent1, agent2):
```
zannkukai, line 115: Multiple remarks :
  * is it a good candidate for `__cmp__` special method implementation ?
  * if you choose to use 2 agents parameters, this method should be a `@classmethod` (move logging ?)
  * if you would keep self (for logging), use only one `agent` params.
  
  As you wrote it, it's just like you have 1 apple to compare 2 other apples

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+        # get all agents from the document over all agent fields:
+        # contribution and subjects
+        agents = [
+            subject for subject in doc.get('subjects', [])
+            if subject.get('$ref')
+        ]
+        agents += [
+            c['agent'] for c in doc.get('contribution', [])
+            if c.get('agent', {}.get('$ref'))
+        ]
```
zannkukai, line 187: Does they already talks about small variable name ? In my mind, we choose to use at least 3 carac variable name (except with lambda).
```python
agents += [
    contrib['agent'] 
    for contrib in doc.get('contribution', [])
    if contrib.get('agent', {}).get('$ref'))
]
```
🚨🚨 bad closing parenthesis : `if c.get('agent', {}.get('$ref'))` --> `if c.get('agent', {}).get('$ref')`

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+        pids_to_replace
+    ):
+        """Updates the contribution and subjects in document.
+
+        :param doc_pid - string: document pid
+        :param source - string: agent source i.e. gnd, rero, mef
+        :param old_agent_pid - string: old agent pid from the database
+        :param new_agent_pid - string: new agent pid from the mef server
+        :param new_mef_pid - string: new MEF pid
+        :returns: true if the document has been updated
```
zannkukai, line 169: It never returns other value than `True`. Is this method really need to return something ?

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+    def get_contributions_pids(self, query='*', from_date=None):
+        """Get contributions identifiers.
+
+        :param query - string: a query to select the MEF record to be updated.
+        :param from_date - string: only the MEF updated on the MEF server after
+                                   the given date will be considered.
+        :returns: the list of the contribution identifiers.
+        :rtype: list of strings.
+        """
+        import logging
```
zannkukai, line 233: move to top of file (or I miss a tip/tricks)

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+
+                # can be mef pid, source pid or metadata
+                if self.agent_are_different(dict(agent), mef):
+                    # need a copy as we want to keep the MEF record
+                    # untouched for the next agent
+                    new_mef_data = deepcopy(mef)
+                    for field in [
+                        '$schema', '_created', '_updated'
+                    ]:
+                        new_mef_data.pop(field, None)
```
zannkukai, line 354: suggestion : maybe more readable - your solution contains 4 lines.

3 lines solutions
```python
fields_to_clean = ['$schema', '_created', '_update']
for field in fields_to_clean:
    new_mef_data.pop(field, None)
 ```
 or
 ```python
new_mef_data.pop('$schema', None)
new_mef_data.pop('_created', None)
new_mef_data.pop('_updated', None)
 ...
 ```

File: rero_ils/modules/contributions/sync.py [hunk 9]
```diff
@@ -0,0 +1,546 @@
...
+
+                    if (old_mef_pid != new_mef_pid):
```
zannkukai, line 356: remove unnecessary parentheses

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+                            doc_updated.update(new_doc_updated)
+                            updated = updated or new_updated
+                            error = error or new_error
+                    # something changed
+                    # update the content
+                    self.logger.info(
+                        f'MEF(pid: {agent.pid}) content has been '
+                        f'updated')
+                    if not self.dry_run:
+                        if (old_mef_pid == new_mef_pid):
```
zannkukai, line 391: unnecessary parentheses

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+                self.logger.info(
+                    f'MEF(pid: {agent.pid}) try to update '
+                    f'documents: {doc_pids}')
+                for doc_pid in doc_pids:
+                    doc_is_updated = self.update_agents_in_document(
+                        doc_pid=doc_pid,
+                        pids_to_replace=pids_to_replace
+                    )
+                    if doc_is_updated:
+                        doc_updated.add(doc_pid)
```
zannkukai, line 415: `self.update_agents_in_document` can only return `True` (maybe it could raise an exception) so :
```python
for pid in docs_pids:
    self.update_agents_in_document(doc_pid=pid, pids_to_replace=pids_to_replace)
....
return doc_pids, updated, error
```

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+        except Exception as e:
+            self.logger.error(f'ERROR: MEF(pid:{pid}) -> {e}')
+            error = True
+            # uncomment to debug
+            # raise
+        return doc_updated, updated, error
+
+    def start_sync(self):
```
zannkukai, line 423: should be a private method --> `_start_sync(self)`

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+        """Add logging information about the starting process."""
+        self.start_timestamp = datetime.now()
+        if self.dry_run:
+            self.logger.info(
+                '--------- Starting synchronization in dry run mode ---------')
+        else:
+            self.logger.info('--------- Starting synchronization ---------')
+
+    def end_sync(self, n_doc_updated, n_mef_updated, mef_errors):
```
zannkukai, line 432: private method too

rero_ils/modules/contributions/sync.py
```diff
@@ -0,0 +1,546 @@
...
+
+    @classmethod
+    def clear_errors(cls):
+        """Removes errors in the cache information."""
+        data = get_timestamp('sync_agents')
+        if data.get('errors'):
+            data['errors'] = []
+            set_timestamp('sync_agents', **data)
+
+    def start_clean(self):
```
zannkukai, line 517: private too

zannkukai (changes requested):

rero_ils/dojson/utils.py
```diff
@@ -422,8 +422,9 @@ def get_contribution_link(bibid, reroid, id, key):
     # In dojson we dont have app. mef_url should be the same as
     # RERO_ILS_MEF_AGENTS_URL in config.py
     # https://mef.test.rero.ch/api/agents/mef/?q=rero.rero_pid:A012327677
-    mef_host = os.environ.get('RERO_ILS_MEF_HOST', 'mef.rero.ch')
-    mef_url = f'https://{mef_host}/api/agents'
+    mef_url = os.environ.get(
+        'RERO_ILS_MEF_AGENTS_URL',
+        'https://mef.rero.ch/api/agents')
```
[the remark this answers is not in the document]
jma: Because we do not want to have multiple configuration to do the same thing. Here I would replace HOST by the full MEF url instead of hard coded '/api/agents'.

rero_ils/modules/contributions/sync.py
jma: 1. I do not think that it is a good candidate for `__cmp__` unless to create a new classe representing an agent which can be sometime, a Contribution object or a dict. And here it is a very specific comparison as we remove some fields such as `organisation.

Here self is not an agent by a Sync object processing this like an human that compare 2 apples.

rero_ils/modules/contributions/sync.py
jma: No as it is used in the cli.py file.

rero_ils/modules/contributions/sync.py
jma: No as it is used in the cli.py file.

rero_ils/modules/contributions/sync.py
jma: No as it is used in the cli.py file.

jma:
> Greats work ! I just have a general question. Could we use sync with long terms queries? Maybe, it's better to write method with pids list in argument. The idea is to launch multiple sync in the same times using the new generic task from rero-invenio-base 😁

Thanks. Unfortunately I do not think that it is safe to do it in parallel as two MEF updates can modify the same document. Perhaps it works for the db as we use transaction, but what about the indexing?

zannkukai (approved):

Issue #3044, opened 2022-08-10
Statistics: implement the wanted indicators, distributions, filters in the backend


pronguen:
As a system librarian, I want to download customizable statistics reports about my organisation.

## Acceptance criteria

- A json config per user defines which reports should be generated each month/year.
- Each report can contain (**(!)** are required):
    - **(!)** A category (circulation, catalogue, user management, administration)
        - **(!)** An indicator (see [column B](https://docs.google.com/spreadsheets/d/19t_qY9U9z4fLS_IzHw7ShaWUjWYIhdxcrgjtsQnwdO0/edit#gid=0))
            - **(!)** A type of indicator: fixed state (number of existing resources at a precise point in time) OR time period
    - *Zero, one or two* distributions for the indicator (for cross-referencing data according to different criteria) (see [column D](https://docs.google.com/spreadsheets/d/19t_qY9U9z4fLS_IzHw7ShaWUjWYIhdxcrgjtsQnwdO0/edit#gid=0))
    - **(!)** A frequency of generation
        - Monthly is generated at the end of every month / Yearly is generated at the end of every year
    - Time range (*only for indicators of type `time period`*): filters the data to only the chosen period. Options: all-time, last month, last year. 
    - A filter-query on the **corresponding index**

- The system admin can set the config (for a professional user with role `full_permissions`) that defines which reports are generated (not yet editable in the UI).

- All indicators  marked as `priority:Essential` with their mentionned distributions can be extracted : https://docs.google.com/spreadsheets/d/19t_qY9U9z4fLS_IzHw7ShaWUjWYIhdxcrgjtsQnwdO0/edit#gid=0

### UI

- A new permission role `pro_statistics` gives access to a view "Statistics" under "Statistics & Reports" that shows a list of generated reports for the user. `full_permissions` also has access.
- The user can download the CSV files for any generated report.

PR #3187, opened 2022-12-19
statistics: add resource for statistics report

vgranata:
Add resource for statistics report:

* create resource stats_cfg for statistics configuration report
* add class for statistics report to resource stats

Co-Authored-by: Valeria Granata <valeria@chaw.com>

https://github.com/rero/rero-ils/issues/3044


zannkukai (changes requested):
As a general remark, I think it's a pity to build a new resource only for stats configuration. 

I'm not closed about create a resource to manage configuration, but we need to think about a general resource that could be stored/manage several configuration models : not only for stats, but maybe for widget (that could be used into UI), for inventory report, ... 

I understand that we use AGILE methodology but it's not a reason to now try to find together a more global and flexible solution, using power of object modelisation inheritance/polymorphism principles.

PR #3193, opened 2022-12-21
user: add entrypoints api for the password

Garfield-fr:
* Adds entrypoint to generate a new password.
* Adds entrypoint to validate a password.
* Updates users api to use the new password generator and validator.
* Updates users fixtures to add the password field.

### Dependencies:

https://github.com/rero/rero-ils-ui/pull/915
https://github.com/rero/ng-core/pull/531

### DIFF
--- a/rero_ils/modules/users/api.py
+++ b/rero_ils/modules/users/api.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # RERO ILS
-# Copyright (C) 2021 RERO
+# Copyright (C) 2021-2023 RERO
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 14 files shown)


rero_ils/modules/utils.py
```diff
@@ -1192,3 +1194,44 @@ def draw_row_content(row, pad=' '):
...
+    password.append(random.choice(string.ascii_uppercase))
+    password.append(random.choice(string.digits))
+    if special_char:
+        password.append(random.choice(string.punctuation))
+
+    while len(password) < length:
+        password.append(random.choice(string.ascii_letters + string.digits))
+
+    random.shuffle(password)
+    return ''.join(password)
```
zannkukai, line 1237: I suggest this approach. Not "THE" approach but a way to ensure that the password is valid according the 'validate' function. (inspiration : https://geekflare.com/password-generator-python-code/)

```python
import secrets
import string

def password_generator(length=8, max_attemps=500):
    attemps = 0
    alphabet = string.ascii_letters + string.digits + string.punctuation
    while True: 
        pwd = ''
        for i in range(length):
            pwd += ''.join(secrets.choice(alphabet))
        print(f"[{attemps}] {pwd}")
        try:
            password_validate(pwd)
            return pwd
        except Exception:
            pass
        attemps += 1
        if attemps > max_attemps:
            raise Exception('max_attemps occurs')
```

Garfield-fr: We have defined that a minimum of one lower case character, one upper case character and one number is required. With this approach to random, it may take a while to get a good string and maybe get an error.

zannkukai (approved):

rero_ils/modules/users/api.py
```diff
@@ -86,10 +108,12 @@ def create(cls, data, **kwargs):
...
             data.pop('roles', None)
             cls._validate(data=data)
-            password = data.pop(
-                'password',
-                data.get('birth_date', '123456')
-            )
+            # Generate password if not present and validate password
+            password = data.pop('password', password_generator())
+            try:
+                password_validator(password)
```
jma, line 114: I would put this in the `_validate` method of the user api.

jma (approved):
Great job.

PR #3198, opened 2023-01-03
dependencies: fix vulnerabilites

Garfield-fr:
When the test is run during the Christmas period, the fees_after_extend
test is in error, because the defined interval is not correct.

* Adds packages exception for safety.
* Fixes cryptography on the latest version 38.
* Fixes pydocstyle on the latest version 6.1.
* Fixes the test_fees_after_extend.

### DIFF
--- a/tests/ui/circulation/test_actions_extend.py
+++ b/tests/ui/circulation/test_actions_extend.py
@@ -72,7 +72,8 @@ def test_fees_after_extend(
         checkout_cipo, dbcommit=True, reindex=True)
 
     # UPDATE LOAN TO BE OVERDUE
-    interval = 10
+    # LIBRARY FIXTURES EXCEPTION: Christmas Holidays is 15 days
+    interval = 15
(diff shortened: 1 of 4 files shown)


PascalRepond (approved):

tests/ui/circulation/test_actions_extend.py
```diff
@@ -72,7 +72,8 @@ def test_fees_after_extend(
         checkout_cipo, dbcommit=True, reindex=True)
 
     # UPDATE LOAN TO BE OVERDUE
-    interval = 10
+    # LIBRARY FIXTURES EXCEPTION: Christmas Holidays is 15 days
+    interval = 15
```
PascalRepond, line 76: Does the test also still work if running outside of the exception interval (other date)?

Garfield-fr: Yes I did a test with the date of 2022-06-15 and everything is OK.

PR #3204, opened 2023-01-10
user: improve register and forgot password

Garfield-fr:
* Adds beautifulsoup4 package for test.
* Disables account rest api endpoint.
* Adds validations on the change password api.
* Updates password on the interface test information.
* Adds the possibility to put a message in the cache
  and display it on the html page.
* Changes of the API response on invalid login.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -410,6 +410,10 @@ vendors = "rero_ils.modules.vendors.mappings"
 operation_logs = "rero_ils.modules.operation_logs.es_templates:list_es_templates"
 rero_ils = "rero_ils.es_templates:list_es_templates"
 
+
+[tool.poetry.group.dev.dependencies]
+beautifulsoup4 = "^4.11.1"
+
 [tool.poe.tasks]
 bootstrap = {cmd = "./scripts/bootstrap", help = "Runs bootstrap"}
 console = {cmd = "./scripts/console", help = "Opens invenio shell"}
--- a/rero_ils/accounts_views.py
+++ b/rero_ils/accounts_views.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # RERO ILS
-# Copyright (C) 2019 RERO
+# Copyright (C) 2019-2023 RERO
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
@@ -17,20 +17,25 @@
 
 """Invenio Account custom views."""
 
-from flask import after_this_request, current_app
+from flask import abort, after_this_request, current_app
 from flask import request as flask_request
+from flask.views import MethodView
 from flask_babelex import gettext as _
+from flask_security.confirmable import requires_confirmation
+from flask_security.utils import get_message, verify_and_update_password
 from invenio_accounts.utils import change_user_password
 from invenio_accounts.views.rest import \
     ChangePasswordView as BaseChangePasswordView
 from invenio_accounts.views.rest import LoginView as CoreLoginView
 from invenio_accounts.views.rest import _abort, _commit, use_args, use_kwargs
-from marshmallow import Schema, fields
+from marshmallow import Schema, fields, validates, validates_schema
 from webargs import ValidationError, validate
 from werkzeug.local import LocalProxy
 
-from .modules.patrons.api import Patron, current_librarian
-from .modules.users.api import User
+from rero_ils.modules.patrons.api import Patron, current_librarian
+from rero_ils.modules.users.api import User
+from rero_ils.modules.utils import PasswordValidatorException, \
+    password_validator
 
 current_datastore = LocalProxy(
     lambda: current_app.extensions['security'].datastore)
@@ -46,6 +51,22 @@ def user_exists(email):
         raise ValidationError(_('INVALID_USER_OR_PASSWORD'))
 
 
+def validate_password(password):
+    """Validate the password."""
+    length = current_app.config.get('RERO_ILS_PASSWORD_MIN_LENGTH', 8)
+    special_char = current_app.config.get('RERO_ILS_PASSWORD_SPECIAL_CHAR')
+    try:
+        password_validator(password, length=length, special_char=special_char)
+    except PasswordValidatorException as pve:
+        raise ValidationError(str(pve)) from pve
+
+
+def validate_passwords(password, confirm_password):
+    """Validate that the 2 passwords are identical."""
+    if password != confirm_password:
+        raise ValidationError(_('The 2 passwords are not identical.'))
+
+
 class LoginView(CoreLoginView):
     """invenio-accounts Login REST View."""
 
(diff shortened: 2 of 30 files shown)


pyproject.toml
```diff
@@ -410,6 +410,10 @@ vendors = "rero_ils.modules.vendors.mappings"
 operation_logs = "rero_ils.modules.operation_logs.es_templates:list_es_templates"
 rero_ils = "rero_ils.es_templates:list_es_templates"
 
+
+[tool.poetry.group.dev.dependencies]
+beautifulsoup4 = "^4.11.1"
```
zannkukai, line 415: Youhou è  ! a new dependencies 🥳

rero_ils/accounts_views.py
```diff
@@ -46,6 +51,35 @@ def user_exists(email):
...
+                         ' case letter, one lower case letter'
+                         ' and one number.'),
+            'invalid_special_char': _('The password must contain at least'
+                                      ' one upper case letter, one lower'
+                                      ' case letter, one number and one'
+                                      ' special character.')
+        }
+        message = messages['invalid_special_char'] if special_char \
+            else messages['invalid']
+        raise ValidationError(message)
```
zannkukai, line 74: Didn't you already create a validation password method ? Why don't use it .... I think I miss something :s

rero_ils/accounts_views.py
```diff
@@ -71,23 +104,51 @@ def post(self, **kwargs):
         self.login_user(user)
         return self.success_response(user)
 
+    def verify_login(self, user, password=None, **kwargs):
+        """Verify the login via password."""
+        if not user.password or not verify_and_update_password(password, user):
+            _abort(_('INVALID_USER_OR_PASSWORD'))
```
zannkukai, line 110: You mark the string `INVALID_USER_OR_PASSWORD` for translation ? Why use a snake case string ?
maybe it should be `get_message('INVALID_USER_OR_PASSWORD')[0]` ?

rero_ils/accounts_views.py
```diff
@@ -139,3 +199,15 @@ def post(self, args):
...
+class DisabledAuthApiView(MethodView):
+    """Disabled Account rest auth api."""
+
+    def post(self, **kwargs):
+        """Post Data."""
+        abort(404)
+
+    def get(self, **kwargs):
+        """Get Data."""
+        abort(404)
```
zannkukai, line 213: Suggestion : You could write 
```python
    def post(self, **kwargs):
        """Post Data."""
        abort(404)

    get = post
```
in this case, the GET and POST method are always the same.

rero_ils/modules/patrons/api.py
```diff
@@ -236,7 +236,8 @@ def load(cls, data):
     def removeUserData(cls, data):
         """Remove the user data."""
         data = deepcopy(data)
-        profile_fields = User.profile_fields + ['username', 'email']
+        profile_fields = User.profile_fields + \
+            ['username', 'email', 'password']
```
zannkukai, line 240: Oh yes. A future rebase conflict 🤪

rero_ils/modules/users/validators.py
```diff
@@ -0,0 +1,50 @@
...
+
+"""Password validator."""
+
+from wtforms import ValidationError
+
+from rero_ils.modules.utils import PasswordValidatorException, \
+    password_validator
+
+
+class PasswordValidator(object):
```
zannkukai, line 26: With Python3, it's recommended to not inherit from `object` --> just write `class PasswordValidator:`

rero_ils/modules/users/validators.py
```diff
@@ -0,0 +1,50 @@
...
+    """Password Validator."""
+
+    def __init__(self, length=8, special_char=False):
+        """Password validator class initializer."""
+        self.length = length
+        self.special_char = special_char
+
+    def __call__(self, form, field):
```
zannkukai, line 34: missing docstring about params and possible `raise` statements

rero_ils/modules/users/validators.py
```diff
@@ -0,0 +1,50 @@
...
+        """Call."""
+        field_length = field.data and len(field.data) or 0
+        messages = {
+            'length': field.ngettext(
+                'Field must be at least %(min)d character long.',
+                'Field must be at least %(min)d characters long.',
+                self.length)
+        }
```
zannkukai, line 42: Seems a little bit overkill to declare a dict with only 1 key. You could simply use `field.ngettext` into the belowed `if` statement --> 
```python
raise ValidationError(field.ngettext(
      'Field must be at least %(min)d character long.',
      'Field must be at least %(min)d characters long.',
      self.length
))
```
Additionnaly, it should be translated. Isn't it ?

rero_ils/theme/templates/rero_ils/_info_test.html
```diff
@@ -20,31 +20,31 @@
   <h3>Test RERO ILS</h3>
   <p>As an anonymous user</p>
   <p>As a patron, login with:</br>
-  reroilstest+giulia@gmail.com / 123456</p>
+  reroilstest+giulia@gmail.com / Pw123456</p>
```
zannkukai, line 23: I love security 🤣

rero_ils/accounts_views.py
```diff
@@ -46,6 +53,25 @@ def user_exists(email):
         raise ValidationError(_('INVALID_USER_OR_PASSWORD'))
 
 
+def validate_password(password):
+    """Validate the password."""
+    length = current_app.config.get('RERO_ILS_PASSWORD_MIN_LENGTH', 8)
+    special_char = current_app.config.get('RERO_ILS_PASSWORD_SPECIAL_CHAR')
+    if len(password) < length:
+        raise ValidationError(
+            _('Field must be at least %(min)d characters long.', min=length))
```
zannkukai, line 62: As password length check is already include into `password_validator`, why this test here ?

rero_ils/modules/users/validators.py
```diff
@@ -0,0 +1,52 @@
...
+        self.special_char = special_char
+
+    def __call__(self, form, field):
+        """Call.
+
+        :param form: the current form.
+        :param field: the password field.
+        :raise ValidationError: If the password isn't valid.
+        """
+        field_length = field.data and len(field.data) or 0
```
zannkukai, line 41: Seems strange... `field.data` should return boolean. So your test `field.data and len(field.data)` will compare a boolean and an integer ; this integer will be casted to boolean. So `field_length` should return `True|False` (impossible to return 0).
Next line, you compare `field_length < self.length`, so the boolean value is now casted to int (0|1).

A solution should be to use the ugly python ternary operator : `field_length = len(field.data) if field.data else 0`

rero_ils/modules/users/validators.py
```diff
@@ -0,0 +1,52 @@
...
+        if field_length < self.length:
+            raise ValidationError(field.ngettext(
+                'Field must be at least %(min)d character long.',
+                'Field must be at least %(min)d characters long.',
+                self.length
+            ) % {"min": self.length})
```
zannkukai, line 47: same remark. The length validation is already included into `password_validator`

rerowep (approved):

zannkukai (approved):

PR #3206, opened 2023-01-11
fix: reinforce search/export of patron transaction 

zannkukai:
If patron transaction related resources are delete, the search and export caused error. This PR fix all these problems.

### DIFF
--- a/rero_ils/modules/patron_transaction_events/serializers/csv.py
+++ b/rero_ils/modules/patron_transaction_events/serializers/csv.py
@@ -18,6 +18,7 @@
 
 """Patron transaction event CSV serializer."""
 import csv
+from functools import wraps
 
 from babel.numbers import format_decimal
 from ciso8601 import parse_datetime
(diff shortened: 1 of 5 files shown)


rero_ils/modules/patron_transaction_events/serializers/csv.py
```diff
@@ -114,6 +115,15 @@ def generate_csv():
         return stream_with_context(generate_csv())
 
 
+def _skip_if_no_record(func):
```
jma, line 118: A method starting with '_' is considered as private. What does it means for a function?

jma: This function can be generalized and added as a generic decorator.

lauren-d (approved):

jma (approved):
Please change your commit message as defined in the guide: https://github.com/rero/developer-resources/blob/e8be0aab1cdc1d2a5b50782807e38dda72a54ffd/git-github/git.md#subject

Garfield-fr (approved):

benerken (approved):

rero_ils/modules/patron_transaction_events/serializers/csv.py
zannkukai: @jma due to the default return statement, I think this decorator isn't so generic and can't be generalized.

PR #3219, opened 2023-01-23
imports: LOC

rerowep:
* Corrects transformation errors for bad searches for LOC.

### DIFF
--- a/rero_ils/modules/imports/api.py
+++ b/rero_ils/modules/imports/api.py
@@ -18,6 +18,7 @@
 
 """Import from extern resources."""
 
+
 from __future__ import absolute_import, print_function
 
 import pickle
@@ -350,6 +351,7 @@ class Import(object):
                                 tag='{http://www.loc.gov/zing/srw/}'
                                     'record'):
                             yield element
+
                     xml_records = _split_stream(BytesIO(response.content))
 
                     for xml_record in xml_records:
@@ -363,7 +365,7 @@ class Import(object):
                         record = self.to_json_processor(json_data)
 
                         id = self.get_id(json_data)
-                        if record:
+                        if record and id:
                             data = {
                                 'id': id,
                                 'links': {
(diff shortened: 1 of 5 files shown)


rero_ils/modules/imports/api.py
```diff
@@ -375,11 +378,13 @@ def _split_stream(stream):
...
                             self.data.append(json_data)
                             self.results['hits']['hits'].append(data)
-                    self.results['hits']['remote_total'] = int(etree.parse(
-                            BytesIO(response.content))
-                            .find('{*}numberOfRecords').text)
+                    self.results['hits']['remote_total'] = 0
+                    with contextlib.suppress(Exception):
+                        self.results['hits']['remote_total'] = int(etree.parse(
+                                BytesIO(response.content))
+                                .find('{*}numberOfRecords').text)
```
zannkukai, line 385: I think it's not the pythonic way to use. `contextlib.suppress(Exception)` should be used only as last option.
In this case, a simple `try...except` seems better :
```python
try:
    self.results['hits']['remote_total'] = int(etree.parse(
          BytesIO(response.content))
          .find('{*}numberOfRecords').text)
except Exception:
   self.results['hits']['remote_total'] = 0
```

zannkukai (changes requested):

zannkukai (approved):

Garfield-fr (approved):

jma (changes requested):
Can you provide some tests? I would be great to have a more explicit commit message. Thanks.

rero_ils/modules/imports/api.py
```diff
@@ -484,7 +487,10 @@ def get_id(self, json_data):
         :param id: json document
         :return: id of the record
         """
-        return json_data.get('010__').get('a').strip()
+        id = None
+        if json_data.get('010__'):
+            id = json_data.get('010__').get('a').strip()
+        return id
```
lauren-d, line 493: 1. In python, the default return value for a function is `None`
(see: https://realpython.com/python-return-statement/#:~:text=If%20you%20don%E2%80%99t%20supply%20an%20explicit%20return%20statement%20with%20an%20explicit%20return%20value%2C%20then%20Python%20will%20supply%20an%20implicit%20return%20statement%20using%20None%20as%20a%20return%20value)
so, you can just do :
```python
if json_data.get('010__'):
  return json_data.get('010__').get('a').strip()
```
2. Are you sure that tag `010` always have a subfield `a` ?

lauren-d (changes requested):

rero_ils/modules/imports/api.py
rerowep: I had already forgotten about it, thank you

lauren-d (approved):

jma (approved):

PR #3239, opened 2023-02-01
indexing: fix wildcard queries

jma:
* Fixes wildcard queries for field containing boolean sub fields. For example: `q=subject.%5C*:test`

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

### DIFF
--- a/rero_ils/query.py
+++ b/rero_ils/query.py
@@ -377,6 +377,9 @@ def search_factory(self, search, query_parser=None):
     def _default_parser(qstr=None, query_boosting=None):
         """Default parser that uses the Q() from elasticsearch_dsl."""
         query_type = 'query_string'
+        # avoid elasticsearch errors when it can't convert a boolean or
+        # numerical values during the query
+        lenient = True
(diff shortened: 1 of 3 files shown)


rerowep (approved):

rero_ils/query.py
```diff
@@ -387,14 +390,14 @@ def _default_parser(qstr=None, query_boosting=None):
             qstr = _PUNCTUATION_REGEX.sub(' ', qstr)
             qstr = re.sub(r'\s+', ' ', qstr).rstrip()
             if not query_boosting:
-                return Q(query_type, query=qstr,
+                return Q(query_type, lenient=lenient, query=qstr,
                          default_operator=default_operator)
             else:
                 return Q('bool', should=[
```
zannkukai, line 396: Note (to be tested) : instead of `Q('bool', should=[Q(), Q(), ...]` we could use simpler notation `return Q() | Q()`

zannkukai (approved):

Issue #3220, opened 2023-01-23
documents: invalid identifiers should be indexed


PascalRepond:
### How it works
Currently, invalid identifiers are not indexed because they are normalized first 

https://github.com/rero/rero-ils/blob/168f6f11377d3d0d532d66aeee4109bae72e45f5/rero_ils/modules/commons/identifiers.py#L168-L179

### Improvement suggestion
As a user, I want to be able to search a document by its identifier even if this identifier is invalid.
- Correct the dump function so that invalid identifiers can be indexed.
- Add a unit test.
- Add an alembic script to migrate existing data that is not correctly indexed.

PR #3240, opened 2023-02-01
document: fix EAN identifier indexing

zannkukai:
Invalid EAN identifiers were not dumped into ES during indexing process. Now, even invalid EAN are indexed and could be searched. 
Closes rero/rero-ils#3220.

### migration ###

execute the script :
```python
from rero_ils.modules.documents.api import DocumentsSearch, DocumentsIndexer
from elasticsearch_dsl.query import Q

query = DocumentsSearch().filter('nested',
    path='nested_identifiers',
    query=Q('bool', must=[
        Q('term', nested_identifiers__type='bf:Ean'),
        ~Q('exists', field='nested_identifiers.value')
    ])).source(False)
uuids = [hit.meta.id for hit in query.scan()]

indexer = DocumentsIndexer()
indexer.bulk_index(uuids)
_, (indexer_count, error_count) = indexer.process_bulk_queue()
print(f'{indexer_count} documents indexed, {error_count} errors')
```

### DIFF
--- a/rero_ils/modules/commons/identifiers.py
+++ b/rero_ils/modules/commons/identifiers.py
@@ -241,11 +241,15 @@ class EANIdentifier(Identifier):
 
     def normalize(self) -> str:
         """Get the normalized value for this EAN."""
-        return canonical(self.value)
+        return canonical(self.value) or self.value
     __str__ = normalize
 
+    def is_valid(self) -> bool:
+        """Check if the identifier is valid."""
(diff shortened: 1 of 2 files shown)


rerowep (approved):

PascalRepond (approved):
What about migration of current data? Document reindexing is enough? If yes, precise it in the commit message and PR description please.

PR #3247, opened 2023-02-07
Merge `US-Permissions` to `staging`

zannkukai:
And then... crying for rebase


rero_ils/alembic/5f0b086e4b82_patron_role_migration.py
```diff
@@ -0,0 +1,112 @@
...
+    patron_uuids = []
+
+    for hit in query.scan():
+        patron = Patron.get_record(hit.meta.id)
+        original_roles = patron.get('roles')
+        migrated_roles = get_new_roles(original_roles)
+        LOGGER.info(f'* Updating ptrn#{patron.pid} [{patron.formatted_name}]')
+        LOGGER.info(f'\t - Original roles are : [{original_roles}]')
+        LOGGER.info(f'\t - New roles are : [{migrated_roles}]')
+        patron['roles'] = migrated_roles
```
PascalRepond, line 63: What about a user with original roles [`patron`, `librarian`]? They would lose their patron role with this script. We should replace existing roles in the user and not rewrite.

PascalRepond (changes requested):

rero_ils/alembic/5f0b086e4b82_patron_role_migration.py
zannkukai: It should not happens because (from my memories) we won't have a `Patron` resources with `patron` and `librarian` (it causes some problems for authentication ?) In such case, this `User` must have two related `Patron` resources : One with `patron` role, other with `staff roles`

PascalRepond (approved):

benerken (approved):

lauren-d (approved):

PR #3248, opened 2023-02-07
utils: timestamps

rerowep:
* Adds `name` to timestamps for better json parsing in monitoring software.
* Saves also 0 values for `notification-creation` timestamp.

### DIFF
--- a/rero_ils/modules/notifications/tasks.py
+++ b/rero_ils/modules/notifications/tasks.py
@@ -126,8 +126,10 @@ def create_notifications(types=None, tstamp=None, verbose=True):
                     )
         process_notifications(NotificationType.OVERDUE)
     notification_sum = sum(notification_counter.values())
-    counters = {k: v for k, v in notification_counter.items() if v > 0}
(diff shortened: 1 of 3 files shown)


rero_ils/modules/notifications/tasks.py
```diff
@@ -126,6 +126,8 @@ def create_notifications(types=None, tstamp=None, verbose=True):
                     )
         process_notifications(NotificationType.OVERDUE)
     notification_sum = sum(notification_counter.values())
+    # in timestamp we need also 0 values
+    set_timestamp('notification-creation', **counters)
     counters = {k: v for k, v in notification_counter.items() if v > 0}
```
zannkukai, line 131: You tried to use `**counters` before it is initialized.
Both line should be switched

zannkukai (changes requested):

zannkukai (approved):

jma (approved):
Do we need to do a migration?

PR #3261, opened 2023-02-22
contribution: rename `Contribution` to `Entity`

zannkukai:
Renames the `Contribution` resource to more generic `Entity`. This change will cause a lot of import changes in many files of the project.


rero_ils/modules/entities/views.py
```diff
@@ -157,66 +182,29 @@ def translat_unified(data, prefix=''):
...
+    return {
+        translate(f'{prefix}{key}'): value
+        for key, value
+        in data.items()
+    }
 
 
 @blueprint.app_template_filter()
 @api_blueprint.app_template_filter()
 def translat(data, prefix='', seperator=', '):
```
jma, line 194: translat => translate

tests/api/documents/test_documents_dumpers.py
```diff
@@ -0,0 +1,58 @@
...
+    }]
+    with pytest.raises(RecordNotFound):
+        document.dumps(dumper=document_replace_refs_dumper)
+
+
+@pytest.mark.skip(reason="Dumper() not implement 'load()' method")
+def test_multi_dumpers(document_data):
+    """Test MultiDumper."""
+    # TODO :: Try to fix this test.
+    document_title_dumper.load(document_data, Document)
```
jma, line 58: As it is not used. Do you prefer to remove the load function from the MultiDumper?

jma (approved):

tests/api/documents/test_documents_dumpers.py
zannkukai: Nope. The test is now created and could be updated if we play further with `load` ;-)

PR #3263, opened 2023-02-28
tests: Prevent MEF access

rerowep:


Garfield-fr (approved):

jma (changes requested):
Please change you commit message to respect https://github.com/rero/developer-resources/blob/master/git-github/git.md#git-commit-guidelines.

zannkukai (approved):
Change will be OK, Is it to avoid unit test problem ? Mock ?

jma (approved):

PR #3264, opened 2023-03-01
statistics: add permissions on stats_cfg

Garfield-fr:
* Adds create, update and delete permissions on stats_cfg resource.


lauren-d (changes requested):
I think we need a script migration to add new roles in database

PascalRepond:
> I think we need a script migration to add new roles in database

It depends what we want by default. For our users, I think it is OK if only system librarians (= full_permissions) have access by default. They can attribute the permissions on a case by case basis to other users.

Garfield-fr:
> > I think we need a script migration to add new roles in database
> 
> It depends what we want by default. For our users, I think it is OK if only system librarians (= full_permissions) have access by default. They can attribute the `statistics` permissions on a case by case basis to other users.

The access to the configuration is provided only for a system libration. It is not possible for a simple librarian.

lauren-d (approved):

zannkukai (commented):
As @lauren-d says, we need an alambic script to allow 'pro_full_permission' to get new permissions. Without that, no policy will attach to any role.

zannkukai (approved):

PR #3285, opened 2023-03-22
circulation: fix `can_be_requested` circulation method

zannkukai:
Fixes the behavior of `can_be_requested` method use by Invenio circulation to deal with `item.temporary_location`. Without this fix, it's always the item.holding_location_pid (owning location) that will be used.


PascalRepond (approved):

lauren-d (approved):

benerken (approved):

jma (approved):
please change you commit message to be compliant such as no '.' and no uppercase.

PR #3298, opened 2023-04-03
ILL request: check if item exists in organisation

### LINKED ISSUE #3138: Global ILL requests: check if an item exists in organisation before submitting
### How it works
Currently, patrons can request any document from the global view by ILL. Often, they wrongly request documents that are present in the organisation, so they shouldn't be ILL requests.

### Improvement suggestion
Before submitting an ILL request, the app checks 

- If the `document_link` leads to a document in the current instance (with regex)
- If the requested document exists in the same organisation as the selected `pickup_location`

If yes, a modal appears : "The requested document exists in your current library network (link to document with local viewcode), are you sure you want to submit an inter-library loan request?" "Yes" "No"

This prevents patrons from submitting ILL requests for documents that should be standard requests.

Garfield-fr:
The system checks if a copy exists in case the url matches the current catalog. The user has the possibility to continue even if the document exists.

* Closes #3138.

### DIFF
--- a/rero_ils/modules/ill_requests/forms.py
+++ b/rero_ils/modules/ill_requests/forms.py
@@ -126,7 +126,7 @@ class ILLRequestSourceForm(FlaskForm):
         if self.origin.data:
             self.url.validators = [
                 validators.DataRequired(),
-                validators.URL()
+                validators.URL(require_tld=False)
             ]
         return super().validate()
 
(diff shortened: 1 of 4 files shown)


zannkukai (approved):
my jQuery is very rusted, but seems OK

ManaDeweerdt:
- [ ] nothing happens when we try to add an ILL request. https://ilsdev.test.rero.ch/global/ill_requests/new
image

benerken (approved):

lauren-d (approved):

Garfield-fr:
> * REP/11.04: Cannot send ILL request  
>   * sometimes, both warning modals keep alternating
>   * sometimes, there is an error
> * if the ill_request doesn't have a URL, validating the modal doesn't do anything

I can't reproduce it. I tried several urls without any problem.
Can you provide me with tested examples?

PascalRepond (approved):

PR #3315, opened 2023-04-18
document: linked concept for genreForm field.

zannkukai:
* Adds configuration to search on remote MEF server for specific
  'genre/forme Rameau' topics.
* Allows linked `genreForm` document field to be linked to a remote
  MEF topic.


Ewa666:
In this record : https://ilsdev.test.rero.ch/professional/records/documents/detail/2000732 I can't see MEF genreForm in public view (MEF genreForm is "Guides touristiques et de visite").

lauren-d (approved):

rero_ils/modules/documents/utils.py
```diff
@@ -283,33 +283,48 @@ def create_authorized_access_point(agent):
...
+
+    :param entity: the entity to transform.
+    """
+    if entity.get('pid'):
+        # in such case, it means that's an entity linked to an `Entity` record.
+        # and we don't need to transform it. Just return the current entity
+        # without any modifications.
+        return entity
+
+    if access_point := entity.pop('authorized_access_point', None):
```
jma, line 320: Maybe it can be a good idea to make a copy.

rero_ils/modules/documents/views.py
```diff
@@ -252,42 +252,64 @@ def can_request(item):
...
+    :param language: current language on interface.
+    :param part_separator: the string use to glue parts of the entity label.
+    :returns: the best possible label to display.
+    """
+    parts = []
+    if 'pid' in entity:
+        entity = Entity.get_record_by_pid(entity['pid'])
+        parts.append(entity.get_authorized_access_point(language=language))
+    else:
+        default_key = 'authorized_access_point'
```
jma, line 302: Seems overkill to create a variable.

rero_ils/query.py
```diff
@@ -38,20 +38,22 @@
 _PUNCTUATION_REGEX = re.compile(r'[:,\?,\,,\.,;,!,=,-]+(\s+|$)')
 
 
-def and_term_filter(field, must=[], must_not=[]):
```
jma, line 41: Usually I prefer explicit instead of implicit. But it is up to you.

rero_ils/query.py
```diff
@@ -68,22 +70,27 @@ def inner(values):
...
-        language = request.args.get("lang", current_i18n.language)
+        language = request.args.get('lang', current_i18n.language)
         if not language or language not in get_i18n_supported_languages():
             language = current_app.config.get('BABEL_DEFAULT_LANGUAGE', 'en')
-        i18n_field = '{field}_{language}'.format(
-            field=field,
-            language=language
-        )
+        i18n_field = f'{field}_{language}'
         must = [Q('term', **{i18n_field: value}) for value in values]
```
jma, line 86: We can probably use a term set filters/query: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-set-query.html

jma (approved):
Great job, thanks.

rero_ils/config.py
```diff
@@ -1615,111 +1615,78 @@ def _(x):
...
-                terms=dict(field='intendedAudience.value',
-                           size=DOCUMENTS_AGGREGATION_SIZE)
-            ),
-            year=dict(date_histogram=dict(
-                field='provisionActivity.startDate',
-                interval='year',
-                format='yyyy')
-            )
+            status=dict(terms=dict(field='holdings.items.status', size=DOCUMENTS_AGGREGATION_SIZE)),
+            genreForm=dict(terms=dict(field='genre_form_authorized_access_point', size=DOCUMENTS_AGGREGATION_SIZE)),
```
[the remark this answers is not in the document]
zannkukai: Thx for review ! it was completely wrong ! I need to use "facet_genre_form_XX" ES field with new i18n filter

rero_ils/modules/documents/dumpers/indexer.py
```diff
@@ -164,28 +165,45 @@ def _process_identifiers(record, data):
...
-            data.pop('contribution', None)
-            data['contribution'] = contributions
-
-        # TODO: compare record with those in DB to check which authors have
-        # to be deleted from index
+    @staticmethod
+    def _process_local_field(record, data):
+        """Add local field data related to this document."""
+        from rero_ils.modules.local_fields.api import LocalField
+        fields = list(LocalField.get_local_fields_by_id('doc', record['pid']))
```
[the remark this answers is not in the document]
zannkukai: because if generator return empty dataset, it will cause an empty array into ES document ... not really incorrect, but in the project, for such case, we don't sent any data instead of empty array (there is a uni-test about that)

rero_ils/modules/documents/dumpers/replace_refs.py
```diff
@@ -66,8 +66,16 @@ def dump(self, record, data):
         return data
 
 
-class ReplaceRefsSubjectsDumper(ReplaceRefsEntitiesDumperMixin):
-    """Replace linked subjects in document."""
+class ReplaceRefsEntitiesDumper(ReplaceRefsEntitiesDumperMixin):
+    """Replace linked entities in document."""
+
+    def __init__(self, *args):
```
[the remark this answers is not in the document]
zannkukai: with `*args` the call is `ReplaceRefsEntitiesDumper('fieldA', 'fieldB', ...)` (Pythonic way)

with `field_names` the call is  `ReplaceRefsEntitiesDumper(['fieldA', 'fieldB', ...])` and we need to check into `__init__` if `field_names` is None to assign a default value.

rero_ils/modules/documents/dumpers/replace_refs.py
```diff
@@ -76,13 +84,14 @@ def dump(self, record, data):
...
-            dict(entity=self._replace_entity(subject) if subject.get('$ref')
-                 else subject)
-            for subject in [d['entity'] for d in data.get('subjects', [])]
-        ]
-        if entities:
-            data['subjects'] = entities
+        for field_name in self.field_names:
+            entities = []
+            for entity in [d['entity'] for d in data.get(field_name, [])]:
+                if entity.get('$ref'):
```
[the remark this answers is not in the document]
zannkukai: not really : I need to proceed each 'entity' (local/mef_related), but for "mef_related" entity, i need to dump it.

rero_ils/modules/documents/utils.py
zannkukai: We don't need to keep the original "authorized_access_point" field in ES because for i18n field, we search exclusively on "authorized_access_point_xx" field. If we want i18n fields, we need to search on the correct key with expert search.

rero_ils/query.py
zannkukai: we can't use empty array as function default param because this is "Immutable" (it's you that explain me this fact).
So we need to use `None` as default param signature and write a `must = must or []` into the function body to solve this problem.

PR #3317, opened 2023-04-19
notifications: correct process counts

rerowep:
* Corrects notification processed counts.
* Deletes double time stamps for notifications.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

### DIFF
--- a/rero_ils/modules/notifications/tasks.py
+++ b/rero_ils/modules/notifications/tasks.py
@@ -72,8 +72,10 @@ def create_notifications(types=None, tstamp=None, verbose=True):
         for loan in get_due_soon_loans(tstamp=tstamp):
             try:
                 logger.debug(f"* Loan#{loan.pid} is considered as 'due_soon'")
-                loan.create_notification(_type=NotificationType.DUE_SOON)
-                notification_counter[NotificationType.DUE_SOON] += 1
+                notifications = loan.create_notification(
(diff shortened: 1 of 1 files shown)


Garfield-fr (approved):

rero_ils/modules/notifications/tasks.py
```diff
@@ -72,8 +72,10 @@ def create_notifications(types=None, tstamp=None, verbose=True):
         for loan in get_due_soon_loans(tstamp=tstamp):
             try:
                 logger.debug(f"* Loan#{loan.pid} is considered as 'due_soon'")
-                loan.create_notification(_type=NotificationType.DUE_SOON)
-                notification_counter[NotificationType.DUE_SOON] += 1
+                notifications = loan.create_notification(
+                    _type=NotificationType.DUE_SOON)
+                notification_counter[NotificationType.DUE_SOON] += len(
+                    notifications)
```
zannkukai, line 78: Technically it seems correct.
But logically, how could be send more than 1 `due_soon` notification at day for a specific loan ?
It could happen if the task generating the notifications isn't working for few days ; in such case, the `loan.create_notification()` method could generate the "due_soon_for_day_-5", and "due_soon_for_day-2". But it means that the end user will receive multiple notifications at same time...

zannkukai (approved):

PR #3325, opened 2023-05-04
locations: add RERO serializers

zannkukai:
* Adds `rero+json` serializer on `Location` resource.

### DIFF
--- a/rero_ils/modules/locations/serializers/json.py
+++ b/rero_ils/modules/locations/serializers/json.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+#
+# RERO ILS
+# Copyright (C) 2019-2023 RERO
+# Copyright (C) 2019-2023 UCLouvain
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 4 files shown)


rero_ils/modules/locations/serializers/json.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _postprocess_search_hit(self, hit):
+        """Post-process each hit of a search result."""
+        metadata = hit.get('metadata', {})
+
+        # Add label for some $ref fields.
+        pid = metadata.get('library', {}).get('pid')
+        if pid and (resource := self.get_resource(LibrariesSearch(), pid)):
+            metadata['library']['code'] = resource.get('code')
+            metadata['library']['name'] = resource.get('name')
```
jma, line 37: No translation right?

rerowep: Should we add the test for an pid into the `get_resource` function?

rerowep (approved):

rero_ils/modules/locations/serializers/json.py
zannkukai: @jma no translation. I get name and code encoded into the library resource.

Issue #2663, opened 2022-01-27
Update loans `end_date` when a closed date/period is added into library exceptions


zannkukai:
**How it works**

When a new "close" exception date/period is included into the library exception dates list, no modifications are done on loans ended during this exception.

**Improvement suggestion**

All loans ended during the new exception date/period should be updated to the next open day.

zannkukai:
I added the "effort: important" because it will be tricky to manage : 
  * exception_dates update detection.
   * repeatable exception management

PR #3367, opened 2023-06-08
library: detect changes on library open hours

zannkukai:
* Creates extension to detect any changes into library calendar (regular hours & exception dates). If changes are detected, run this extension starts a tasks to update related loans.
* Creates a task to update loans related a library if calendar changes.
* Closes rero/rero-ils#2663.

### DIFF
--- a/rero_ils/modules/libraries/extensions.py
+++ b/rero_ils/modules/libraries/extensions.py
@@ -1,0 +1,92 @@
+# -*- coding: utf-8 -*-
+#
+# RERO ILS
+# Copyright (C) 2019-2023 RERO
+# Copyright (C) 2019-2023 UCLouvain
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, version 3 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Invenio extensions for `Library` resources."""
+from celery import current_app as celery_app
+from deepdiff import DeepDiff
+from invenio_cache import current_cache
+from invenio_records.extensions import RecordExtension
+
+from .tasks import calendar_changes_update_loans
+
+
+class LibraryCalendarChangesExtension(RecordExtension):
+    """Handle any changes on library calendar.
+
+    When the library calendar changes, it could impact related active loans :
+    If the loan ends now at a closed day/exception date, this loan end_date
+    must be adapted to next_open_date.
+    As it could potentially concern many loans, this operation can't be
+    synchronously operated. So a detached Celery task will be called to execute
+    changes.
+
+    NOTE: If user operates multiple changes on library calendar in a small
+          period of time, we need to ensure than only last detached task will
+          be performed. So before running a new task, we MUST abort previously
+          running task for the same library.
+    """
+
+    def __init__(self, tracked_fields):
+        """Initialization method.
+
+        :param tracked_fields: (list<String>) list of tracked fields.
+        """
+        if not isinstance(tracked_fields, list) or not tracked_fields:
+            raise TypeError("'tracked_fields' is required")
+        self._changes_detected = False
+        self._tracked_fields = tracked_fields
+        super().__init__()
+
+    def pre_commit(self, record):
+        """Called before a record is committed.
+
+        :param record: the new record data.
+        """
+        db_record = record.db_record()
+        for field_name in self._tracked_fields:
+            original_field = db_record.get(field_name)
+            new_field = record.get(field_name)
+            if DeepDiff(original_field, new_field, ignore_order=True):
+                self._changes_detected = True
+                break
+
+    def post_commit(self, record):
+        """Called after a record is committed.
+
+        :param record: the `Library` updated record
+        """
+        if self._changes_detected:
+            self._changes_detected = False  # Reset changes detection
+            task = calendar_changes_update_loans.s(record).apply_async()
+            self._cache_current_task(record, task)
+
+    @staticmethod
+    def _cache_current_task(record, task):
+        """Store the task_id into the application cache.
+
+        :param record: the touched library record.
+        :param task: the task related to the library.
+        """
+        content = current_cache.get('library-calendar-changes') or {}
+        # If a previous task is still present into this cache entry, revoke it.
+        # DEV NOTE : the task MUST clean (remove) this cache entry when task is
+        #            finished.
+        if task_id := content.pop(record.pid, None):
+            celery_app.control.revoke(task_id, terminate=True)
+        content[record.pid] = task.id
+        current_cache.set('library-calendar-changes', content)
--- a/rero_ils/modules/libraries/tasks.py
+++ b/rero_ils/modules/libraries/tasks.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+#
+# RERO ILS
+# Copyright (C) 2019-2023 RERO
+# Copyright (C) 2019-2023 UCLouvain
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, version 3 of the License.
+#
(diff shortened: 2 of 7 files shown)


rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+
+"""Tasks on `Library` resource."""
+
+from celery import shared_task
+from invenio_cache import current_cache
+
+
+@shared_task(bind=True, name='library-calendar-changes-update-loans')
+def calendar_changes_update_loans(self, record_data):
+    """Task used to update...
```
jma, line 27: Change this comment.

rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+
```
jma, line 28: Can you explain that you considers only a day that becomes closes and not the reverse?

rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+        #   it's easier to create a small specific function.
+        cache_content = current_cache.get('library-calendar-changes') or {}
+        cache_content.pop(library.pid, {})
+        current_cache.set('library-calendar-changes', cache_content)
+
+    from rero_ils.modules.loans.api import LoansIndexer, \
+        get_on_loan_loans_for_library
+    from .api import Library
+
+    library = Library(record_data)
```
jma, line 53: Why not getting the library from the db?

rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+    for loan in get_on_loan_loans_for_library(library.pid):
+        active_loan_counter += 1
+        if not library.is_open(loan.end_date):
+            loan['end_date'] = library \
+                .next_open(loan.end_date) \
+                .astimezone(library.get_timezone()) \
+                .replace(hour=23, minute=59, second=0, microsecond=0)\
+                .isoformat()
+            changed_loan_uuids.append(loan.id)
+            loan.update(loan, dbcommit=True, reindex=False)
```
jma, line 64: the `dbcommit` can be done by bulk

rero_ils/modules/loans/api.py
```diff
@@ -1181,6 +1181,21 @@ def get_loans_count_by_library_for_patron_pid(patron_pid, filter_states=None):
...
+
+    :param library_pid: the library pid.
+    :returns a generator of `Loan` record.
+    """
+    query = current_circulation.loan_search_cls() \
+        .filter('term', library_pid=library_pid) \
+        .filter('term', state=LoanState.ITEM_ON_LOAN) \
+        .source(False)
+    ids = [hit.meta.id for hit in query.scan()]
+    for id_ in ids:
```
jma, line 1195: You can merge this line with the previous.

jma (approved):

rero_ils/modules/libraries/extensions.py
```diff
@@ -0,0 +1,90 @@
...
+    def post_commit(self, record):
+        """Called after a record is committed.
+
+        :param record: the `Library` updated record
+        """
+        if self._changes_detected:
+            self._changes_detected = False  # Reset changes detection
+            task = calendar_changes_update_loans\
+                .s(record)\
+                .apply_async(countdown=5)
```
rerowep, line 77: Why waiting 5 seconds to start the task?

rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Tasks on `Library` resource."""
+
+from celery import shared_task
+from invenio_cache import current_cache
+
+
+@shared_task(bind=True, name='library-calendar-changes-update-loans')
+def calendar_changes_update_loans(self, record_data):
```
rerowep, line 26: The documentation of python says:
The `self` parameter is a reference to the current instance of the class, and is used to access variables that belongs to the class.

This is not a `class` and I would prefer not to use `self` here.

rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+        cache_content = current_cache.get('library-calendar-changes') or {}
+        cache_content.pop(library.pid, {})
+        current_cache.set('library-calendar-changes', cache_content)
+
+    from rero_ils.modules.loans.api import LoansIndexer, \
+        get_on_loan_loans_for_library
+    from .api import Library
+
+    library = Library(record_data)
+    active_loan_counter, changed_loan_uuids = 0, []
```
rerowep, line 54: Yes it is possible to write it like this, but for readability we can effort one line more ?
```python
active_loan_counter = 0
changed_loan_uuids = []
```

rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,70 @@
...
+    for loan in get_on_loan_loans_for_library(library.pid):
```
rerowep, line 55: The proposed, preferred, faster way in python to do counts is to use `enumerate`.

tests/api/circulation/test_library_calendar_changes.py
```diff
@@ -0,0 +1,89 @@
...
+    #     updated.
+    library['opening_hours'][0] = {'day': 'sunday', 'is_open': False}
+    library['exception_dates'].append({
+        'is_open': False,
+        'title': 'exception date',
+        'start_date': loan.end_date[:10]
+    })
+    library = library.update(library, dbcommit=True, reindex=False)
+
+    time.sleep(5)  # TODO :: find a better way to detect task is finished.
```
rerowep, line 77: Task if finished if `library-calendar-changes` is cleaned?

rero_ils/modules/libraries/extensions.py
zannkukai: In the case when staff member click on "closed" for the bad day. 
He could edit the library to re-open the correct day.

Maybe 5sec isn't enough, difficult to determine the best gap.

rero_ils/modules/libraries/tasks.py
zannkukai: I didn't need the task reference (previously I check the task_id) so I can remove the `self` reference.
But using the `@shared_task(bind=True)`, it seems convenient/usage to use `self` as first argument.

rero_ils/modules/libraries/tasks.py
zannkukai: what's the difference between :
```python
counter = 0
for hit in my_generator():
   counter += 1
```
and 
```python
counter = 0
for idx, hit in enumerate(my_generator()):
    counter = idx
```

Assuming that `get_on_loan_loans_for_library()` is a generator (and must remain) and that `active_loan_counter` is the length of this generator.

rero_ils/modules/libraries/tasks.py
zannkukai: to save two database calls (one for persistent_identifier, one for data).
My task receive the dumped `Library` dict (not possible to pass a `Library` instance to Celery).
So I recreate the `Library` instance from data without any db call (that get me, 99,999999%, the same data - the possible difference is if library record is updated between task creation and task running process).

rero_ils/modules/libraries/extensions.py
zannkukai: It was for debugging. No need to wait... just need to run asynchronously.

rero_ils/modules/libraries/tasks.py
rerowep: No like this:
```python
counter = 0
for counter, hit in enumerate(my_generator(), 1):
    ...
```

rero_ils/modules/libraries/tasks.py
zannkukai: I tried to be more explicit with my french_english.... :-)

rero_ils/modules/libraries/tasks.py
```diff
@@ -0,0 +1,89 @@
...
+    from .api import Library
+
+    library = Library(record_data)
+    active_loan_counter = 0
+    changed_loan_uuids = []
+    for loan in get_on_loan_loans_for_library(library.pid):
+        active_loan_counter += 1
+        if not library.is_open(loan.end_date):
+            loan['end_date'] = library \
+                .next_open(loan.end_date) \
```
PascalRepond, line 78: In the unlikely case where a library has no standard opening_hours and we created an exception open date in the past, next_open is going to get stuck in an infinite while loop.
With PR #3375, it will raise a LibraryNeverOpen exception. Maybe take this into account for added robustness?

lauren-d (approved):

PascalRepond (approved):

PR #3372, opened 2023-06-13
operation logs: assign current library request

### LINKED ISSUE #3310: users: library used for the operation logs is wrong
### How it works

Currently, `patron.libraries` contains all libraries where the user has rights, in an unordered list. This causes problems for librarians with multiple libraries since when we need a user's library in some data, we use the first of the list.

### Improvement suggestion

- Find a way to use the current active library (from the UI switch-library menu) everywhere we need a professional user's library to inject in operation logs:
    - document
    - item
    - holding
    - Other? **To be specified**

Garfield-fr:
When the current_library parameter is present in the request, it is assigned to the library in the logs operation.

* Closes #3310.

### DIFF
--- a/rero_ils/modules/operation_logs/extensions.py
+++ b/rero_ils/modules/operation_logs/extensions.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # RERO ILS
-# Copyright (C) 2019-2022 RERO
+# Copyright (C) 2019-2023 RERO
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
@@ -22,6 +22,7 @@ from functools import partialmethod
 
 import pytz
 from deepdiff import DeepDiff
+from flask import request as flask_request
 from invenio_records.extensions import RecordExtension
 
 from rero_ils.modules.patrons.api import current_librarian
@@ -61,7 +62,7 @@ class OperationLogObserverExtension(RecordExtension):
             'user_name': 'system'  # default value, could be override
         }
         if current_librarian:
-            oplg.update({
+            oplg |= {
                 'user_name': current_librarian.formatted_name,
                 'user': {
                     'type': 'ptrn',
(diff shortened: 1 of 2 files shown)


rero_ils/modules/operation_logs/extensions.py
```diff
@@ -61,25 +63,42 @@ def _build_operation_log(self, record, operation):
...
+            oplg |= {
                 'user_name': current_librarian.formatted_name,
                 'user': {
                     'type': 'ptrn',
                     'value': current_librarian.pid
                 },
                 'organisation': {
                     'type': 'org',
-                    'value': current_librarian.organisation_pid
+                    'value': current_librarian.organisation_pid,
```
zannkukai, line 74: change not needed

rero_ils/modules/operation_logs/extensions.py
```diff
@@ -61,25 +63,42 @@ def _build_operation_log(self, record, operation):
...
+                        'organisation': {
+                            'type': 'org',
+                            'value': extracted_data_from_ref(
+                                library.get('organisation'))
+                        },
+                        'library': {
+                            'type': 'lib',
+                            'value': library.get('pid')
+                        }
+                    }
```
zannkukai, line 96: you should avoid to load `library` to get organisation using `current_librarian.organisation_pid`.
```python
if (lib_pid := flask_request.args.get('current_library')) and lib_pid in current_librarian.manageable_library_pids:
    oplg |= {
          'organisation': {
              'type': 'org',
              'value': current_librarian.organisation_pid
          },
          'library': {
              'type': 'lib',
              'value': lib.pid
          }
      }
```

Garfield-fr: I did this to make sure I had the right organization for the library. It was for the future if a librarian can be admin in several organizations.

jma (approved):

PascalRepond (approved):

zannkukai (approved):

PR #3384, opened 2023-06-23
public view: disable masked documents detailed

### LINKED ISSUE #1789: User can display masked documents in public interface if s/he knows the document pid
A masked document is usually not visible in the public interface.
However, if a user knows the document pid, s/he can access the document directly by visiting the URL:

https://~/documents/2000537

1. Mask a document in the pro interface and take a note of the document_pid
2. Logout and visit the public interface
3. visit the document page: https://~/documents/<document_pid>.
4.  You can display the document and its unmasked holdings/items

* server:  [ilsdev.test.rero.ch]
* version: `v1.1.0` 

masked record in the pro interface
Screenshot 2021-03-22 at 15 48 37

record is visible by accessing directly the document URL
Screenshot 2021-03-22 at 15 48 50

PascalRepond:
- Closes #1789.

## How to test?

- Set a document as masked and try to access it from the public view with its permalink. It should display an info message instead of the document detailed view.


zannkukai (commented):
Could be done with permissions ? 
Updating the "doc_read" permissions.

PascalRepond:
> Could be done with permissions ? Updating the "doc_read" permissions.

Wouldn't it imply creating a new permission class to restrict by logged in user AND by "_masked" field. Might be overkill and restrict the public API, which we probably don't want?

Garfield-fr (changes requested):
I don't think this is the right solution. The document should not be displayed if it is hidden (public view). We need to take action at view controller level. A 404 exception should be lifted.

PascalRepond:
> I don't think this is the right solution. The document should not be displayed if it is hidden (public view). We need to take action at view controller level. A 404 exception should be lifted.

@Garfield-fr You're right. I changed it, thank you.

Garfield-fr:
> > I don't think this is the right solution. The document should not be displayed if it is hidden (public view). We need to take action at view controller level. A 404 exception should be lifted.
> 
> @Garfield-fr You're right. I changed it, thank you.

You also need to check the export function, and if ever you need to use the same logic

PR #3385, opened 2023-06-26
entities: configure operation logs for local entities

lauren-d:
* Uses default operation log extension.

Implements task: https://tree.taiga.io/project/rero21-reroils/task/2668


rero_ils/modules/entities/local_entities/api.py
```diff
@@ -35,6 +35,8 @@
 from ..api import Entity
 
 # provider
+from ...operation_logs.extensions import OperationLogObserverExtension
```
zannkukai, line 38: absolute import path

zannkukai (approved):

rero_ils/modules/entities/local_entities/api.py
lauren-d: Done

PR #3389, opened 2023-06-29
items+patrons: fix unique barcodes

### LINKED ISSUE #3386: Barcode fields should be unique
### How it works

For circulation purposes, the barcode field in the item should be unique. In the editor, the validation "valueAlreadyExists" prevents me to save an item if the barcode exists in another organisation. However, there seems to be no validation process at item creation (except for the editor).

- Creating items with the same barcode from the API doesn't raise an error.
- In the editor, if I enter the same data in different tabs before saving and then save the two tabs, two items can be created with the same barcode. 

### Improvement suggestion
- When creating or updating an item, there should be a data validation to prevent having the same barcode within an instance.
- Same thing should be implemented for patron barcodes but with only check uniquess within organisation becaus a patron can use the same library card barcode in multiple organisations.

PascalRepond:
- Adds an extended validation method to patrons and items to ensure that barcode fields are unique.
- Deletes barcodes in item test fixtures so that they are generated and avoid duplicate barcode errors on item creation.
- Fixes test errors related to duplicate item and patron barcodes.
- Closes #3386.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

### DIFF
--- a/rero_ils/modules/items/api/record.py
+++ b/rero_ils/modules/items/api/record.py
@@ -19,6 +19,7 @@
 from datetime import datetime, timezone
 
 import pytz
+from contextlib import suppress
 from elasticsearch_dsl.query import Q
 from flask_babelex import gettext as _
 
@@ -54,6 +55,8 @@ class ItemRecord(IlsRecord):
     def extended_validation(self, **kwargs):
         """Add additional record validation.
 
+        Ensures that barcode field is unique.
+
         Ensures that item of type issue is created only on
         holdings of type serials.
 
@@ -64,7 +67,8 @@ class ItemRecord(IlsRecord):
 
         Ensures that only one note of each type is present.
 
-        :return: False if
+        :return: Error message if
+            - barcode already exists
             - holdings type is not journal and item type is issue.
             - item type is journal and field issue exists.
             - item type is standard and field issue does not exists.
@@ -73,6 +77,17 @@ class ItemRecord(IlsRecord):
             - temporary_item_type isn't in the future (if specified)
 
         """
+        from . import ItemsSearch
+        if barcode := self.get('barcode'):
+            if (
+                ItemsSearch()
+                .exclude('term', pid=self.pid)
+                .filter('term', barcode=barcode)
+                .source('pid')
+                .count()
+            ):
+                return _(f'Barcode {barcode} is already taken.')
+
         from ...holdings.api import Holding
         holding_pid = extracted_data_from_ref(self.get('holding').get('$ref'))
         holding = Holding.get_record_by_pid(holding_pid)
@@ -351,7 +366,7 @@ class ItemRecord(IlsRecord):
             yield item_pid_to_object(item.pid)
 
     @classmethod
-    def get_item_by_barcode(cls, barcode, organisation_pid):
+    def get_item_by_barcode(cls, barcode, organisation_pid=None):
         """Get item by barcode.
 
         :param barcode: the item barcode.
@@ -362,15 +377,15 @@ class ItemRecord(IlsRecord):
         :rtype `rero_ils.modules.items.api.api.Item`
         """
         from . import ItemsSearch
+        filters = Q('term', barcode=barcode)
+        if organisation_pid:
+            filters &= Q('term', organisation__pid=organisation_pid)
         results = ItemsSearch()\
-            .filter('term', barcode=barcode)\
-            .filter('term', organisation__pid=organisation_pid)\
+            .filter(filters)\
             .source(includes='pid')\
             .scan()
-        try:
+        with suppress(StopIteration):
             return cls.get_record_by_pid(next(results).pid)
-        except StopIteration:
-            return None
 
     def get_organisation(self):
         """Shortcut to the organisation of the item location."""
--- a/rero_ils/modules/patrons/api.py
+++ b/rero_ils/modules/patrons/api.py
@@ -23,7 +23,7 @@ from functools import partial
 
 from elasticsearch_dsl import Q
 from flask import current_app
-from flask_babelex import gettext as _
+from flask_babelex import ngettext, gettext as _
 from flask_login import current_user
 from invenio_circulation.proxies import current_circulation
 from invenio_db import db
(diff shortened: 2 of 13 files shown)


rero_ils/modules/patrons/jsonschemas/patrons/patron-v0.0.1.json
```diff
@@ -184,16 +184,6 @@
             "form": {
               "templateOptions": {
                 "doNotSubmitOnEnter": true
-              },
-              "validation": {
```
jma, line 188: Seems a bad idea. The frontend should be able to do the validation. You can probably do this in a similar way than the locations: https://github.com/rero/rero-ils/blob/469b76f275b535c61dd634cffefe8213eb13d931/rero_ils/modules/locations/jsonschemas/locations/location-v0.0.1.json#L88.

rero_ils/modules/patrons/api.py
```diff
@@ -122,6 +122,33 @@ class Patron(IlsRecord):
...
+
+        :return: Error message if
+            - barcode already exists in organisation
+        """
+        from .api import Patron
+        org_pid = self.organisation_pid
+        ptrn = self.get('patron', {})
+        barcodes = ptrn.get('barcode', [])
+        taken_barcodes = []
+        for barcode in barcodes:
```
jma, line 138: I would do that using an elasticsearch query.

jma (changes requested):

rero_ils/modules/patrons/jsonschemas/patrons/patron-v0.0.1.json
PascalRepond: There is no organisation pid in the data model of the patron so I'm not sure how to filter.

rero_ils/modules/items/api/record.py
```diff
@@ -362,15 +381,17 @@ def get_item_by_barcode(cls, barcode, organisation_pid):
         :rtype `rero_ils.modules.items.api.api.Item`
         """
         from . import ItemsSearch
+        if not barcode:
+            return None
```
zannkukai, line 385: I don't understand why you add the possibility to don't set `barcode` parameter and add a check `if not barcode: return None`. Why this parameter isn't required ? Additionally the name of the function is `get_item_by_barcode`, without barcode parameter, it seems strange to use this method...

rero_ils/modules/patrons/api.py
```diff
@@ -122,6 +122,39 @@ class Patron(IlsRecord):
...
+                .source(['pid', 'patron.barcode'])\
+                .scan()
+            taken_barcodes = []
+            for hit in results:
+                if hit.pid != self.get('pid'):
+                    taken_barcodes.extend(
+                        barcode
+                        for barcode in hit.patron.barcode
+                        if barcode in patron_barcodes
+                    )
```
zannkukai, line 148: To be tested 

```python
taken_barcodes = [
   barcode
   for barcode in patron_barcodes
   if (patron := Patron.get_patron_by_barcode(barcode, org_pid)) and patron.pid != self.get('pid')
]
```

additionally, seems strange too to have possibility to have `None` as barcode parameter value for `get_patron_by_barcode` (don't "blame" who wrote this :-) I just reorganized the class - I didn't write this method ^^)

PascalRepond: @zannkukai This might work but it creates multiple API calls if the patron has multiple barcodes instead of one ES query.

rero_ils/modules/items/api/record.py
PascalRepond: You are right, I based it on the get_patron_by_barcode but the barcode param should be required in both cases.

rero_ils/modules/items/api/record.py
```diff
@@ -73,6 +77,17 @@ def extended_validation(self, **kwargs):
             - temporary_item_type isn't in the future (if specified)
 
         """
+        from . import ItemsSearch
+        if barcode := self.get('barcode'):
+            filters = Q('term', barcode=barcode)
+            results = ItemsSearch()\
+                .filter('bool', must=[filters])\
+                .source(['pid'])\
+                .scan()
```
zannkukai, line 86: It isn't necessary to create `Q` in this case. 
If you need only one field, you can avoid brackets

You could simply write
```python
results = ItemsSearch()\
    .filter('term', barcode=barcode)\
    .source('pid').scan()
```

rero_ils/modules/items/api/record.py
```diff
@@ -362,15 +377,15 @@ def get_item_by_barcode(cls, barcode, organisation_pid):
         :rtype `rero_ils.modules.items.api.api.Item`
         """
         from . import ItemsSearch
+        filters = Q('term', barcode=barcode)
+        if organisation_pid:
+            filters &= Q('term', organisation__pid=organisation_pid)
         results = ItemsSearch()\
-            .filter('term', barcode=barcode)\
-            .filter('term', organisation__pid=organisation_pid)\
+            .filter('bool', must=[filters])\
```
zannkukai, line 384: you could write `.filter(filters)` in this case (just discovered and tested yesterday)

rero_ils/modules/patrons/api.py
```diff
@@ -122,6 +122,39 @@ class Patron(IlsRecord):
...
+            - barcode already exists in organisation
+        """
+        org_pid = self.organisation_pid
+        if patron_barcodes := self.get('patron', {}).get('barcode', []):
+            filters = Q('terms', patron__barcode=patron_barcodes)
+            filters &= Q('term', organisation__pid=org_pid)
+            results = PatronsSearch()\
+                .filter('bool', must=[filters])\
+                .source(['pid', 'patron.barcode'])\
+                .scan()
```
zannkukai, line 140: maybe more readable (check other devs opinion)

```python
results = PatronsSearch()\
    .filter('terms', patron__barcode=patron_barcodes)\
    .filter('term', organisation__pid=org_pid)\
    .source(['pid', 'patron.barcode']).scan()
```

rero_ils/modules/patrons/api.py
zannkukai: you're right ; but in my opinion it's better to delegate the search job to a specific method instead of write the query here. Maybe the best solution is to adapt `get_patron_by_barcode()` method to accept one or multiple barcodes parameters... See with other devs opinion (@rerowep, @lauren-d, @jma, @Garfield-fr)

zannkukai (approved):
Your code should work even without integrate my comments ; so I approve.
Could be interesting to have opinion from other devs on some comments.
Thx for this job 👍👍

tests/unit/documents/test_documents_dojson_marc21.py
```diff
@@ -685,6 +685,8 @@ def test_holdings_items_to_marc21(app, marc21_record, document,
     assert result == record
 
     record = {'pid': document.pid}
+    item2_lib_sion['barcode'] = '87121336'
+    item2_lib_sion.update(item2_lib_sion, dbcommit=True, reindex=True)
```
jma, line 689: Did you forget to restore the initial value for item2_lib_sion at the end of the test?

rerowep (approved):

PR #3395, opened 2023-07-04
loan: fix LoansIndexer

rerowep:
* Deletes wrong function `_get_record_class` in `LOansIndexer` class.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?


rero_ils/modules/api.py
```diff
@@ -629,6 +629,9 @@ def _get_record_class(self, payload):
 
         # take the first defined doc type for finding the class
         pid_type = payload.get('doc_type', 'rec')
+        # correct loan
+        if pid_type == 'loan':
+            pid_type = 'loanid'
```
zannkukai, line 634: I'm not fan about specific `if` statement into a generic function.... 
I will keep the current existing `LoanIndexer._get_record_class()` function but, of course, returning a correct value.
What's other devs opinion

rerowep: `_get_record_class()` is not working for the task

rerowep: Sorry have not corrected it on the right place.

jma (approved):

PascalRepond (approved):

PR #3396, opened 2023-07-04
circulation: fix has_is_open library method

PascalRepond:
- Fixes `_has_is_open` library method so that it doesn't return True if the library has no standard opening hours but has an exception closed date in the future.

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?


rero_ils/modules/libraries/api.py
```diff
@@ -193,11 +193,12 @@ def _has_is_open(self):
...
-                # avoid next_open infinite loop if an open exception date is
-                # in the past
-                if start_date > datetime.now(pytz.utc):
-                    return True
+                if exception_date['is_open']:
+                    date = date_string_to_utc(exception_date['start_date'])
+                    # avoid next_open infinite loop if an exception date
+                    # exists in the past
+                    if date > datetime.now(pytz.utc):
+                        return True
```
zannkukai, line 201: maybe I will wrote 
```python
...
current_timestamp = datetime.now(pytz.utc)
for date in filter(lambda d: d['is_open'], self.get('exception_dates', []):
    start_date = date_string_to_utc(exception_date['start_date'])
    # avoid next_open infinite loop if an open exception date is
    # in the past
    if start_date > current_timestamp:
        return True
...
```

But your code should work too

zannkukai (approved):

rero_ils/modules/libraries/api.py
PascalRepond: @zannkukai Thanks, you're right it's more elegant. :)

rerowep (approved):

PascalRepond:
> It would be great to have a test.

@jma @rerowep Done.

PR #3397, opened 2023-07-04
entities: link to document

lauren-d:
* Create generic method to load entity.
* Adapts jinja filter for document detail view.
* Reorganizes dumpers for entities.
* Adapts tests.

Implements task: https://tree.taiga.io/project/rero21-reroils/task/2704


rero_ils/modules/entities/local_entities/indexer.py
```diff
@@ -0,0 +1,72 @@
...
+
+    record_cls = LocalEntity
+    # data dumper for indexing
+    record_dumper = indexer_dumper
+
+    def index(self, entity, arguments=None, **kwargs):
+        """Index a Local entity record."""
+        super().index(entity)
+        eta = datetime.utcnow() + current_app.config[
+            "RERO_ILS_INDEXER_TASK_DELAY"]
```
zannkukai, line 64: use `.get("RERO_ILS_INDEXER_TASK_DELAY", 0)` to be safe

rero_ils/modules/ext.py
```diff
@@ -149,6 +150,95 @@ def load_actions(sender, app):
...
+    # TODO: we can also migrate other method like:
+            - rero_ils.utils:
+                * get_endpoint_configuration
+                * extracted_data_from_ref
+                * get_pid_type_from_schema
+                * get_schema_for_resource
+                * get_ref_for_pid
+        maybe we use this implementation everywhere is it possible.
+    """
+
```
zannkukai, line 165: As discuss in real life, i think it's maybe overkill... but a nice way.
What is your opinion @jma ?

rero_ils/modules/entities/local_entities/indexer.py
lauren-d: done

rero_ils/modules/ext.py
jma: If I understand well the purpose of this, it seems to me a bit overkill. We will not "feel" the difference in term of the use of the system. Moreover, this can have some side effects, for example in the test we modify the app.config on the fly and if we cache some informations, we need to invalidate the cache which can takes several hours to understand or discover. If we decide to do it anyway, we need to have a very good documentation.

PR #3398, opened 2023-07-05
entities: proxy to search local entities.

zannkukai:
Creates an API endpoint to search local entities by category (a category
could represent multiple ES filters coupling AND/OR/... filters).


rero_ils/modules/entities/local_entities/proxy.py
```diff
@@ -0,0 +1,69 @@
...
+    'person': Q('term', type=EntityType.PERSON),
+    'organisation': Q('term', type=EntityType.ORGANISATION),
+    'concepts': Q('term', type=EntityType.TOPIC),
+    'concepts-genreForm':
+        Q('term', type=EntityType.TOPIC) & Q('term', genreForm=True)
+}
+
+
+class LocalEntityProxy:
+    """Local entity proxy."""
```
lauren-d, line 37: I know the context, but from my point of view, it is not really a proxy.
I think this is better to do that in search class.

lauren-d (approved):

Issue #3379, opened 2023-06-20
Claim: error 500 if the vendor is not indicated on the holding


ManaDeweerdt:
### Bug description:

If the vendor field is missing from the holding, when we try to add a claim we get an error 500.
![image](https://github.com/rero/rero-ils/assets/12396034/abda379c-5a65-45f1-aa3a-353891715e88)

### Expected behavior:

- The message must be changed to: `Error: Vendor is missing in the holdings record`. 
- This message must be translated for the user.

### Context

**Server:**
* dev: [ilsdev.test.rero.ch](https://ilsdev.test.rero.ch)

**Version:**

rero-ils version [1c384b04](https://github.com/rero/rero-ils/commit/1c384b04baafa439b8b1f54cbebe7cccf664ef83)
rero-ils-ui version [41b95197](https://github.com/rero/rero-ils-ui/commit/41b951974ea7f5bf22209dc0276799da08c50796)

PR #3399, opened 2023-07-10
claims: manage missing data for claim notification

zannkukai:
* Closes rero/rero-ils#3379.
* Checks if all requires data are available to render the claim notification content. If not, return a specific exception to specify the missing data.

### DIFF
--- a/rero_ils/modules/items/dumpers.py
+++ b/rero_ils/modules/items/dumpers.py
@@ -20,6 +20,7 @@
 
 from invenio_records.dumpers import Dumper as InvenioRecordsDumper
 
+from rero_ils.modules.commons.exceptions import MissingDataException
 from rero_ils.modules.documents.dumpers import \
     TitleDumper as DocumentTitleDumper
 from rero_ils.modules.holdings.api import Holding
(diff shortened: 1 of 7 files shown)


Garfield-fr (approved):

rero_ils/modules/items/dumpers.py
```diff
@@ -82,19 +83,23 @@ class ClaimIssueNotificationDumper(InvenioRecordsDumper):
 
     def dump(self, record, data):
         """Dump an item issue for claim notification generation."""
-        assert record.is_issue, 'record must be an `ItemIssue` resource'
-        assert (holding := record.holding), \
-            'record must be related to an existing `Holding`'
+        if not record.is_issue:
+            raise TypeError('record must be an `ItemIssue` resource')
+        if not (holding := record.holding):
+            raise MissingDataException('item.holding')
```
jma, line 89: Why not using a KeyError exception?

rero_ils/modules/items/views/api_views.py
```diff
@@ -572,7 +573,11 @@ def claim_notification_preview(item_pid):
     if not record.is_issue:
         abort(400, 'Item isn\'t an issue')
 
-    issue_data = record.dumps(dumper=ClaimIssueNotificationDumper())
+    try:
+        issue_data = record.dumps(dumper=ClaimIssueNotificationDumper())
+    except (TypeError, MissingDataException) as exp:
+        abort(500, str(exp))
```
jma, line 579: Can we return an other status than `500`? Is that means that the request should not be done for this kind of items? Or is it a mistake in the data?

jma (approved):

rero_ils/modules/items/dumpers.py
zannkukai: To be more `specific` in the error management ; additionally `MissingDataException` inherits from `KeyError`

rero_ils/modules/items/views/api_views.py
zannkukai: I don't found any other HTTP status code : It's an error from server (not from client sending data) --> so 5xx family error. According https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_server_errors, 500 seems the best possible option.

PR #3425, opened 2023-08-02
SRU: correct unnecessary log for bf:Temporal transformation

rerowep:

- Which issue does it fix?

## Dependencies

* rero/rero-ils-ui#<xx>

## How to test?

### DIFF
--- a/rero_ils/modules/documents/dojson/contrib/jsontomarc21/model.py
+++ b/rero_ils/modules/documents/dojson/contrib/jsontomarc21/model.py
@@ -708,6 +708,12 @@ def reverse_subjects(self, key, value):
     Sujet Organisation > 610 OU 611 Conference
     Sujet Concept > 650
     """
+    def add_identified_by(result, identified_by):
+        """Adds $2 and $0 to result."""
+        result = add_value(result, '2', identified_by['type'].lower())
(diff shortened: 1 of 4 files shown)


PascalRepond (approved):

PascalRepond (approved):

rero_ils/modules/documents/dojson/contrib/jsontomarc21/model.py
```diff
@@ -708,6 +708,12 @@ def reverse_subjects(self, key, value):
     Sujet Organisation > 610 OU 611 Conference
     Sujet Concept > 650
     """
+    def add_identified_by(result, identified_by):
+        """Adds $2 and $0 to result."""
+        result = add_value(result, '2', identified_by['type'].lower())
+        result = add_value(result, '0', identified_by['value'])
+        return result
```
zannkukai, line 715: You don't need to return `result` because python use reference for parameters. 
You can just write :
```
def add_identified_by(result, identified_by):
    """Adds $2 and $0 to result."""
    result = add_value(result, '2', identified_by['type'].lower())
    result = add_value(result, '0', identified_by['value'])
```

and use `add_identified_by(result, identified_by)` without affectation into result.

PR #3427, opened 2023-08-07
document: add 3 new type of subjects

Garfield-fr:
* Adds place, temporal and work type on subject filters.


PascalRepond:
Will this really be enough for the autocomplete to work as intended?

PascalRepond (approved):

zannkukai (approved):
editor should work as expected but need to be tested

PR #3450, opened 2023-08-31
loan: add location pickup_name on dump

### LINKED ISSUE #3433: Circulation UI: display pickup location for at_desk check-in
### Problem description
When checking in an item that is or becomes "at_desk", if my library has more than one pickup location, there is no way to know to which location the item will be picked up. It only shows the item status "At desk".

### Missing feature
When an item is at desk, the component should display the pickup location name after the status so that I can differentiate if my library has multiple pickup locations. 

### Mockup
image

Garfield-fr:
* Adds the location pickup_name field on circulation information.
* Closes #3433.

### DIFF
--- a/rero_ils/modules/loans/dumpers.py
+++ b/rero_ils/modules/loans/dumpers.py
@@ -47,14 +47,13 @@ class CirculationDumper(InvenioRecordsDumper):
             data['patron']['name'] = ', '.join((
                 ptrn_data.last_name, ptrn_data.first_name))
 
-        # only for pending requests
-        if record.get('pickup_location_pid') \
-                and record.get('state') == LoanState.PENDING:
+        if record.get('pickup_location_pid'):
             location = Location.get_record_by_pid(
                 record.get('pickup_location_pid'))
             data['pickup_location'] = {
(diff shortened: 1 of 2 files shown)


rerowep (approved):

zannkukai (approved):

jma (approved):

rero_ils/modules/loans/dumpers.py
```diff
@@ -47,14 +47,20 @@ def dump(self, record, data):
...
                 record.get('pickup_location_pid'))
+            library = location.get_library()
             data['pickup_location'] = {
                 'name': location.get('name'),
-                'library_name': location.get_library().get('name')
+                'library_name': library.get('name'),
+                'pickup_name': location.get(
+                    'pickup_name',
+                    f'{library["code"]}: {location["name"]}'
+                )
```
zannkukai, line 63: Maybe I create a `Location.pickup_name` property method and use it here. Maybe the same behavior could be used other places into the app in future... But implementation is OK now.

zannkukai (approved):

zannkukai (approved):

PR #3463, opened 2023-09-20
acquisitions: sort lines in order

### LINKED ISSUE #2653: Order lines in the notifications are not in the same sequence than in the order detailed view
**Describe the bug**

We have an order with various lines linked to the same document: different volumes are wanted
image

When generating the notification, the lines are not anymore in the same sequence, which is confusing for the vendor.

**To Reproduce**

1. Create an order with at least three lines
2. Send the order
3. See the preview: the lines are not in the same sequence

**Expected behavior**

* In the notification, the lines are in the same sequence than in the order detailed view
* [nice to have] I can sort the lines by title, date added, priority
    * ideally, I can reorder them manually

**Context**

* server: bib.rero.ch
* version: `v1.8.0`

[1]: https://ils.test.rero.ch
[2]: https://ilsdev.test.rero.ch

Garfield-fr:
Line sorting is now done on the PID to keep the same order as the command.

* Closes #2653.

### DIFF
--- a/rero_ils/modules/acquisition/acq_orders/api.py
+++ b/rero_ils/modules/acquisition/acq_orders/api.py
@@ -350,6 +350,16 @@ class AcqOrder(AcquisitionIlsRecord):
         :param includes: a list of statuses to include order lines.
         :return a generator of related order lines (or length).
         """
+
+        def preserve_order(query):
+            """Keep results in order.
+
+            :param: query: the es query.
+            :return the es query.
+            """
+            return query.params(preserve_order=True) \
+                .sort({'pid': {"order": "asc"}})
+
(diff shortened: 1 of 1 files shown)


lauren-d (approved):

rero_ils/modules/acquisition/acq_orders/api.py
```diff
@@ -357,9 +357,14 @@ def get_order_lines(self, output=None, includes=None):
         if output == 'count':
             return query.count()
         elif output == 'query':
-            return query
+            return query \
+                .params(preserve_order=True) \
```
jma, line 361: You can probably modify the query variable instead of adding this two times.

jma (approved):

jma (approved):

PR #3511, opened 2023-11-06
remote_entities: fix ReplaceIdentifiedBy

rerowep:
* Fixes wrong source pid for updated MEF records.

### DIFF
--- a/rero_ils/modules/entities/remote_entities/replace.py
+++ b/rero_ils/modules/entities/remote_entities/replace.py
@@ -96,17 +96,16 @@ class ReplaceIdentifiedBy(object):
         self.logger.warning(f'Problem get {url}: {res.status_code}')
         return {}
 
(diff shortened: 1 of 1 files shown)


lauren-d (approved):

rero_ils/modules/entities/remote_entities/replace.py
```diff
@@ -106,7 +106,7 @@ def _find_other_source(self, source, source_pid, mef_data):
         :returns: found source and source pid
         """
         if source in ('idref', 'gnd'):
-            return source, source_pid
+            return source, mef_data[source]['pid']
```
PascalRepond, line 109: You could delete param source_pid since you don't use it?

PascalRepond (approved):

zannkukai (approved):

Issue #3457, opened 2023-09-07
Serials: items that are not `received` should not have a `received_date`


PascalRepond:
### How it works
Serial items can currently have an `issue.received_date` field even when their `issue.status` is not RECEIVED. This is inconsistent data.

When an issue becomes late, it is created by the system through the `receive_regular_issue` function which adds a received_date. Then, the status is changed to LATE but the `issue.received_date` field is kept. 

### Improvement suggestion
- Fix the behavior for issue management by the system so that it never saves a not received item with a `received_date`
- Add data validation if necessary
- Add tests

PR #3522, opened 2023-11-09
items: fix late issue

rerowep:
* Deletes `received_date` from issue if LATE.
* Closes: #3457.


PascalRepond (approved):

zannkukai (commented):
Are you sure that it works also with PU/POST API calls ?
It should be tested. If not, then maybe a good place would be into an `extentions` (pre_commit, pre_create)

Garfield-fr (approved):

lauren-d (approved):

zannkukai (approved):

PR #3533, opened 2023-11-20
ill_request: add month filter

### LINKED ISSUE #3527: Hide ILL requests older than 6 months in the patron views
# Reformulation

### How it works
With the current implementation (https://github.com/rero/rero-ils/pull/3533 in v1.21.0) there is no way to display an ILL request that is more than 6 months old unless it is pending. Either the ILL Requests search view or the API will hide them. The only way to see these resource is through a CLI, which is is not what was described. API must return unfiltered data.

### Improvement suggestion

- **ONLY in the patron views** (`/professional/circulation/patron/xxxxx/ill`), only display the ILL requests that have been _created_ less than 6 months ago (unless the status is 'pending' or 'validated').
- In the public view (`/patrons/profile`), remove current filter by status and use the exact same behaviour as described above.
- In the ill_requests API and in the professionnal ILL list (`/professional/records/ill_requests`), only filter the list same as before 1.21.0. according to usual permissions (logged user -> organization, etc.).
- [Nice-to-have] If possible, to facilitate testing, create some ILL requests in the fixtures with a creation date older than six months, with each possible status. 

---

# Original issue

### How it works
There is currently no anonymization mechanism for ILL requests because these resources are needed longer than standard loans. This means most ILL requests are shown in the patron's history (public + pro).

### Improvement suggestion

- To limit the displayed lists in the patron view (public + pro), filter it to not display requests that have been created more than 6 months ago (unless the status is still 'pending').
- In the public view, remove current filter by status and use the same behaviour as described above.

Garfield-fr:
* closes #3527.

### DIFF
--- a/rero_ils/modules/ill_requests/api.py
+++ b/rero_ils/modules/ill_requests/api.py
@@ -18,8 +18,12 @@
 
 """API for manipulating ill_requests."""
 
+from datetime import datetime, timezone
 from functools import partial
 
+from dateutil.relativedelta import *
+from elasticsearch_dsl.query import Q
+from flask import current_app
 from flask_babelex import gettext as _
 
 from rero_ils.modules.api import IlsRecord, IlsRecordsIndexer, IlsRecordsSearch
@@ -31,7 +35,7 @@ from rero_ils.modules.utils import extracted_data_from_ref
 
 from .extensions import IllRequestOperationLogObserverExtension
 from .models import ILLRequestIdentifier, ILLRequestMetadata, \
-    ILLRequestNoteStatus
+    ILLRequestNoteStatus, ILLRequestStatus
 
 # provider
 ILLRequestProvider = type(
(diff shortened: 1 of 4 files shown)


rero_ils/modules/ill_requests/api.py
```diff
@@ -60,7 +64,15 @@ def get_ill_requests_total_for_patron(self, patron_pid):
...
         :return: return total of ill requests.
         """
-        return self.filter('term', patron__pid=patron_pid).count()
+        months = current_app.config.get('RERO_ILS_ILL_HIDE_MONTHS', -6)
+        date_delta = datetime.now() + relativedelta(months=months)
+        filters = Q(
+            'range',
+            _created={'lte': 'now', 'gte': date_delta}
+        )
+        filters |= Q('term', status='pending')
```
PascalRepond, line 73: You could use class ILLRequestStatus constant.

rero_ils/modules/ill_requests/api.py
```diff
@@ -60,7 +64,15 @@ def get_ill_requests_total_for_patron(self, patron_pid):
         :param patron_pid: the patron pid being searched.
         :return: return total of ill requests.
         """
-        return self.filter('term', patron__pid=patron_pid).count()
+        months = current_app.config.get('RERO_ILS_ILL_HIDE_MONTHS', -6)
```
PascalRepond, line 67: Maybe adapt the docstring to give the info that the total includes only last six months (total active ill requests).

PascalRepond (approved):

rero_ils/modules/ill_requests/api.py
```diff
@@ -55,12 +59,22 @@ class Meta:
...
+        """Get total of ill requests greater than delta date to a patron.
+
+        Months defined in config.py.
 
         :param patron_pid: the patron pid being searched.
         :return: return total of ill requests.
         """
-        return self.filter('term', patron__pid=patron_pid).count()
+        months = current_app.config.get('RERO_ILS_ILL_HIDE_MONTHS', -6)
+        date_delta = datetime.now() + relativedelta(months=months)
```
rerowep, line 70: Do we have to work with UTC?

rero_ils/query.py
```diff
@@ -280,10 +282,17 @@ def ill_request_search_factory(self, search, query_parser=None):
         search = search.filter(
             'terms',
             patron__pid=[ptrn.pid for ptrn in current_patrons])
-    # exclude to_anonymize records
-    search = search.filter('bool', must_not=[Q('term', to_anonymize=True)])
 
-    return search, urlkwargs
+    months = current_app.config.get('RERO_ILS_ILL_HIDE_MONTHS', -6)
+    date_delta = datetime.now() + relativedelta(months=months)
```
rerowep, line 287: see above

rerowep (approved):

jma (approved):

PR #3535, opened 2023-11-21
release: v1.20.0

PascalRepond:


PascalRepond:
## Tests

- [ ] Stats configs list: in the stat configs route, set showFacetsIfNoResults to True so that when opening the cfg list, I can disable the default filter if there are no results

jma:
> * Category and indicator are not translated in config brief and detailed view, **and in the facet**
> * [ ]  Label "Is active" is not translated in the config detailed view

These values seems not be in the translation files nor translated.

PR #3573, opened 2024-01-02
partof: allow strings for volume and issue

### LINKED ISSUE #1818: Bibliographic record : display of the host document's numbering
**Describe the bug**
In the host document's numbering (field `partOf`) :
 - it's impossible to enter a combined numbering (e.g.: number 1/2). 
If the combined numbering is entered, the bibliographic record can be saved, but the value is not taken into account (not displayed and saved).
- If we duplicated the numbering field, in the public interface (detailed view), the repeated numbering goes to the line.

**To Reproduce**

1. Edit a bibliographic record with a host document
2. Try to add a combined numbering
3. See that it's not displayed (and not saved when editing again the record). 
4. Add multiple numbering
5. See that each is displayed on the line.
Example : https://ils.test.rero.ch/professional/records/documents/detail/2000556 and https://ilsdev.test.rero.ch/global/documents/2009631

**Expected behavior**
One on the following : 
- The fields "year", "volume" and "issue" are repeatable. The display is change to add "/" between repeted values.
  OR
- The "year", "volume" and "issue" should be a Text area.
Note : sometimes "numbering" are letters (part A / part B for example).

The numbering should display on the same line as the "Host document" in the detail view of document in the public interface (the pro interface is ok).

**Context**

* server: [ils.test.rero.ch][1] or [ilsdev.test.rero.ch][2]
* version: `v1.1.0` 

[1]: https://ils.test.rero.ch
[2]: https://ilsdev.test.rero.ch

**Screenshots**
image

### LINKED ISSUE #3570: PartOf: some numbering fields should be strings instead of integers
### How it works
In the related documents fields, the subfield numbering requires an integer. However, sometimes it is needed to input more detailed info, for example "Serie 2, Band 2". Accepting something else than integers would be good in those cases.

### Improvement suggestion
In the `partOf` field of the document, the subfields `numbering.volume` and `numbering.issue` should allow string inputs.

PascalRepond:
* Closes #3570.
* Closes #1818.
* ⚠️ On migration, we need to convert the fields
`document.partOf.numbering.volume` and `document.partOf.numbering.issue`
from integers to strings in the db (the mappings stay the same).

### DIFF
--- a/data/documents_small.json
+++ b/data/documents_small.json
@@ -1441,7 +1441,7 @@
         },
         "numbering": [
           {
-            "volume": 81
+            "volume": "81"
           }
         ]
       }
@@ -3470,7 +3470,7 @@
         },
         "numbering": [
           {
-            "volume": 4
+            "volume": "4"
           }
         ]
       }
@@ -7871,7 +7871,7 @@
         },
         "numbering": [
           {
-            "volume": 2
+            "volume": "2"
           }
         ]
       }
@@ -8468,7 +8468,7 @@
         },
         "numbering": [
           {
-            "volume": 16
+            "volume": "16"
           }
         ]
       }
@@ -9207,7 +9207,7 @@
         },
         "numbering": [
           {
-            "volume": 1
+            "volume": "1"
           }
         ]
       }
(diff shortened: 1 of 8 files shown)


rerowep (approved):

jma (approved):
What about the migration?

PR #3579, opened 2024-01-16
fix(circ): item can wrongfully become at desk

### LINKED ISSUE #3403: Circulation: item in transit can wrongfully become at_desk
### Bug description:
In a certain specific case, an item that is in transit and with another pending request can become directly `at_desk` and sends an availability notification when the first request is cancelled. This seems to happen only if the transaction library of both requests is the same. 

Linked to an incomplete fixing of issue https://github.com/rero/rero-ils/issues/2454
May be linked to #3569 

### Expected behavior:
Cancelling a request when the status of the item is `in_transit_for_pickup` or `in_transit_to_house` should never become `at_desk` or send an availability notification. In this case, the item should stay `in_transit_for_pickup` and the notification should only be generated when the item is recieved.

### Steps to Reproduce:

1. (Log in with syslib)
2. **With active library A**, request an item from library B to pickup in library A for **patron1**
3. With active library B, validate the request.
4. **With active library A**, request the same item to pickup in library A for **patron2**
5. **With active library A**, cancel the first request.
6. See that a notification is generated for patron2 and the item status becomes `at_desk` when it should still be `in_transit_for_pickup` because it has not been recieved by library A yet.

**Version:**

`v1.17.0`

### Screenshots
tempsnip

### LINKED ISSUE #3569: Operation log `cancel_request` is not created if the request has already been validated
### Bug description:
If a request is cancelled after being validated, there is no operation log for the cancellation of the request.

### Expected behavior:
A request cancellation should always create an operation log so that I can see when the request has been cancelled in the circulation history.

### Steps to Reproduce:
1. Request an item for pickup at another library
2. Validate the request so that the item becomes `in_transit`
3. Cancel the request
4. See that there is no operation log on this loan for the `cancel_request`

### Context

- Versions: v1.20.0

PascalRepond:
* Fixes incorrect test for action CANCEL_REQUEST_5_1_2.
* Closes #3403.
* Closes #3569.

### DIFF
--- a/rero_ils/modules/items/api/circulation.py
+++ b/rero_ils/modules/items/api/circulation.py
@@ -474,28 +474,32 @@ class ItemCirculation(ItemRecord):
             self.status_update(
                 self, dbcommit=True, reindex=True, forceindex=True)
             actions.update({LoanAction.UPDATE: loan})
-            item = self
         elif actions_to_execute.get('validate_first_pending'):
             pending = self.get_first_loan_by_state(state=LoanState.PENDING)
-            # When the loan to cancel is ITEM_IN_TRANSIT_FOR_PICKUP, we need
-            # virtually to undo the last transaction and restart from the last
-            # location and validate the first pending request.
-            # to achieve this, we pass the cancelled loan pickup_location_pid
-            # as a transaction_location_pid for the validation of the next
-            # pending loan. This way the validation logic is not modified.
-            cancelled_loan_pickup_pid = None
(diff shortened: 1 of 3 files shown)


rero_ils/modules/items/api/circulation.py
```diff
@@ -477,23 +477,30 @@ def cancel_item_request(self, pid, **kwargs):
             item = self
```
rerowep, line 477: Maybe we can move `item = self` before the `if`'s end delete it everywhere else.

jma (approved):

lauren-d (approved):

Issue #3611, opened 2024-02-26
Pro requests list: display temporary location


PascalRepond:
### How it works
In the requests view (https://ils.test.rero.ch/professional/circulation/requests), the "location" field always shows the main location.

### Improvement suggestion
If a temporary location exists in the item, display it under the main location, so that the librarian knows where the item is.

PR #3621, opened 2024-03-04
feat(loan): add temporary location on item

Garfield-fr:
* Closes rero/rero-ils#3611.

### DIFF
--- a/rero_ils/modules/loans/api.py
+++ b/rero_ils/modules/loans/api.py
@@ -261,6 +261,7 @@ class Loan(IlsRecord):
 
         :param library_pid: the library pid.
         """
+        # TODO: Refactor this with a dump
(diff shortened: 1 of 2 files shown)


lauren-d (approved):

PascalRepond (approved):

rero_ils/modules/loans/api.py
```diff
@@ -421,6 +421,13 @@ def item_type_by_pid(pid, known_ittys):
                         item_data['location']['pid'],
                         locations
                     )['name']
+                    if 'temporary_location' in item_data:
```
jma, line 424: Can you add a comment on the top of this method: it should be rewritten.

jma (changes requested):
Tests are missing.

jma (approved):

Issue #2971, opened 2022-06-24
Search: exclusion operator doesn't work as expected for most fields


PascalRepond:
### How it works
Using a `-` operator in simple search should exclude results that contain the term in any text field. As of now it only exclude the result if the specified term is present in the `title` or `subject`. 

### Example

1. Take [this search](https://ils.test.rero.ch/global/search/documents?q=*&page=1&size=10&sort=bestmatch&author=Colin,%20Fabrice,%201972-....) with 4 results: 
3. [This query](https://ils.test.rero.ch/global/search/documents?q=-gallimard&page=1&size=10&sort=bestmatch&author=Colin,%20Fabrice,%201972-....) should exclude the last document containing "Gallimard" in the provisionActivity.
4. I added some fields to [this document](https://ils.test.rero.ch/professional/records/documents/edit/203) to test exclusion behavior of other fields. 

### Improvement suggestion
Adapt ES mappings or templates so that all of the following fields respond to the exclusion operator `-`:
- adminMetadata.note
- note
- originalTitle
- provisionActivity._text.value
- seriesStatement._text
- summary.value

PR #3639, opened 2024-03-26
documents: add files support

jma:
- Uses rero-invenio-files to add file support to documents.
- Pushes full text into the document index.
- Creates a jsonschema store to support local and bib.rero.ch prefix.
- Adds full text boosting field.
- Creates randomly files during the setup.
- Adds documents with files to the 'online' filter.
- Removes the fulltext fields from the default fields search.
- Adds a new url params "&fulltext=1" to add the fulltext fields to the search fields.
- Adds files locations configuration.
- Closes: #2971.


jma:
## PO tests

- [x] REP: If I upload a file with an existing file name in the same bucket, the thumbnail is not displayed correctly.
image
- [x] REP: When uploading a file, a warning should be displayed to the librarian “The files uploaded here will be publicly available. Make sure that you are legally allowed to distribute a file before adding it to a document.”
- [ ] REP: Collection list editor
    - [x] Label should be "Digital collections"
    - [ ] The tooltip is in a different style/font than in the editors, should be the same
- [ ] REP: PRO UI: "My files" tab
    - [x] I would rename the tab label to "Files" and adjust the title inside to "Files (my library)", if possible
    - [ ] [nice-to-have] The files in the list should be unextended by default
    - [ ] The button to extend could be clearer (an arrow on the left, like in the rest of the app instead of the "i")
    - [ ] [nice-to-have] There should be a button at the top to extend all elements at once
    - [ ] [nice-to-have] The thumbnail could be displayed in this component
- [ ] REP: Adding a file with certain characters in file name results in an error. Tried with "Collaborateur_trice bibliothèque (80 %).pdf" in https://ilsdev.test.rero.ch/professional/records/documents/detail/1008 (Aoste Lycée)

PascalRepond:
## PO tests (12.04.2024)

- [ ] Pro UI: component "my files" in the tab "Files" lets me add a file but does not show the list of existing files or the digital collections
- [ ] Changing the label when uploading file seems to change the file name also and the file is not readable if I remove the file extension from the label ([example](https://ilsdev.test.rero.ch/api/records/054b9-x0h16/files/fusil/content) : https://ilsdev.test.rero.ch/professional/records/documents/detail/620
- [ ] The iframe for file preview is quite short. It would be nice if it could take a bit more of the viewport's height (like in SONAR).
- [ ] The MIME-type thumbnails do not work.
image
- [ ] Availability component (brief views): If a document contains a file, maybe the document should display as "available"
- [ ] [nice-to-have] In the pro document detail view, it would be nice that the button "Add" has a third option "a file" that opens the modal to add a file.
- [ ] See comment [above](https://github.com/rero/rero-ils/pull/3639#issuecomment-2020043390) for other unadressed problems.

PascalRepond:
## Tests 08.05

- [ ] Also update the librarian and legacy stats (but not pricing) for "number of documents" to count documents with only files (use document.organisation_pid and library_pid) instead of "holdings" same as for the facets.
    - Specs: "Librarian stats config (+ legacy librarian stats): the number_of_documents indicator count files as belonging to a library if they have files in this library (even if they have no holdings in this library) == facets"

PR #3640, opened 2024-03-26
feat(item): add extra data on item dumper

### LINKED ISSUE #1321: Flash message at the check-in if the item in an active exhibition/course or in a temporary location
**How it works**

Currently, an item in an active exhibition/course or temporary location is check-in/received without message.

**Improvement suggestion**

Scanning (recieve or no action) an item **in its owning library** should trigger a warning flash message if it is linked to an exhibition (regardless of the collection dates) or if it has a temporary location:
* This item is in exhibition/course _collection.name_.
    * If the item has multiple collections linked, display the collection.name of the first one and add "and X other(s)."
* This item is in temporary location _temporary-location.name_.

Garfield-fr:
Add the name of the temporary location as well as
the name of the collections.

* Closes #1321.

### DIFF
--- a/rero_ils/modules/items/dumpers.py
+++ b/rero_ils/modules/items/dumpers.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # RERO ILS
-# Copyright (C) 2019-2023 RERO
+# Copyright (C) 2019-2024 RERO
 # Copyright (C) 2019-2023 UCLouvain
 #
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


rerowep (approved):

rero_ils/modules/items/dumpers.py
```diff
@@ -144,4 +145,17 @@ def dump(self, record, data):
...
             ]
+        # add temporary location name
+        if temporary_location_pid := item.get('temporary_location', {}).get(
+            'pid'
+        ):
+            data['temporary_location']['name'] = Location.get_record_by_pid(
+                temporary_location_pid).get('name')
+        # add collections
+        query = CollectionsSearch().active_by_item_pid(item['pid'])
+        if query.count() > 0:
```
jma, line 156: This is not optimal as it requires two es query. You can do:

```python
query = query.params(preserve_order=True).source('title').scan()
if collections := [collection.title for collection in collections]:
    data['collections'] = collections
```

jma (approved):

Issue #3600, opened 2024-02-15
Monitoring: standard usage warnings should be ignored in Sentry


PascalRepond:
### How it works
Some warnings that are not system errors but just standard usage problems (linked to wrong user input in editors for example) are displayed in Sentry but they are not useful.

### Improvement suggestion
At least the following types of errors should not show up in Sentry or should be easily ignorable:

- AcqReceiptLine err:La quantité reçue est supérieure à la quantité commandée …
- AcqReceipt err:'df' is too short
- Document err:'periodical' is not one of ['article', 'materialUnit', 'privateFile', 'privateSubfile']
- Document err:'XXXX' is a required property
- Holdings: list index out of range (prédiction)
- Holding err:Les fréquences régulières doivent avoir la prochaine date prévue renseignée.
- Item err:Barcode XXXXX is already taken
- Item err:La catégorie de prêt temporaire ne peut pas être identique à la catégorie de prêt par défaut
- LocalField err:'fields' is a required property
- Patron err:'barcode' is a required property
- Patron err:At least one email should be defined for an email communication channel. …

PR #3644, opened 2024-04-03
monitoring: better standard usage warnings

rerowep:
* Standard usage warnings should be ignored in Sentry.
* Close: #3600


Garfield-fr (approved):

rero_ils/modules/api.py
```diff
@@ -252,8 +252,8 @@ def create(cls, data, id_=None, delete_pid=False,
         # TODO: remove validation error once the angular editor
         #       validation is complete
```
PascalRepond, line 253: Should we remove this comment, now?

PascalRepond (approved):

PR #3657, opened 2024-04-25
fix(circulation history): number of operations

PascalRepond:
* Fix the circulation stats on item to not take into account operations of type notification when calculating number of operations (checkout, renwal, etc.).


rerowep (approved):

Garfield-fr (commented):
The tests must be fixed

PR #3659, opened 2024-04-29
fix(operation_log): circulation operation count

### LINKED ISSUE #3656: Item circulation history: wrong operation counts
### Bug description:
In the item circulation log, the numbers for "number of checkouts" and "number of renewals" are incorrect. Because the item stats count the operation logs without filtering out the operation_logs of type "notification".

### Expected behavior:
The item circulation stats should only count circulation operations and not notifications ! 

### Steps to Reproduce:

1. Go to https://bib.rero.ch/professional/records/items/detail/3488715 (connected as RBNJ)
2. Click on "circulation history"
3. See the "number of checkouts" does not correspond to the number of operation in the list
4. It is not due to the legacy checkout count because these items were created after the migration and have 0 legacy checkouts.

### Context

- Versions: v1.21.0 (bug is present since we implemented the notifications in the circulation history)

Garfield-fr:
To obtain the correct count, operation logs must be filtered using loan records only.

* Closes #3656.

### DIFF
--- a/tests/api/items/test_items_rest_views.py
+++ b/tests/api/items/test_items_rest_views.py
@@ -195,3 +195,16 @@ def test_item_stats(
             {
                 'total': {'checkin': 2},
                 'total_year': {'checkin': 1}}
+
+
+def test_item_stats_endpoint(
+    item_at_desk_martigny_patron_and_loan_at_desk,
+    client, librarian_martigny
+):
+    """Test loan filter on stats endpoint with real data."""
+    login_user_via_session(client, librarian_martigny.user)
+    res = client.get(url_for(
+        'api_item.stats',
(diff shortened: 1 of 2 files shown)


rerowep (approved):

tests/api/items/test_items_rest_views.py
```diff
@@ -195,3 +195,16 @@ def test_item_stats(
             {
                 'total': {'checkin': 2},
                 'total_year': {'checkin': 1}}
+
+
+def test_item_stats_endpoint(
```
PascalRepond, line 200: Are we sure this test is correct? Does the fixture "item_at_desk_martigny_patron_and_loan_at_desk" contains a notification operation log that is ignored.

PascalRepond (approved):

jma (approved):

PR #3670, opened 2024-05-07
dependencies: fix babel

rerowep:
* Fixes dependencies.
* Solves babel test fails. No `,` after day names in version 2.15


Garfield-fr (approved):

PascalRepond (approved):

lauren-d (approved):

pyproject.toml
```diff
@@ -117,7 +117,6 @@ pyparsing = "^3.1.1"
 flask-wiki = "^0.3.1"
 dnspython = ">2.4.2"
 cryptography = ">41.0.7"
-babel = "<2.15.0"  # NOTE: remove when https://github.com/python-babel/babel/issues/1081 is solved
```
jma, line 120: babel < 2.15.0 will broke the test, thus we need to add a constraint to force babel > 2.14.0

jma (approved):

pyproject.toml
rerowep: This PR will correct the babel tests.

PascalRepond: Not for babel 2.14.0 and lower, that is why to be completely correct with the tests, we should add `babel = ">2.14.0"`

PR #3671, opened 2024-05-07
fix(items): remove legacy fields

### LINKED ISSUE #3667: Delete legacy fields on items
### How it works

For migration, we used `legacy_checkout_count` and `legacy_circulation_rules` fields as a safety for migrating libraries, so that they could easily see some info about recent circulation stats. 

Now that RERO ILS has been in production for three years, this data is no longer needed or relevant, we can get rid of it completely, which will simplify the code and lighten the load.

Prior Virtua clients can see this data in the Virtua dumps that they recieved, if they really need it.

### Improvement suggestion

- Delete the fields `legacy_checkout_count` and `legacy_circulation_rules` in the item
- Get rid of code that used these fields:
    - Fixtures generation
    - Item stats for display in the UI
    - Serializers and tests
    - Other?

PascalRepond:
* Closes #3667.
* :warning: Requires update mapping and deleting fields `legacy_checkout_count` and `legacy_circulation_rules` in existing items' data.


Garfield-fr:
There are other places in the code where Legacy is used. For example, in cli. You need to search the `legacy` code ?

jma (approved):

PR #3681, opened 2024-05-16
release: v1.22.0

PascalRepond:


PascalRepond:
## RC tests (PO + clients)

- [ ] REP: **Item inherited call numbers**: When updating a serial holdings call numbers, related existing items are not reindexed and so they don't get the `inherited_(first and second)_call_number` and can't be searched by this field. 
    - @rerowep Maybe expand the function `update_items_locations_and_types` in holdings listener to take into account the call numbers?
    - Also this function seems to use bulk index, make sure it still works since we realized bulk index is disabled?

rerowep (approved):

Garfield-fr (approved):

PR #3685, opened 2024-05-23
fix(holdings): items inherit call numbers 

PascalRepond:
* Serial items should correctly inherit their inherited_call_numbers
in the index after editing the parent serial holdings.
* Refactored the `update_items_locations_and_types` function.
* Fixed incorrect tests for item-issue fields inheritance.

### DIFF
--- a/rero_ils/modules/holdings/listener.py
+++ b/rero_ils/modules/holdings/listener.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # RERO ILS
-# Copyright (C) 2019-2022 RERO
+# Copyright (C) 2019-2024 RERO
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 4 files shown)


rero_ils/modules/holdings/listener.py
```diff
@@ -75,48 +76,56 @@ def update_items_locations_and_types(sender, record=None, **kwargs):
...
                     if item_temp_loc_pid == hold_loc_pid:
                         item.pop('temporary_location', None)
                     item['location'] = {'$ref': get_ref_for_pid(
                         'locations', hold_loc_pid)}
+
                 # remove the item temporary_item_type if it is equal to the
                 # new item item_type.
-                temporary_type = item.get('temporary_item_type', {})
-                if temporary_type:
+                if temporary_type := item.get('temporary_item_type', {}):
```
jma, line 128: Default value is not required.

jma (approved):

rerowep (approved):

PR #3687, opened 2024-05-23
fix(babelio): fix tab name

PascalRepond:


Garfield-fr:
Change the commit name with `fix(babelio): fix tab name` (No reference to the customer)

rerowep (approved):

PR #3694, opened 2024-06-03
chore: remove version on docker files

Garfield-fr:
* Removes version on docker file (obsolete)
* Adds data/files path on gitignore.


Garfield-fr:
Info: https://docs.docker.com/compose/compose-file/04-version-and-name/

PascalRepond (approved):

benerken (approved):

PR #3717, opened 2024-07-09
fix(jsonschemas): fix minItems

### LINKED ISSUE #3709: Order line editor: cannot add a note
### Bug description:
When editing an existing order line, adding a "Note" field does nothing, there is no way to add this field except recreating the order line.

### Expected behavior:
- I should be able to add a note field

### Steps to Reproduce:
1. Go to an order and edit an existing order line
2. See issue

### Context

- Versions: 1.22.0

PascalRepond:
* minItems: 0 prevents some fields to be displayed in edit mode, especially the notes field in acquisition receipts and orders.
* Closes #3709.


Garfield-fr (approved):

rerowep (approved):

ManaDeweerdt:
- [ ] : cannot add the note field on an existing order line.

PR #3739, opened 2024-08-28
sip2: fix format_patron_address

### LINKED ISSUE #3727: SIP2 returns None for a user/patron with any missing address field
### Bug description:
If a user is missing an address field (street, postal_code, or city), there is an error in the function `format_patron_address(patron)` when trying to strip those fields. This prevents the user data to be returned correctly for SIP2.

### Expected behavior:

- Allow patron info to be returned correctly even if the patron has no address fields.

### Steps to Reproduce:

1. Create a patron without an address
2. Try to get patron_information from SIP2
3. See that it returns None

rerowep:
* Closes #3727.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -995,17 +995,17 @@ files = [
 
 [[package]]
 name = "deepdiff"
-version = "7.0.1"
+version = "8.0.1"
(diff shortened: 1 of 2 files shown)


jma (approved):
Why did you update the dependencies?

Issue #3745, opened 2024-09-03
Allow more than two barcodes per patron


PascalRepond:
### How it works
The list for patron.patron.barcode is currently limited to 2 items. This causes problems when loading new users from files. Sometimes they have already 2 barcodes and we cannot add a third.

### Improvement suggestion
Remove "maxItems: 2" in `patron.patron.barcode` or extend the limit to 5 barcodes.

PR #3746, opened 2024-09-03
fix(patron barcodes): extend the number of barcodes

PascalRepond:
* Closes https://github.com/rero/rero-ils/issues/3745.


jma (approved):
This should be well tested.

PR #3751, opened 2024-09-04
import: fix SLSP

rerowep:
* Fixes error for SLSP imports with incorrect 008.

### DIFF
--- a/rero_ils/modules/documents/dojson/contrib/marc21tojson/slsp/model.py
+++ b/rero_ils/modules/documents/dojson/contrib/marc21tojson/slsp/model.py
@@ -102,10 +102,11 @@ def marc21_to_language(self, key, value):
     language = do_language(self, marc21)
     # is fiction
     self["fiction_statement"] = DocumentFictionType.Unspecified.value
-    if value[33] in ["1", "d", "f", "j", "p"]:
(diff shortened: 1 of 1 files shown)


rero_ils/modules/documents/dojson/contrib/marc21tojson/slsp/model.py
```diff
@@ -102,10 +102,11 @@ def marc21_to_language(self, key, value):
     language = do_language(self, marc21)
     # is fiction
     self["fiction_statement"] = DocumentFictionType.Unspecified.value
-    if value[33] in ["1", "d", "f", "j", "p"]:
-        self["fiction_statement"] = DocumentFictionType.Fiction.value
-    elif value[33] in ["0", "e", "h", "i", "s"]:
-        self["fiction_statement"] = DocumentFictionType.NonFiction.value
+    if len(value) >= 33:
```
PascalRepond, line 105: Should be ">33" because if `len(value) == 33`, `value[33]` will be an error

PascalRepond (changes requested):

PascalRepond (approved):

jma (approved):
Tests are missing.

Issue #2734, opened 2022-03-01
Add user info to sentry errors


jma:
### How it works
Actually it is not easy or not possible to retrieve the connected user for a sentry error.

### Improvement suggestion
Configure sentry on our invenio instance to add the logged user information.

PR #3765, opened 2024-10-23
logging: add user information to sentry

rerowep:
* Closes: #2734.

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -3458,6 +3458,11 @@ RERO_ILS_STATS_BILLING_TIMEFRAME_IN_MONTHS = 3
 LOGGING_SENTRY_LEVEL = "ERROR"
 #: Sentry: use celery or not
 LOGGING_SENTRY_CELERY = True
+# Invenio logging
+LOGGING_SENTRY_INIT_KWARGS = {
+    # instuct sentry to send user data attached to the event
+    "send_default_pii": True
+}
 
 ROLLOVER_LOGGING_CONFIG = {
     "version": 1,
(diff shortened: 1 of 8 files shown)


rero_ils/config.py
```diff
@@ -3455,6 +3455,11 @@ def _(x):
 #: Sentry: use celery or not
 LOGGING_SENTRY_CELERY = True
 
+LOGGING_SENTRY_INIT_KWARGS = {
+    # Instruct Sentry to send user data attached to the event
+    "send_default_pii": True,
```
PascalRepond, line 3460: pii?

rero_ils/modules/utils.py
```diff
@@ -1091,6 +1092,7 @@ def set_user_name(sender, user):
 
     user_name = None
     remove_user_name(sender, user)
+    set_user_data = {"id": user.id, "email": user.email}
```
PascalRepond, line 1095: User email is not useful and might be a security concern to put in the logs.

PascalRepond: More useful information to debug and reproduce would be : `patron.pid` `patron.patron.barcode[0]` and the patron's organisation.

rero_ils/config.py
rerowep: See sentry sdk

PascalRepond (approved):

rero_ils/modules/utils.py
```diff
@@ -1085,22 +1084,35 @@ def close(self):
         self.__del__()
 
 
+def user_formatted_name(user):
```
jma, line 1087: You can use the Patron.formatted_name method.

jma (approved):

PR #3775, opened 2024-11-11
general: fix cached timeout

rerowep:


Garfield-fr (approved):

PascalRepond (approved):

rero_ils/modules/cli/utils.py
```diff
@@ -273,7 +273,7 @@ def add_org_lib_doc(item, doc_pid="dummy"):
     return item
 
 
-@cached()
+@cached(timeout=2 * 60 * 60, query_string=True)  # 2 hour timeout
```
jma, line 276: You need to use key_prefix for non route function and meomize (need to change get_document_types_from_schema ) for class methods: https://flask-caching.readthedocs.io/en/latest/#caching-other-functions. Maybe we need to use flask-caching directely.

jma (changes requested):

jma (approved):

PR #3777, opened 2024-11-12
fix(documents): delete orphan harvested documents

### LINKED ISSUE #3776: Periodically delete orphan harvested documents
### How it works

Bibliomedia batches are now `harvested:true`, but it prevents the document manual deletion when these batches are retracted and the item is returned.

### Improvement suggestion

Add a periodic task that automatically deletes documents when they are `harvested:true` and are not linked to any other resource.
Make sure that it does not delete ebooks; it shouldn't as they have holdings but no items.

PascalRepond:
* Closes #3776.

### DIFF
--- a/rero_ils/modules/documents/tasks.py
+++ b/rero_ils/modules/documents/tasks.py
@@ -37,6 +37,46 @@ def reindex_document(pid):
 
 
 @shared_task(ignore_result=True)
+def delete_orphan_harvested(delete=False, verbose=False):
+    """Delete orphan harvested documents.
+
+    :param delete: if True delete from DB and ES.
+    :param verbose: Verbose print.
+    :returns: count of deleted documents.
+    """
+    query = (
+        DocumentsSearch()
+        .filter("term", harvested=True)
+        .exclude("exists", field="holdings")
+    )
+    pids = [hit.pid for hit in query.source("pid").scan()]
+    count = 0
+
+    if verbose:
+        click.secho(f"Orphan harvested documents count: {len(pids)}", fg="yellow")
+    for pid in pids:
+        if doc := Document.get_record_by_pid(pid):
+            if verbose:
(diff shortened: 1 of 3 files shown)


rero_ils/modules/documents/tasks.py
```diff
@@ -36,6 +36,43 @@ def reindex_document(pid):
...
+    if verbose:
+        click.secho(f"Orphan harvested documents count: {count}", fg="yellow")
+    for pid in pids:
+        if doc := Document.get_record_by_pid(pid):
+            if verbose:
+                click.secho(f"Deleting orphan harvested: {pid}", fg="yellow")
+            if delete:
+                try:
+                    if doc.reasons_not_to_delete() == {"others": {"harvested": True}}:
+                        doc.delete(dbcommit=True, delindex=True, force=True)
```
rerowep, line 64: Not sure `force=True` will do the trick. Force will delete the persistent identifier in DB. I think yo have to set `harvested` to `False` before deleting.

rerowep (changes requested):

rero_ils/modules/documents/tasks.py
```diff
@@ -36,6 +36,46 @@ def reindex_document(pid):
...
+                    # only delete documents that have no links to me, only reason not to delete should be 'harvested'
+                    if doc.reasons_not_to_delete() == {"others": {"harvested": True}}:
+                        doc.pop("harvested")
+                        doc.replace(doc, dbcommit=True, reindex=True)
+                        doc.delete(dbcommit=True, delindex=True)
+                        count += 1
+                except Exception:
+                    msg = f"COULD NOT DELETE ORPHAN HARVESTED: {pid} {doc.reasons_not_to_delete()}"
+                    if verbose:
+                        click.secho(f"ERROR: {msg}", fg="red")
```
jma, line 72: same

jma (approved):

PR #3808, opened 2025-01-06
files: change template and max_files_count

rerowep:
* Adds `files-*` to ES record template.
* Changes `max_files_count` to 2100.

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -4396,3 +4396,5 @@ RERO_FILES_RECORD_SERVICE_CONFIG = "rero_ils.modules.files.services.RecordServic
 RERO_FILES_RECORD_FILE_SERVICE_CONFIG = (
     "rero_ils.modules.files.services.RecordFileServiceConfig"
 )
+RERO_ILS_APP_FILES_UI_MAX = 600
+RERO_ILS_APP_FILES_FULL_TEXT_MAX = 10 * 1024 * 1024  # 10Mb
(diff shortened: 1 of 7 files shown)


PascalRepond (approved):

rero_ils/config.py
```diff
@@ -4396,3 +4396,6 @@ def search_type(field):
 RERO_FILES_RECORD_FILE_SERVICE_CONFIG = (
     "rero_ils.modules.files.services.RecordFileServiceConfig"
 )
+RERO_ILS_APP_FILES_UI_MAX = 600
```
jma, line 4399: You need to expose this here: https://github.com/rero/rero-ils/blob/977d293e3352ce3ce1f2a888543b88e36e7a3bde/rero_ils/modules/patrons/views.py#L131

rero_ils/config.py
```diff
@@ -4396,3 +4396,6 @@ def search_type(field):
...
+RERO_ILS_APP_FILES_MAX = RERO_ILS_APP_FILES_UI_MAX * 3 + 100
```
jma, line 4400: If you want to make this variable value in function of the UI variable you need to put it into the max_files_count class method.

rero_ils/es_templates/v7/record.json
```diff
@@ -10,6 +10,7 @@
     "circ_policies-*",
     "collections-*",
     "documents-*",
+    "files-*",
```
jma, line 13: Thanks.

jma (changes requested):

rero_ils/config.py
rerowep: Added: `"maxFiles": config.get("RERO_ILS_APP_FILES_UI_MAX", 500),`

jma (approved):

PR #3814, opened 2025-01-15
acquisitions-order-line: Add only on pending

rerowep:
* Fixes add of `acquisitions-order-line` only to pending `acquisitions`.


PascalRepond:
@jma It would be nice to also reflect this in the UI.

PascalRepond (approved):

jma (approved):

PR #3819, opened 2025-02-04
fix(stats): make active patrons period 12 months

PascalRepond:
- The correct date range for number_of_active_patrons in the pricing stats should be 12 months and not the default `RERO_ILS_STATS_BILLING_TIMEFRAME_IN_MONTHS`.

### DIFF
--- a/rero_ils/modules/stats/api/pricing.py
+++ b/rero_ils/modules/stats/api/pricing.py
@@ -58,6 +58,7 @@ class StatsForPricing:
             fmt="YYYY-MM-DDT00:00:00"
         )
         _to = to_date.format(fmt="YYYY-MM-DDT23:59:59")
+        self.to_date = to_date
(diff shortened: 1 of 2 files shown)


rero_ils/modules/stats/api/pricing.py
```diff
@@ -171,11 +171,16 @@ def number_of_active_patrons(self, library_pid):
         :return: the number of matched active patrons
         :rtype: integer
         """
+        _to = self.to_date.format(fmt="YYYY-MM-DDT23:59:59")
```
jma, line 174: Can you add a comment to explain why we do not take the default date_range. And better change the variable name.

jma (approved):
It would be nice to write a test.

rerowep (approved):

jma (approved):

PR #3843, opened 2025-04-08
fix(entities): wrong date in public view

PascalRepond:
* Entities' dates are string fields, they should not be formatted by the interface, otherwise they display wrong data. E.g. if "date of birth" is only a year (very frequent), the formatting will add a day/month.


Garfield-fr (approved):
Don't forget to transfer this to the next-version branch.

PR #3877, opened 2025-06-30
import: DNB, SLSP subject mef link

rerowep:
* Adds 6xx $2 gnd-content to subjects, genre form creation.

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -3913,11 +3913,6 @@ RERO_IMPORT_REST_ENDPOINTS = dict(
     ),
 )
 
-# RERO_ILS_IMPORT_6XX_TARGET_ATTRIBUTE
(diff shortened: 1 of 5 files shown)


Garfield-fr (approved):

PascalRepond:
## Tests

- [ ] REP: Sometimes importing from SLSP, entities are imported double because SLSP contains both GND entity and IdRef entity, but they are linked in MEF. [image](https://github.com/user-attachments/assets/1e04872a-4eb4-493d-a99c-6a36cd22c741)
- [ ] REP: Example 9781035701353. "Sachbilderbuch" is in zone 655 (genre form) but it is imported in "subjects" because it has a GND identifier. It should be imported in genre-form as textual entity with GND identifier. *The group metadata should decide next if genre-form field should accept MEF links that are not BNF-GenreForm*

PR #3893, opened 2025-07-09
ill requests: Fix populate document

rerowep:
* Fixes populate document data in the form if there is no publication start date.


PascalRepond (approved):
Commit message form: `fix(ill): populate document data`

PR #3894, opened 2025-07-09
fix(documents): set the provision activity date

rerowep:
* Sets provision activity undefined date to 9999


PascalRepond (approved):

jma (approved):

jma (commented):
I would change the commit message to: fix(documents): set the provision activity default  date value to 9999

PR #3895, opened 2025-07-09
fix(holdings): inherit issues' reasons not to delete

### LINKED ISSUE #3651: Error when deleting a holdings that still has items with fees
### Bug description:

In the document AND holdings detailed view for serial holdings, there is a button to delete the holdings which also deletes linked items (in cascade) when possible. This button is supposed to be inactive if something prevents the deletion (linked item with active loan, for example). 

If the holdings only contains items with a fee as "reason not to delete", the holdings "Delete" button is active and clicking it leads to a system error. 

### Expected behavior:

In this case, the button should be grayed out, same as if an item in the holdings had an active loan. The tooltip should give the reason(s) not to delete. Example: "has 2 items with fees attached"

image

### Steps to Reproduce:
⚠️ Complex to reproduce in test without data manipulation or adding fixtures since it requires a holdings to contain an item with a linked fee. 

- Find a holdings that contains only items without active loans but with pending fees
- Try to delete the holdings (either from the document detailed view or holdings detailed view)

Example in production RBNJ (18.04.2024):
https://bib.rero.ch/professional/records/holdings/detail/1559437
image

### Context

- Versions: v1.21.0

rerowep:
* Closes #3651.

### DIFF
--- a/rero_ils/modules/holdings/api.py
+++ b/rero_ils/modules/holdings/api.py
@@ -163,7 +163,11 @@ class Holding(IlsRecord):
         document_pid = extracted_data_from_ref(self.get("document").get("$ref"))
         document = Document.get_record_by_pid(document_pid)
         if not document:
-            return _(f"Document does not exist {document_pid}.")
+            return _(
+                "Document does not exist {document_pid}.".format(
+                    document_pid=document_pid
+                )
+            )
 
         if self.is_serial:
             patterns = self.get("patterns", {})
@@ -192,7 +196,11 @@ class Holding(IlsRecord):
             ]
             for field in fields:
                 if self.get(field):
-                    return _(f"{field} is allowed only for serial holdings")
+                    return _(
+                        "{field} is allowed only for serial holdings".format(
+                            field=field
+                        )
+                    )
         # No multiple notes with same type
         note_types = [note.get("type") for note in self.get("notes", [])]
         if len(note_types) != len(set(note_types)):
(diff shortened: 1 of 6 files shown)


rero_ils/modules/holdings/api.py
```diff
@@ -472,15 +457,26 @@ def reasons_not_to_delete(self):
...
+                for reason in reasons_not_to_delete:
+                    counts.setdefault(reason, 0)
+                    if (
+                        reason == "loans"
+                        and issue_status == ItemIssueStatus.RECEIVED
+                        or reason != "loans"
+                    ):
+                        counts[reason] += 1
+                if cannot_delete_msgs := {
+                    f"has {value} items with {name} attached": value
```
PascalRepond: extract and translate:
```_(f"has {value} items with {name} attached"): value```

**Review by PascalRepond — CHANGES_REQUESTED** [review]

**Comment by PascalRepond** [comment]
Commit message: 
```markdown
fix(holdings): inherit issues' reasons not to delete

* Closes #3651.
```

rero_ils/modules/holdings/api.py
```diff
@@ -472,15 +457,26 @@ def reasons_not_to_delete(self):
...
-            if not_deleteable_items:
-                count = len(not_deleteable_items)
-                cannot_delete["others"] = {
-                    _(f"has {count} items with loan attached"): count
-                }
+            counts = {}
+            for item in self.get_all_items():
+                reasons_not_to_delete = item.reasons_not_to_delete().get("links", {})
+                # local_fields isn't a reason to block holding suppression
+                reasons_not_to_delete.pop("local_fields", None)
```
PascalRepond: This is not needed as is already done in `item.reasons_not_to_delete()`

rero_ils/modules/holdings/api.py
```diff
@@ -472,15 +465,30 @@ def reasons_not_to_delete(self):
...
+                issue_status = item.issue_status
+                for reason in reasons_not_to_delete:
+                    counts.setdefault(reason, 0)
+                    if (
+                        reason == "loans"
+                        and issue_status == ItemIssueStatus.RECEIVED
+                        or reason != "loans"
```
PascalRepond, line 478: `reason == "loans" or reason != "loans"` not needed

rero_ils/modules/holdings/api.py
```diff
@@ -472,15 +465,29 @@ def reasons_not_to_delete(self):
...
+                reasons_not_to_delete = item.reasons_not_to_delete().get("links", {})
+                issue_status = item.issue_status
+                for reason in reasons_not_to_delete:
+                    counts.setdefault(reason, 0)
+                    # Add reason count for received loans and all other reasons.
+                    if (
+                        reason == "loans"
+                        and issue_status == ItemIssueStatus.RECEIVED
+                        or reason != "loans"
+                    ):
```
PascalRepond, line 479: Still pretty sure that this whole `if`-statement is unecessary. It excludes nothing. 🤔

PascalRepond (approved):

PR #3897, opened 2025-07-09
invenio: new version

rerowep:
* Updates dependencies.
* Removes `invenio-admin`.
* Uses python 3.12.
* Changes scheduler backend from REDIS to DB.


PascalRepond (approved):

jma (approved):
Remove in your commit message: Remove invenio-admin. This is not the case anymore.

PR #3901, opened 2025-07-10
fix(translations): remove f-strings from gettext

PascalRepond:
"Python f-strings cannot be used directly with gettext functions because f-string expressions are evaluated before they reach gettext." See: https://docs.astral.sh/ruff/rules/f-string-in-get-text-func-call/


PascalRepond:
Duplicate of : https://github.com/rero/rero-ils/pull/3895

PR #3927, opened 2025-08-04
fix(patron editor): make the `country` field bigger

PascalRepond:


jma (approved):
The commit message is not clear to me.

PR #3931, opened 2025-08-20
fix(cipo): sort the reminders by delay

### LINKED ISSUE #3930: The circulation policy reminder is sometimes not correct
### Bug description:

If a CIPO defines a list of reminders that are not in the correct order in the data (ascending for "overdue" and descending for "due_soon"). The notifications are sent with incorrect parameters because it uses the order of the reminders in the data to know which reminder to use.

### Expected behavior:

When sending a circulation reminder-notification, choosing which reminder params to use should be done by ordering the reminders by "delay" (according to the notification type). The order of these reminders in the data is sometimes not coherent. 

### Steps to Reproduce (very convoluted to reproduce)

1. Create a cipo with two "overdue" reminders in this order:

```json
{
        "communication_channel": "mail",
        "days_delay": 49,
        "template": "email/overdue",
        "type": "overdue"
      },
      {
        "communication_channel": "patron_setting",
        "days_delay": 1,
        "template": "email/overdue",
        "type": "overdue"
      }
```
2. This should send a notification by "patron_setting" one day after the due date and one by "mail" 49 days after the dure date.
3. Checkout an loan with this CIPO and wait one day after the due date.
4. See that the notification is sent by "mail"

PascalRepond:
* Closes #3930.
* In order for the communication channel of a notification to always be correct, the reminders in the cipo should always be returned sorted by their "days_delay" subfield.

### DIFF
--- a/rero_ils/modules/circ_policies/api.py
+++ b/rero_ils/modules/circ_policies/api.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # RERO ILS
-# Copyright (C) 2019-2022 RERO
+# Copyright (C) 2019-2025 RERO
 # Copyright (C) 2019-2022 UCLouvain
 #
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


rerowep (approved):

rero_ils/modules/circ_policies/api.py
```diff
@@ -356,15 +356,31 @@ def get_overdue_intervals(self):
         return intervals
 
     def get_reminders(self, reminder_type=DUE_SOON_REMINDER_TYPE, limit=None):
-        """Get reminders corresponding to arguments.
+        """Get reminders and order them by 'days_delay'.
```
jma: This function can be optimized in the sense that sorting reminders is not required for a type != OVERDUE

jma (approved):

rero_ils/modules/circ_policies/api.py
PascalRepond: Good point, I optimised it.

PR #3933, opened 2025-08-26
fix(files): update nginx config on docker

PascalRepond:
* Adds types config for nginx to avoid blocking of PDF previews.


Garfield-fr (approved):

jma (approved):
Be careful we need to do this in production as the nginx configuration used is not the file provided by this repo.

PR #3934, opened 2025-08-28
chore(import): change swisscovery URL to the new one

PascalRepond:
* Also changes the new url on the client frontpages when applicable.


jma (approved):
Lets see if the swissconvery URL should be a configuration value.

PascalRepond:
> Lets see if the swissconvery URL should be a configuration value.

I don't think it is necessary as it usually shouldn't change.

rerowep (approved):

PR #3944, opened 2025-09-16
fix(poe): correct server command for parameters

rerowep:


pyproject.toml
```diff
@@ -90,7 +90,7 @@ dependencies = [
     "babel (>2.16.0)",
     "levenshtein (>=0.26.0)",
     "unidecode (>=1.3.8)",
-    "poethepoet",
+    "poethepoet (< 0.37.0)",
```
PascalRepond: Why add this constraint?

PascalRepond (changes requested):
Can you also add "RUF059" to `tool.ruff.lint.ignore` so that test can pass with new ruff version?

jma (approved):

PascalRepond (approved):

PR #3945, opened 2025-09-17
fix(import): update UGent import url

PascalRepond:


rerowep (approved):

jma (changes requested):
Tests are missing for the entity import.

PascalRepond:
> Tests are missing for the entity import.

@jma What tests ? This is currently not tested in any of these imports, it seems out of the scope of this fix to me.

PR #3954, opened 2025-09-30
items: delete even if in collection

### LINKED ISSUE #3911: Remove an item from linked collections automatically on delete
### Problem description

If an item is referenced in a `collection` resource, it cannot be deleted. This means that the user needs to edit the related collection, to remove the item they want to delete.

### Missing feature

References in collections should not be a reason not to delete items (same as local fields). When an item is deleted, links to this item should be automatically removed from all collection resources that contain it.

rerowep:
* Deletes item from collection during item deletion.
* Closes #3911

### DIFF
--- a/rero_ils/modules/items/api/api.py
+++ b/rero_ils/modules/items/api/api.py
@@ -104,6 +104,13 @@ class Item(ItemCirculation, ItemIssue):
         },
     }
 
+    def delete(self, force=False, dbcommit=False, delindex=False):
+        """Delete record."""
+        from rero_ils.modules.items.tasks import delete_item_in_collections
+
+        delete_item_in_collections.delay(self.pid, dbcommit=True, reindex=True)
+        return super().delete(force, dbcommit, delindex)
+
     def delete_from_index(self):
         """Delete record from index."""
         with contextlib.suppress(NotFoundError, ValueError):
--- a/rero_ils/modules/items/tasks.py
+++ b/rero_ils/modules/items/tasks.py
@@ -22,6 +22,7 @@ from celery import shared_task
 from flask import current_app
 
 from rero_ils.modules.api import IlsRecordError
+from rero_ils.modules.collections.api import Collection, CollectionsSearch
 from rero_ils.modules.holdings.api import Holding
 from rero_ils.modules.holdings.utils import create_next_late_expected_issues
 from rero_ils.modules.utils import extracted_data_from_ref, set_timestamp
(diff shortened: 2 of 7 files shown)


PascalRepond (changes requested):
Test in local, doesn't work. Trying to display the collection after deleting the item gets an error 500

rero_ils/modules/items/api/api.py
```diff
@@ -104,6 +104,14 @@ class Item(ItemCirculation, ItemIssue):
         },
     }
 
+    def delete(self, force=False, dbcommit=False, delindex=False):
+        """Delete record."""
+        from rero_ils.modules.items.tasks import delete_item_in_collections
+
+        delete_item_in_collections.delay(self.pid, dbcommit=True, reindex=True)
+        super().delete(force, dbcommit, delindex)
```
jma: why not return super?

rero_ils/modules/items/tasks.py
```diff
@@ -121,3 +122,19 @@ def delete_holding(holding_pid, force=False, dbcommit=True, delindex=True):
...
             current_app.logger.warning(f"Holding not deleted: {holding_pid}")
+
+
+@shared_task
+def delete_item_in_collections(item_pid, dbcommit=True, reindex=True):
+    """Delete item in collections."""
+    query_collections = CollectionsSearch().filter("term", items__pid=item_pid)
+
+    for hit in query_collections.source("pid").scan():
+        if coll := Collection.get_record(hit.meta.id):
```
jma: If not maybe it can be a good idea to add an error log?

rero_ils/modules/items/tasks.py
```diff
@@ -121,3 +122,19 @@ def delete_holding(holding_pid, force=False, dbcommit=True, delindex=True):
...
+            new_items = []
+            for item in coll["items"]:
```
jma: This can be a comprehension list or the python filter function.

tests/api/collections/test_collections_api.py
```diff
@@ -21,3 +21,6 @@
 def test_get_items(item_lib_martigny, item2_lib_martigny, coll_martigny_1):
     """Test get items for a collection"""
     assert coll_martigny_1.get_items() == [item_lib_martigny, item2_lib_martigny]
+
```
jma, line 25: Please, can you add a comment?

jma (approved):

jma (approved):

PascalRepond (approved):

PR #3958, opened 2025-10-06
scheduler: disable tasks for development

rerowep:


jma (approved):
You can add a message to print the command to enable task if it is needed.

PR #3959, opened 2025-10-07
fix(translations): disable interpolation for translated strings

### LINKED ISSUE #3920: Formatted string not translated in the interface
### How it works

For certain messages, the backend sends a formatted string through the API. This formatted string is extracted but cannot be translated by the frontend because it is sent with the dynamic values.

Image

### Improvement suggestion

In such cases, the API should return a data structure (e.g. dict) and this would then be used by the frontend to display the message and translate it.

PascalRepond:
- The backend should never translate a formatted string with variable interpolation, because it has no notion of language and the frontend will not be able to translate the string with dynamic values.
- Closes #3920.

### DIFF
--- a/rero_ils/modules/acquisition/acq_order_lines/api.py
+++ b/rero_ils/modules/acquisition/acq_order_lines/api.py
@@ -86,7 +86,7 @@ class AcqOrderLine(AcquisitionIlsRecord):
         # NOTES fields testing
         note_types = [note.get("type") for note in self.get("notes", [])]
         if len(note_types) != len(set(note_types)):
-            return _("Can not have multiple notes of the same type.")
+            return _("Cannot have multiple notes of the same type.")
         from rero_ils.modules.acquisition.acq_orders.api import AcqOrder
         from rero_ils.modules.acquisition.acq_orders.models import AcqOrderStatus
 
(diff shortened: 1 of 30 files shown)


Garfield-fr:
I'm really not a fan of putting all messages in “other,” because every time you add a message, you have to change both projects. We really need to find a solution so that the backend sends the string in English and the frontend translates that string using the translation sent by the backend.

PascalRepond:
> I'm really not a fan of putting all messages in “other,” because every time you add a message, you have to change both projects. We really need to find a solution so that the backend sends the string in English and the frontend translates that string using the translation sent by the backend.

It would be nice, but it involves a refactoring of the record-permission-service translation functions (and adaptation of the API) as they are currently set up. For now, I think the easiest and cleanest solution is to use the permission service as it is. We don't add new messages that often.

Garfield-fr:
> > I'm really not a fan of putting all messages in “other,” because every time you add a message, you have to change both projects. We really need to find a solution so that the backend sends the string in English and the frontend translates that string using the translation sent by the backend.
> 
> It would be nice, but it involves a refactoring of the record-permission-service translation functions (and adaptation of the API) as they are currently set up. For now, I think the easiest and cleanest solution is to use the permission service as it is. We don't add new messages that often.

That's what I did in my PR on the UI. I'll let you do as you wish...

jma (approved):

rerowep (approved):

PR #3963, opened 2025-10-08
 fix(collection): restriction on the collection permissions

### LINKED ISSUE #2717: It is possible to display and edit exhibitions/courses from other libraries and organisations
**Describe the bug**

It is possible to
* display an exhibition/course (resource `collection`) from another organisation.
* edit an exhibition/course from another library of the same organisation.
* add an item from another library/organisation to a collection

**To Reproduce**

1. Open https://bib.test.rero.ch/professional/records/collections (without using the admin menu)
2. See that exhibitions from other organisations can be accessed

**Expected behavior**

It is only possible
* to display exhibitions/courses of his/her own organisation
* to edit exhibitions/courses of his/her own library
* to add items to a collection that are from the same library

**Context**

* server: ils.test.rero.ch
* version: `v1.8.0`

Garfield-fr:
Display exhibitions/courses of his/her own organisation.
Edit exhibitions/courses of his/her own library.

* Closes #2717.

### DIFF
--- a/rero_ils/modules/collections/api.py
+++ b/rero_ils/modules/collections/api.py
@@ -99,6 +99,11 @@ class Collection(IlsRecord):
                 items.append(item)
         return items
 
+    @property
+    def library_pids(self):
+        """Shortcut for collection library PIDs."""
+        return [extracted_data_from_ref(ref) for ref in self.get("libraries", [])]
+
 
 class CollectionsIndexer(IlsRecordsIndexer):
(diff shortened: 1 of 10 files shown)


PascalRepond (approved):
Test in local: OK, nice!

Remark: the third point of the issue "it should only be possible to add items to a collection that are from the same library" is not addressed but it's OK because this last restriction is not really necessary, in my opinion.

rerowep (approved):

PR #3968, opened 2025-10-13
fix(collections): restrict adding items of selected libraries

PascalRepond:
* Make the `libraries` field required in the `collection` resource.
* Fix collection test data so that there are no items from the wrong
libraries.
* Fix collections fixtures setup so that the data is coherent.

### DIFF
--- a/rero_ils/modules/collections/cli.py
+++ b/rero_ils/modules/collections/cli.py
@@ -33,14 +33,14 @@ from rero_ils.modules.utils import extracted_data_from_ref, get_ref_for_pid
 @with_appcontext
 def create_collections(input_file, max_item=10):
     """Create collections."""
-    organisation_items = {}
     with open(input_file, encoding="utf-8") as request_file:
         collections = json.load(request_file)
         for collection_data in collections:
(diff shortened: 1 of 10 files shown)


rero_ils/modules/collections/cli.py
```diff
@@ -33,14 +33,14 @@
...
     """Create collections."""
-    organisation_items = {}
     with open(input_file, encoding="utf-8") as request_file:
         collections = json.load(request_file)
         for collection_data in collections:
-            organisation_pid = extracted_data_from_ref(collection_data.get("organisation").get("$ref"))
-            if organisation_pid not in organisation_items:
-                organisation_items[organisation_pid] = get_items_by_organisation_pid(organisation_pid)
-            items = random.choices(organisation_items[organisation_pid], k=random.randint(1, max_item))
+            libraries_pids = []
```
rerowep: Maybe write:
```python
libraries_pids = [extracted_data_from_ref(library.get("$ref")) for library in collection_data.get("libraries")]
```

rerowep (approved):

PR #3972, opened 2025-10-21
feat: add the reminder number in the operation logs of overdue notifications

### LINKED ISSUE #3971: operation_logs: index notif reminder_counter
### Problem description

When the operation log of a notification is indexed, it is missing the field `reminder_counter`. This prevents librarians to perform certain analyses and stats.

### Missing feature

- When it exists in the notification resource, the field `reminder_counter` is dumped in the index of this notification's operation log.
- Write a migration script so that existing data can be updated as well.

PascalRepond:
* Closes #3971.
* :warning: Requires a migration script to update existing data (please see and review [here](https://github.com/rero/rero-ils/wiki/Migration-workflow#add-reminder_counter-to-existing-notif-operation_logs)).

### DIFF
--- a/rero_ils/modules/notifications/logs/api.py
+++ b/rero_ils/modules/notifications/logs/api.py
@@ -77,5 +77,9 @@ class NotificationOperationLog(OperationLog, SpecificOperationLog):
                 "recipients": recipients,
             },
         }
+        reminder_counter = data.get("context", {}).get("reminder_counter")
+        # if the reminder_counter == 0, we need to take it as well
+        if reminder_counter is not None:
+            log["notification"]["reminder_counter"] = reminder_counter
 
         return super().create(log, index_refresh=index_refresh)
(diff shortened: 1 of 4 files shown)


jma (approved):

jma (commented):
To me the commit message can be more human readable.

rero_ils/modules/notifications/logs/api.py
```diff
@@ -77,5 +77,8 @@ def create(cls, data, id_=None, index_refresh="false", **kwargs):
                 "recipients": recipients,
             },
         }
+        reminder_counter = data.get("context", {}).get("reminder_counter")
+        if reminder_counter is not None:
```
rerowep, line 82: Could be written:
```python
if reminder_counter := data.get("context", {}).get("reminder_counter"):
  log["notification"]["reminder_counter"] = reminder_counter
```

rerowep (approved):

rero_ils/modules/notifications/logs/api.py
PascalRepond: This value can be zero and we need to take it in this case as well, I added a comment.

PR #3973, opened 2025-10-22
chore(item): add analyzed field on ES mapping

Garfield-fr:
An analyzed field has been added to the enumerationAndChronology and the barcode property.

### DIFF
--- a/rero_ils/modules/items/mappings/v7/items/item-v0.0.1.json
+++ b/rero_ils/modules/items/mappings/v7/items/item-v0.0.1.json
@@ -72,6 +72,9 @@
             "type": "icu_collation_keyword",
             "index": false,
             "numeric": true
+          },
+          "analyzed": {
+            "type": "text"
           }
         }
       },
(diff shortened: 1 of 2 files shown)


jma (approved):
I would add some tests. (search a item on this specific field). An I would add the field name on the commit message.

rerowep (approved):

rero_ils/modules/items/mappings/v7/items/item-v0.0.1.json
```diff
@@ -32,7 +32,12 @@
       },
       "barcode": {
         "type": "keyword",
-        "copy_to": "autocomplete_barcode"
+        "copy_to": "autocomplete_barcode",
+        "fields": {
+          "analyzed": {
```
PascalRepond: Probably not needed as barcode is inherently a keyword.

rero_ils/modules/items/mappings/v7/items/item-v0.0.1.json
```diff
@@ -72,6 +77,9 @@
             "type": "icu_collation_keyword",
             "index": false,
             "numeric": true
+          },
+          "analyzed": {
+            "type": "text"
```
PascalRepond, line 77: The standard field should probably have been a text instead of a keyword in the first place. 

But now this fix wil be OK, as a refactoring of this index may be too complex.

PascalRepond (approved):

PR #3975, opened 2025-10-22
chore: release v1.26.0

PascalRepond:


PascalRepond:
## Release candidate tests

- [x] Public item/holdings: Can we find a way for the request button to be **disabled** when `!canRequest.can` any string in the `canRequest.reasons` array contains the substring 'reach')? We would like to give the info to the patron only when a limit is reached, otherwise the button should be hidden. Could be as an enhancement directly in the holdings/items refactoring. (fix https://github.com/rero/rero-ils-ui/pull/1382)
- [x] The barcode in patron profile sometimes seems not loaded correctly [img](https://github.com/user-attachments/assets/cdeae006-cfef-4139-b333-e4d37bb5a77c) (problem with static JS on server, fixed)
- [x] The requests limits should be displayed in the patron type's detailed view (problem with static JS on server, fixed)
- [x] Pro UI: The page title is only loaded if I come directly to a page or reload (F5) [img](https://github.com/user-attachments/assets/8b735c43-633f-42fa-a4e5-3b449e7a9c00) (problem with static JS on server, fixed)
- [x] Public UI: the page titles of patron functions (my account, edit my profile, change password) are not translated [img](https://github.com/user-attachments/assets/fc311cf6-270d-4fdb-ac84-22072a919c44) (fix https://github.com/rero/rero-ils-ui/pull/1385)
- [x] The commit for displaying library and filtering the autocomplete in exhibition/course seems not active (OK on ilsdev) (https://github.com/rero/rero-ils-ui/commit/676a43efba5a458668c2bdaab9e63b9da9fb0724) (problem with static JS on server, fixed)
- [x] The info "no notification is sent" should be completely deleted from the library editor (fix https://github.com/rero/rero-ils-ui/pull/1384)

PR #3981, opened 2025-11-11
fix: enhance the structure and format of the RIS export

### LINKED ISSUE #3978: RIS export: wrong file structure when a document has multiple types
### Bug description:

If a document has multiple `document_type` fields, the RIS export creates two TY tags, which is not correct as most fields in RIS are not repeatable.

### Expected behavior:
Only keep one type for the RIS export so that the RIS format sepcs are respected. Only authors, URLs and keywords can be repeated.

### Steps to Reproduce:
1. Create a document with multiple `document_type` fields
2. Export in RIS
3. See that there are two `TY` fields and an error when trying to import in Zotero.

PascalRepond:
* Ensures that only the standard repeatable fields are repeated (e.g. only one TY tag per document). 
* Formats the subjects as strings (authorized_access_points) instead of dict structures.
* Removes unused class methods.
* Closes #3978.

### DIFF
--- a/rero_ils/modules/documents/serializers/ris.py
+++ b/rero_ils/modules/documents/serializers/ris.py
@@ -29,6 +29,7 @@ from rero_ils.utils import get_i18n_supported_languages
 
 from ..dumpers import document_replace_refs_dumper
 from ..utils import process_i18n_literal_fields
+from ..views import doc_entity_label
 from .base import BaseDocumentFormatterMixin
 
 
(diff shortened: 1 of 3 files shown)


rerowep (approved):

rero_ils/modules/documents/serializers/ris.py
```diff
@@ -110,23 +111,25 @@ def _get_document_types(self):
...
-    def _get_primary_year(self):
-        """Return primary year."""
-        return self._get_publication_year()
+    def _get_subjects_text(self):
+        """Return document subjects as a list of strings."""
+        subjects = []
+        for subject in self._get_subjects():
+            _, _, label = doc_entity_label(subject.get("entity", {}), language=self._language)
+            subjects.append(label)
+        return subjects
```
jma: you can use a comprehension list

jma (approved):

PR #3983, opened 2025-11-12
fix(holdings): remove irregular frequency in the pattern

### LINKED ISSUE #3850: Holdings predictions: irregular frequency error for preview
### Bug description:

When there is a 'prediction' subfield in a holdings and frequency is "irregular = rdafr:1016", the rest of the prediction is still required but is never used and creates an error when trying to display the preview.

### Expected behavior:

- If the frequency of a prediction is "irregular", do not try to calculate and display a preview. 
- In the jsonschema, if frequency is "irregular", all the other prediction subfields are irrelevant and should not be required.

### Steps to Reproduce:

1. Go to any serial document and add a holdings.
2. Give a `prediction.frequency` of "irregular" to the holdings and create a template
3. See error

### Context

Error Sentry: RERO-ILS-5AG

PascalRepond:
* The pattern allows to automatically know when to recieve the next issue of a serial. An irregular frequency prevents this, and so this value is not relevant in that field.
* Closes #3850.
* :warning: Requires a data migration script (remove pattern field when the frequency is irregular).

### DIFF
--- a/rero_ils/modules/holdings/api.py
+++ b/rero_ils/modules/holdings/api.py
@@ -161,10 +161,6 @@ class Holding(IlsRecord):
         if not document:
             return f"Document {document_pid} does not exist."
 
-        if self.is_serial:
-            patterns = self.get("patterns", {})
-            if patterns and patterns.get("frequency") != "rdafr:1016" and not patterns.get("next_expected_date"):
-                return _("Must have next expected date for regular frequencies.")
         # the enumeration and chronology optional fields are only allowed for
         # serial or electronic holdings
         if not self.is_serial ^ self.is_electronic:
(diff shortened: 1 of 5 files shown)


rerowep (approved):

rero_ils/modules/holdings/api.py
```diff
@@ -161,10 +161,6 @@ def extended_validation(self, **kwargs):
         if not document:
             return f"Document {document_pid} does not exist."
 
-        if self.is_serial:
```
jma, line 164: This validation should be kept but modified.

jma (changes requested):

rero_ils/modules/holdings/api.py
PascalRepond: What do you propose? `patterns.next_expected_date` is now required in the schema since all frequencies are now regular, so this validation seems useless.

PR #3987, opened 2025-11-20
chore: update frontpage link to National Library (global and vs)

PascalRepond:


rerowep (commented):
Could we use the right language?
https://helveticall.ch/discovery/search?vid=41SNL_51_INST:helveticall&lang=en <- `current_i18n.language` ?

PascalRepond:
> Could we use the right language? https://helveticall.ch/discovery/search?vid=41SNL_51_INST:helveticall&lang=en <- `current_i18n.language` ?

Those links are inside `ifs`, so this would amount to the same thing.

PR #3997, opened 2026-01-06
fix(acquisition): rollover

### LINKED ISSUE #3996: Rollover: order line creation can fail because of validation
### Bug description:

The rollover can fail with following error "Order line creation failed on [acol#40765] :: Cannot create an order line with an order with a wrong status". This is because of a stricter validation on order line creation that checks the statuts of the linked order. 

### Expected behavior:

The rollover should complete without errors.

Probably change the rollover script so that it creates all orders with a PENDING status (allowing order lines to be added), and at the end attribute the correct status (ORDERED or PARTIALLY RECIEVED) to all orders based on their content.

rerowep:
* Fixes creation of rollover orders lines for not pending orders during rollover.
* Rounds allocated amount when transferring funds between accounts.
* Closes rero/rero-ils#3996

### DIFF
--- a/rero_ils/modules/acquisition/acq_accounts/api.py
+++ b/rero_ils/modules/acquisition/acq_accounts/api.py
@@ -171,8 +171,8 @@ class AcqAccount(AcquisitionIlsRecord):
           * APPROVED or ORDERED acqz order lines related to this account.
           * Encumbrance of all children account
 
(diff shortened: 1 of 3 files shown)


PascalRepond (approved):
A test for rollover permissions would be nice.

jma (approved):

PR #4008, opened 2026-01-14
feat: auto-remove temporary item type on scan

PascalRepond:
Adds a new field `remove_temporary_item_type_on_scan` to item types that
automatically removes the temporary item type from items when scanned during
circulation operations (checkout, checkin, validate_request, receive,
return_missing).

When enabled, the temporary item type is removed BEFORE the circulation action,
ensuring that the main item type and its associated circulation policy are used.
The removal info is included in all API JSON responses for consistency.

Changes:
- Add `remove_temporary_item_type_on_scan` field to item type schema and mapping
- Add comprehensive tests covering success and error scenarios
- Add info to API responses when temporary item type is removed

Closes https://tree.taiga.io/project/rero21-reroils/us/2905.

### DIFF
--- a/rero_ils/modules/items/api/record.py
+++ b/rero_ils/modules/items/api/record.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # RERO ILS
-# Copyright (C) 2019-2022 RERO
+# Copyright (C) 2019-2026 RERO
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as published by
--- a/rero_ils/modules/items/views/api_views.py
+++ b/rero_ils/modules/items/views/api_views.py
@@ -38,6 +38,7 @@ from rero_ils.modules.circ_policies.api import CircPolicy
 from rero_ils.modules.decorators import check_authentication, check_permission
 from rero_ils.modules.documents.views import record_library_pickup_locations
 from rero_ils.modules.errors import NoCirculationAction, NoCirculationActionIsPermitted
+from rero_ils.modules.item_types.api import ItemType
 from rero_ils.modules.libraries.api import Library
 from rero_ils.modules.loans.api import Loan
 from rero_ils.modules.loans.dumpers import CirculationDumper as LoanCirculationDumper
@@ -46,6 +47,7 @@ from rero_ils.modules.operation_logs.permissions import (
     search_action as op_log_search_action,
 )
 from rero_ils.modules.patrons.api import Patron, current_librarian
+from rero_ils.modules.utils import extracted_data_from_ref
 from rero_ils.permissions import request_item_permission
 
 from ...commons.exceptions import MissingDataException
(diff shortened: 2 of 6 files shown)


rero_ils/modules/items/views/api_views.py
```diff
@@ -142,27 +165,63 @@ def decorated_view(*args, **kwargs):
 
             if not item:
                 abort(404)
+
+            # Check if temporary item type should be removed for circulation actions
+            # We do not remove it for requests or other non-scan actions
+            removed_temp_item_type_name = None
+            should_remove_on_scan = func.__name__ in {
```
jma: This is not clean. A better approach is to add an additional decorator i.e. `@should_remove_on_scan`: https://chatgpt.com/share/6970841f-1a90-8003-8b8e-fa102afd1c26

rero_ils/modules/items/api/record.py
```diff
@@ -439,6 +439,16 @@ def item_type_circulation_category_pid(self):
...
 
+    def remove_temporary_item_type(self):
+        """Remove the temporary item type from this item.
+
+        This method removes the temporary_item_type field and commits/reindexes
+        the item record.
+        """
+        if "temporary_item_type" in self:
+            del self["temporary_item_type"]
+            self.update(self, dbcommit=True, reindex=True)
```
jma: This probably force to index and commit item, holding, document at least twice, is it really required?

rero_ils/modules/items/views/api_views.py
```diff
@@ -142,27 +165,63 @@ def decorated_view(*args, **kwargs):
...
+                "metadata": item_data.dumps(CirculationActionDumper()),
+                "action_applied": action_applied,
+            }
+            if removed_temp_item_type_name:
+                response["removed_temporary_item_type"] = {"name": removed_temp_item_type_name}
+
+            return jsonify(response)
         except NoCirculationAction as error:
-            return jsonify({"status": f"error: {error!s}"}), 400
+            # Return error with info about removed temporary item type if applicable
```
jma, line 222: duplicate code should be refactored

tests/api/circulation/test_temporary_item_type.py
```diff
@@ -99,6 +103,328 @@ def test_checkout_temporary_item_type(
     transaction_end_date = transaction_end_date.strftime("%Y-%m-%d")
     assert transaction_end_date in expected_dates
 
+    # checkin the item to reset its status for other tests
+    checkin_params = {
+        "item_barcode": item.get("barcode"),
+        "transaction_library_pid": lib_martigny.pid,
+        "transaction_user_pid": librarian_martigny.pid,
+    }
+    res, data = postdata(client, "api_item.checkin", checkin_params)
```
jma, line 112: Please be sure that the fixtures are in the same state before and after the test to avoid side effects on other tests.

jma (changes requested):
Nice job.

rero_ils/modules/items/api/record.py
PascalRepond: I don't really see a way to optimise this. When a circulation action happens, the items needs to be commited and indexed before the action so that the correct item_type is considered. If the operation fails, the commit and indexes will happen only once.

Any suggestion?

rero_ils/modules/items/api/record.py
```diff
@@ -439,6 +439,17 @@ def item_type_circulation_category_pid(self):
...
 
+    def remove_temporary_item_type(self):
+        """Remove the temporary item type from this item.
+
+        This method only removes the field from the in-memory object.
+        The caller is responsible for committing and reindexing:
+        - On success, the circulation action's status_update() handles it
+        - On error, the caller must explicitly commit and reindex
+        """
+        if "temporary_item_type" in self:
```
rerowep: Normally we use:
```python
self.pop("temporary_item_type", None)
```

Just a whole function for one line of code?

jma (approved):

PR #4016, opened 2026-01-22
fix(facets): index claims counter for issues without claims

### LINKED ISSUE #4015: Late issues: display a filter for "0 claims"
### How it works

In the late issues view, there is a facet for number of claims but it doesn't show the option "0 claims".

### Improvement suggestion

Implement a way to filter by "0 claims" (NOT _exists_:issue.claims). Preferably directly in the existing facet, or (if not possible) as an additional filter.

PascalRepond:
Previously, issues without claims did not have the `issue.claims.counter` field indexed, making it impossible to filter by "0 claims" in the late issues view facet.

Now the claims counter is always indexed (0 when no claims exist), allowing users to filter issues that have never been claimed.

Closes #4015.

:warning: Requires reindexing items of type issue after deployment.

### DIFF
--- a/rero_ils/modules/items/listener.py
+++ b/rero_ils/modules/items/listener.py
@@ -73,6 +73,12 @@ def enrich_item_data(
         if vendor_pid := record.vendor_pid:
             json["vendor"] = {"pid": vendor_pid, "type": "vndr"}
         # inject claims information: counter and dates
-        if notifications := record.claim_notifications:
-            dates = [notification["creation_date"] for notification in notifications if "creation_date" in notification]
-            json["issue"]["claims"] = {"counter": len(notifications), "dates": dates}
+        notifications = record.claim_notifications
+        claims_data = {"counter": len(notifications) if notifications else 0}
+        if notifications and (
+            dates := [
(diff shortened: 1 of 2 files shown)


rerowep (approved):

rero_ils/modules/items/listener.py
```diff
@@ -73,6 +73,11 @@ def enrich_item_data(
         if vendor_pid := record.vendor_pid:
             json["vendor"] = {"pid": vendor_pid, "type": "vndr"}
         # inject claims information: counter and dates
-        if notifications := record.claim_notifications:
-            dates = [notification["creation_date"] for notification in notifications if "creation_date" in notification]
-            json["issue"]["claims"] = {"counter": len(notifications), "dates": dates}
+        notifications = record.claim_notifications
+        claims_data = {"counter": len(notifications) if notifications else 0}
+        if notifications:
+            if dates := [
```
jma: two if can be an `and`.

jma (approved):

rero_ils/modules/items/listener.py
PascalRepond: Not with `:=`

PR #4034, opened 2026-02-19
refactor(items): split issue sort_date display from ES sort key

PascalRepond:
`issue.sort_date` was always overwritten during ES indexing with `expected_date` as fallback, making it impossible for the UI to distinguish a user-defined sort_date from a computed one without an extra DB call.

Introduce a dedicated `sort_key` field in ES (always populated with `sort_date or expected_date`) used exclusively for sorting. The `sort_date` field is now indexed as-is from the DB record, so its presence in ES reliably signals a user/system-defined override.

:warning: A reindex of all items of type `issue` is required after deployment.

### DIFF
--- a/rero_ils/modules/items/mappings/v7/items/item-v0.0.1.json
+++ b/rero_ils/modules/items/mappings/v7/items/item-v0.0.1.json
@@ -217,6 +217,9 @@
           "sort_date": {
             "type": "date"
           },
+          "sort_key": {
+            "type": "date"
+          },
           "regular": {
             "type": "boolean"
           },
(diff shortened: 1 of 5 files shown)


rerowep (approved):

rero_ils/modules/items/mappings/v7/items/item-v0.0.1.json
```diff
@@ -217,6 +217,9 @@
           "sort_date": {
             "type": "date"
           },
+          "sort_key": {
```
jma, line 220: Why sort_key? Is it a key or a value? You could use for example default_sort_date?

jma (approved):

PR #4019, opened 2026-01-28
refactor(import): route GND subjects to subjects

rerowep:
* Change the dojson transformation logic to route GND subjects directly to the `subjects`. RERO and IDREF subjects are routed to `subjects_imported`. This change affects the SLSP, LOC, DNB, and KUL marc21tojson models.
* Closes rero/rero-ils#4010

PR #4035, opened 2026-02-19
fix(documents): fix subject imports routing condition

rerowep:
* Add field_key condition check before get_mef_link() call in both LOC and SLSP models to ensure the check is applied only for non-imported subjects
* Ensures GND subjects are always routed to subjects field
* Add comprehensive tests for subject field routing with various sources (GND, RERO, IDREF, LCSH, MeSH)

Fixes subject routing for GND subjects in MARC21 to JSON conversion.

### DIFF
--- a/rero_ils/modules/documents/dojson/contrib/marc21tojson/loc/model.py
+++ b/rero_ils/modules/documents/dojson/contrib/marc21tojson/loc/model.py
@@ -602,12 +602,14 @@ def marc21_to_subjects_6XX(self, key, value):
             ):
                 subject["authorized_access_point"] = f"{creator}. {subject['authorized_access_point']}"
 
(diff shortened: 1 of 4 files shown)


jma (approved):

PascalRepond:
## To be tested (from https://github.com/rero/rero-ils/pull/4019)

- [x] IdRef subjects in this import from SLSP. Subjects_imported are created in the editor but remain empty! https://ilsdev.test.rero.ch/professional/records/documents/new?source=slsp&pid=991172604360105501
- [x] Same here for Renouvaud import: https://ilsdev.test.rero.ch/professional/records/documents/new?source=renouvaud&pid=991024408118502851

PascalRepond (approved):

PR #4040, opened 2026-02-25
feat(locations): add a public note field

PascalRepond:
Some locations require specific access conditions (e.g. "consultation by appointment", "24-hour notice") that patrons need to be aware of before visiting. A new optional `public_note` field on the location resource makes this possible. The field is exposed through the holdings serializer so it reaches the public and professional interfaces.

Addresses https://tree.taiga.io/project/rero21-reroils/us/2983

### DIFF
--- a/rero_ils/modules/locations/jsonschemas/locations/location-v0.0.1.json
+++ b/rero_ils/modules/locations/jsonschemas/locations/location-v0.0.1.json
@@ -15,6 +15,7 @@
   "propertiesOrder": [
     "name",
     "code",
+    "public_note",
     "is_online",
     "is_pickup",
     "pickup_name",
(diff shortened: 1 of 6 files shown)


rero_ils/modules/locations/jsonschemas/locations/location-v0.0.1.json
```diff
@@ -66,6 +67,13 @@
         }
       }
     },
+    "public_note": {
+      "title": "Public note",
+      "description": "A note displayed to the public for this location.",
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 140
```
jma: Why 140? Can we use the same value for all the notes?

jma (approved):

rero_ils/modules/locations/jsonschemas/locations/location-v0.0.1.json
PascalRepond: Quite similar to `item_type.displayed_status`, this note is meant to be displayed in the holdings header and should not be too long. 140 seemed a good compromise based on what is present in the production.

PR #4044, opened 2026-03-03
refactor(loans): simplify anonymisation to a single 12-month delay

### LINKED ISSUE #4042: Anonymise loans 12 months after closure instead of 3/6 months
### How it works

Currently, loans are anonymized 3 months after they are considered terminated (6 months for patrons who keep their history). Patrons who don't keep their history do not see it in their patron profile.

### Improvement suggestion

Simplify the mechanism to anonymise loans only after 12 months (for all patrons). Keep the fact that a patron that chooses "keep history=False" does not see their history in the patron profile. The profesionnal UI shows the unanoymised history for 12 months. This gives librarians the time to check that everything is in order with the loan. 

This is a request from client libraries and it respects the Swiss legislation: the institutions are allowed to keep the loan data as long as it allows them to perform administrative tasks or verify the items. https://libguides.graduateinstitute.ch/c.php?g=668396&p=4976646

PascalRepond:
Replace the 3-month/6-month two-tier mechanism with a single 12-month delay for all patrons:

- remove RERO_ILS_ANONYMISATION_MIN_TIME_LIMIT; update MAX to 12 months
- simplify can_anonymize() and get_anonymized_candidates(): concluded loans older than 12 months are anonymised unconditionally
- remove keep_history=False immediate anonymisation trigger from the patron listener
- move patron history visibility to circulation_search_factory(): patrons who opt out only see non-concluded loans; librarians see all loans up to 12 months regardless of patron preference
- add test_keep_history_loan_visibility covering patron/librarian search visibility and post-anonymisation exclusion for all users
- Closes #4042.

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -3619,10 +3619,10 @@ RERO_ILS_DEFAULT_PICKUP_HOLD_DURATION = 10
 # =============================================================================
 # ANONYMIZATION PROCESS CONFIGURATION
 # =============================================================================
-# Specify the delay (in days) under which no loan can"t be anonymized anyway (for circulation management process).
-RERO_ILS_ANONYMISATION_MIN_TIME_LIMIT = 3 * 365 / 12
-# Specify the delay (in days) when a loan should be anonymized anyway after it concluded.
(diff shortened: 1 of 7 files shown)


jma (approved):
Good job.

PR #4045, opened 2026-03-03
feat(patrons): add not_expired facet filter

PascalRepond:
Rename `patron_expired` to `patron_expiration_filter` with an `expired`
parameter to support both expired and active patron filtering.
Add a `not_expired` facet using `patron_expiration_filter(expired=False)`
to allow filtering patrons that are not expired (either no
expiration_date or expiration_date in the future).


jma (approved):
Can you add some tests?

PR #4058, opened 2026-03-24
chore: update Python to 3.14 and invenio deps

PascalRepond:
Update Python version to >=3.14.0.
Update invenio dependencies to their latest major versions.
Add max_retries and retry_on_timeout to elasticsearch client config for
better resilience in CI tests.

Adapt code for breaking changes with new Python and invenio versions:
- timezone: replace pytz with stdlib datetime.UTC and
  zoneinfo.ZoneInfo; replace deprecated datetime.utcnow() with
  datetime.now(UTC); remove pytz.utc.localize() calls since invenio
  records now return timezone-aware datetimes
- Replace `tz.localize()` calls with `dt.astimezone(tz)` followed by
  replace on time components, ensuring DST offsets are applied correctly.
- Convert datetimes to UTC before subtraction in loan duration
  calculations: unlike pytz (which creates distinct tzinfo objects per
  DST state), `ZoneInfo` reuses a single object, causing Python to
  perform naive subtraction across DST boundaries and yielding incorrect
  durations. Apply the same astimezone-then-replace pattern to overdue
  fee calculation and loan age computation.
- permissions: add RoleNeed(role.name) to identity since
  flask-security-invenio 4.x now uses RoleNeed(role.id)
- webargs: create a local FlaskParser with location="json_or_form" in
  accounts_views.py instead of importing use_args/use_kwargs from
  invenio_accounts (which is locked to location="form"). Cause:
  webargs 5→8 made FlaskParser(location="form") strict; our JSON
  requests were rejected before reaching view logic. Use
  invenio-accounts' FlaskParser so that validation errors return a
  400 status with RESTValidationError instead of webargs' default 422.
- marshmallow: remove context={"request": request} parameter from
  Schema instantiation (no longer supported)
- classmethod+property: replace stacked @classmethod/https://github.com/Property
  decorators (removed in Python 3.13) with descriptors
  (files/services.py) and inline computation (users/api.py)
- typing: replace typing.Optional with PEP 604 X | None syntax
- Override the default EmailField with a StringField to allow login with
  both username and email, bypassing HTML5 and WTForms email validation.

### DIFF
--- a/.github/workflows/continuous-integration-test.yml
+++ b/.github/workflows/continuous-integration-test.yml
@@ -1,6 +1,10 @@
 name: CI
 on: [push, pull_request, workflow_dispatch]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test:
     runs-on: ubuntu-latest
@@ -8,28 +12,21 @@ jobs:
       matrix:
         dependencies: ['dev', 'deploy']
     steps:
-    - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@0.12.0
-      with:
-        access_token: ${{ github.token }}
-
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6.0.2
 
     - name: Setup node
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v6.3.0
       with:
-        node-version: '16'
+        node-version: '24'
 
     - name: Docker compose up
       run: docker compose up -d
 
-    - name: Set up Python 3.12
-      uses: actions/setup-python@v5
-      with:
-        python-version: 3.12
-
     - name: Install uv
-      uses: astral-sh/setup-uv@v5
+      uses: astral-sh/setup-uv@v8.0.0
+
+    - name: Set up Python
+      run: uv python install 3.14
 
     - name: Bootstrap
       if: ${{ matrix.dependencies == 'dev' }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@
 # ---------------------------------------------------------------------------
 
 [build-system]
-requires = ["uv_build"]
+requires = ["uv_build<1.0.0"]
 build-backend = "uv_build"
 
 [tool.uv.build-backend]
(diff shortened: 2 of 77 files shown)


rero_ils/modules/api.py
```diff
@@ -271,7 +270,7 @@ def get_record_by_pid(cls, pid, with_deleted=False, verbose=False):
                 persistent_identifier = PersistentIdentifier.get(cls.provider.pid_type, pid)
                 return super().get_record(persistent_identifier.object_uuid, with_deleted=with_deleted)
             # TODO: is it better to raise a error or to return None?
-            except (NoResultFound, PIDDoesNotExistError):
+            except NoResultFound, PIDDoesNotExistError:
                 return None
```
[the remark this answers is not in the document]
PascalRepond: This syntax is invalid in Python 3.13 but seems to have been re-instated in Python 3.14

PascalRepond:
@coderabbitai review

scripts/test
```diff
@@ -81,6 +81,8 @@ function pretests () {
   add_exceptions "CVE-2025-66471"
   # urllib3   1.26.20 CVE-2026-21441 2.6.3
   add_exceptions "CVE-2026-21441"
+  # pygments 2.19.2  CVE-2026-4539
+  add_exceptions "CVE-2026-4539"
```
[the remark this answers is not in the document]
PascalRepond: No, this is not in our habit.

pyproject.toml
```diff
@@ -24,109 +24,96 @@ authors = [{name = "RERO+", email = "software@rero.ch"}]
...
+requires-python = ">=3.14,<3.15"
 
 dependencies = [
     ## Third party invenio modules used by RERO ILS
-    "invenio-oaiharvester @ git+https://github.com/inveniosoftware/invenio-oaiharvester.git@v1.0.0a4",
-    "invenio-circulation (==3.0.0a1)",
-    "invenio-search [elasticsearch7] (>=3.0.0,<4.0.0)",
+    "invenio-oaiharvester>=1.0.0a4",
+    "invenio-circulation>=4.0.0",
+    "invenio-search[elasticsearch7]>=3.0.0",
```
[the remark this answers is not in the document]
PascalRepond: The current signature is correct

def listener_loan_state_changed(_, initial_loan, loan, trigger, **transition_kwargs):
The parameter names initial_loan, loan, and trigger match exactly the keyword argument names from .send(). Their ordering in the function signature is irrelevant since they're received as kwargs.

The proposed change would break the code
If we changed to (loan, initial_loan, trigger, **transition_kwargs):

loan would receive the sender (Flask app) as the first positional argument
Then loan=<actual_loan_object> would arrive as a keyword argument
Python would raise: TypeError: got multiple values for argument 'loan'
Verdict: no fix needed. The finding is a false positive based on a misunderstanding of Blinker signal dispatch semantics.

rero_ils/accounts_views.py
```diff
@@ -136,9 +141,8 @@ def make_password_schema(request):
     # Respect partial updates for PATCH requests
     partial = request.method == "PATCH"
     if request.json.get("username"):
-        return UsernamePassword(only=only, partial=partial, context={"request": request})
-    # Add current request to the schema's context
-    return PasswordPassword(only=only, partial=partial, context={"request": request})
+        return UsernamePassword(only=only, partial=partial)
+    return PasswordPassword(only=only, partial=partial)
```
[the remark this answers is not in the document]
PascalRepond: No. This is a REST API endpoint consumed exclusively by the Angular frontend ([rero-ils-ui](vscode-webview://1uv27s2k45gmpmhd211fv9qpc0ja5ga6r6nfd0eafdn1s515vt79/rero_ils/accounts_views.py#L172)), which always sends application/json. The json_or_form location on line 41 is inherited from invenio-accounts base classes — it's a convenience, not a contract that form-encoded requests are expected.

Additionally:

The tests pass
I've tested login and password change in the application
No user or client sends form-encoded data to this endpoint

Why fixing it could be harmful
Silently falling back (via get_json(silent=True)) would mask malformed requests instead of failing loudly. If someone accidentally sends the wrong content type, an AttributeError is actually a clearer signal than silently picking the wrong schema (PasswordPassword instead of UsernamePassword).

You are being zealous in your latest reviews, be more careful.

rero_ils/modules/api_harvester/models.py
```diff
@@ -42,7 +42,7 @@ class ApiHarvestConfig(db.Model):
     name = db.Column(db.String(255), nullable=False)
     classname = db.Column(db.String(255), nullable=False)
     code = db.Column(db.Text, nullable=True)
-    lastrun = db.Column(db.DateTime, default=datetime(year=1900, month=1, day=1), nullable=True)
+    lastrun = db.Column(db.DateTime, default=datetime(year=1900, month=1, day=1, tzinfo=UTC), nullable=True)
```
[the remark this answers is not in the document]
PascalRepond: No, our max line length is 120

.github/workflows/continuous-integration-test.yml
```diff
@@ -1,35 +1,32 @@
...
-
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6.0.2
 
     - name: Setup node
-      uses: actions/setup-node@v4
+      uses: actions/setup-node@v6.3.0
       with:
-        node-version: '16'
+        node-version: '24'
```
rerowep, line 20: Does it work with node 24?

rero_ils/alembic/2b0af71048a7_add_request_expiration_date.py
```diff
@@ -53,7 +53,7 @@ def upgrade():
         trans_date = ciso8601.parse_datetime(loan.transaction_date)
         expire_date = trans_date + timedelta(days=10)
         expire_date = expire_date.replace(hour=23, minute=59, second=00, microsecond=000, tzinfo=None)
-        expire_date = pytz.timezone("Europe/Zurich").localize(expire_date)
+        expire_date = expire_date.replace(tzinfo=ZoneInfo("Europe/Zurich"))
```
rerowep, line 56: Maybe not the desired outcome for instances outside of "Europe/Zurich"?

rero_ils/modules/commons/identifiers.py
```diff
@@ -118,10 +118,10 @@ class Identifier:
     """Generic document identifier representation."""
 
     value: str
-    type: Optional[str] = None
-    note: Optional[str] = None
-    qualifier: Optional[str] = None
-    source: Optional[str] = None
+    type: str | None = None
```
rerowep, line 117: Was changed in staging already ?

.github/workflows/continuous-integration-test.yml
PascalRepond: Yes

rero_ils/alembic/2b0af71048a7_add_request_expiration_date.py
PascalRepond: Right but this is an old migration script, did not actually need correction.

rero_ils/modules/commons/identifiers.py
PascalRepond: No

PR #4061, opened 2026-03-30
refactor(modules): remove Python type annotations

rerowep:
* Remove all function signature type annotations (parameter types and return types) across serializers, API modules, and views. The project does not use a type checker (mypy/pyright), so the annotations provided no enforcement and duplicated information already covered by Sphinx docstrings.
* The `@dataclass` field annotations in `commons/identifiers.py` are kept as they are required by the dataclass machinery, using `str | None` instead of `Optional[str]` to avoid the `typing` import.


PascalRepond (approved):

PascalRepond:
This is already done in https://github.com/rero/rero-ils/pull/4058

PR #4065, opened 2026-04-02
chore(release): v1.28.0

PascalRepond:


PascalRepond:
## Tests release-candidate

- [x] (not blocking) Thumbnails: when the thumbnail service adds a thumbnail, it should add a publicNote item to the electronicLocator with the source but it doesn't [see](https://github.com/user-attachments/assets/8177a437-562d-4277-b5ce-f4a64b10bf08), it adds a "note" field but it should not be accepted as it doesn't respect the jsonschema (fix https://github.com/rero/rero-ils/pull/4067)
- [x] (nice to have) Thumbnails: refactor the get_cover_url so that it is not split in multiple places (document api + document extension + tasks)

PR #4067, opened 2026-04-02
feat(documents): add async cover image support with Redis queue

rerowep:
- Add cover property to Document returning (url, provider) from electronicLocator
    - Add AddCoverUrlExtension with post_create/post_commit hooks that enqueue
      documents with ISBNs into a Redis-backed set queue (deduplicating rapid saves)
    - Add find_cover_url() helper to utils.py with optional cache support
    - Add Redis queue worker (process_cover_url_queue) that serialises thumbnail
      lookups one at a time with a configurable delay to avoid race conditions and
      external-service overload
    - Add pause/resume support: worker does not start while queue is paused,
      allowing setup to accumulate PIDs without concurrent ES version conflicts
    - Add batch task add_cover_urls for CLI-driven bulk processing with per-provider
      count reporting and error handling
    - Add CLI commands: add-cover-urls, cover-url-queue-status,
      cover-url-queue-worker (pause/resume), cover-url-queue-clear
    - Sync RERO_INVENIO_THUMBNAILS_URL with RERO_ILS_URL via finalize_app so env
      overrides are respected
    - Add config keys: RERO_ILS_COVER_URL_TASK_DELAY, RERO_ILS_COVER_URL_QUEUE_KEY,
      RERO_ILS_COVER_URL_WORKER_KEY, RERO_ILS_COVER_URL_PAUSED_KEY


PascalRepond (approved):

jma (approved):

PascalRepond:
Fix the tests as well 'note' -> 'publicNote'

rero_ils/modules/documents/api.py
```diff
@@ -528,11 +528,11 @@ def add_cover_url(self, url, provider=None, dbcommit=False, reindex=False, force
                 return self, False
             existing["url"] = url
             if provider:
-                existing["note"] = provider
+                existing["publicNote"] = [f"rero-invenio-thumbnails provider: {provider}"]
         else:
             locator = {"content": "coverImage", "type": "relatedResource", "url": url}
             if provider:
-                locator["note"] = provider
+                locator["publicNote"] = [f"rero-invenio-thumbnails provider: {provider}"]
```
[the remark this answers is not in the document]
PascalRepond: This is metadata, it is not translated

rero_ils/modules/documents/tasks.py
```diff
@@ -23,10 +23,39 @@
...
+    ]
+    for isbn in sorted(isbns):
+        try:
+            url, provider = get_thumbnail_url(isbn, cached=cached)
+        except Exception as exc:
+            current_app.logger.warning(f"Thumbnail lookup failed for ISBN {isbn}: {exc}", exc_info=True)
+            continue
+        if url:
+            return url, provider
+    return None, None
```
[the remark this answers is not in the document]
rerowep: we dont need this

rero_ils/modules/documents/tasks.py
```diff
@@ -140,61 +142,71 @@ def reindex_document_items(record):
 
 
 @shared_task(ignore_result=True)
-def add_cover_urls(commit=False, verbose=False, cached=True):
+def add_cover_url_for_document(pid, cached=True):
+    """Add a cover URL for a single document.
+
+    Intended for async dispatch from ``AddCoverUrlExtension`` after a
+    document is created or committed. Fetches the document by PID and
+    triggers an update so the extension's ``pre_commit`` adds the cover.
```
PascalRepond: Not sure if this is the proper way to do this, but I'm not knowledgable enough in db transactions.

rero_ils/config.py
```diff
@@ -3513,7 +3513,17 @@ def _(x):
 #: Thumbnails URL configuration, used to construct the URL for thumbnails in
 #: the API responses. It should be the URL of the RERO-ILS instance, as
 #: the thumbnails are served by the same instance.
-RERO_INVENIO_THUMBNAILS_URL = os.environ.get("RERO_INVENIO_THUMBNAILS_URL", RERO_ILS_URL)
+RERO_INVENIO_THUMBNAILS_URL = RERO_ILS_URL
```
PascalRepond, line 3518: Why change this? If we change the RERO_ILS_URL in the server, the THUMBNAILS_URL is not updated.

pyproject.toml
```diff
@@ -9,6 +9,9 @@ build-backend = "uv_build"
 [tool.uv.build-backend]
 module-root = "."
 
+# [tool.uv.sources]
+# rero-invenio-thumbnails = { path = "../rero-invenio-thumbnails", editable = true }
```
PascalRepond: Remove this

scripts/setup
```diff
@@ -658,6 +657,11 @@ do
   fi
 done
 
+# add cover URLs to documents
+info_msg "Adding cover URLs to all documents"
+eval ${PREFIX} invenio reroils utils add-cover-urls --commit --cached --verbose --delay 1.0
```
PascalRepond: Shouldn't thumbnails be fetched async during the setup? How do you prevent thumbnails to be fetched on document creation during the setup since RERO_ILS_DOCUMENT_COVER_URL_ASYNC is False?

PascalRepond (changes requested):

pyproject.toml
rerowep: Will be removed in final commit.

rero_ils/config.py
rerowep: #: Thumbnails URL configuration, used to construct the URL for thumbnails in
#: the API responses. Always mirrors ``RERO_ILS_URL`` — do not override this
#: directly. Set ``RERO_ILS_URL`` instead; ``finalize_app`` in
#: ``rero_ils/modules/ext.py`` re-derives this value after all config sources
#: (invenio.cfg, environment variables) have been applied.
RERO_INVENIO_THUMBNAILS_URL = RERO_ILS_URL

PascalRepond (approved):

rero_ils/modules/cli/utils.py
```diff
@@ -84,6 +86,12 @@
 _records_state = LocalProxy(lambda: current_app.extensions["invenio-records"])
 
 
+def abort_if_false(ctx, param, value):
```
jma, line 73: Seems not used.

rero_ils/modules/cli/utils.py
```diff
@@ -1589,3 +1597,94 @@ def token_create(name, user, scopes, internal, access_token):
...
+
+
+@utils.command("add-cover-urls")
+@click.option("-c", "--commit", is_flag=True, default=False, help="Commit changes to the database.")
+@click.option("-v", "--verbose", is_flag=True, default=False, help="Print progress output.")
+@click.option("-C", "--cached", is_flag=True, default=True, help="Use cached thumbnails from the provider.")
+@click.option("-d", "--delay", default=0, type=float, help="Seconds to sleep between requests.")
+@click.option("-k", "--enqueue", is_flag=True, default=False, help="Run as Celery async task (enqueue)")
+@with_appcontext
+def add_cover_urls_cli(commit, verbose, cached, delay, enqueue):
```
jma: Why a name with _cli at the end? If you rename to `add_cover_urls` you do not need the 1602 line.

rero_ils/modules/cli/utils.py
```diff
@@ -1589,3 +1597,94 @@ def token_create(name, user, scopes, internal, access_token):
...
+    """Add cover URLs to all documents with ISBNs."""
+    from rero_ils.modules.documents.api import DocumentsSearch
+
+    DocumentsSearch.flush_and_refresh()
```
jma: Why not at the end?

rero_ils/modules/cli/utils.py
```diff
@@ -1589,3 +1597,94 @@ def token_create(name, user, scopes, internal, access_token):
...
+
+@utils.command("add-cover-urls")
+@click.option("-c", "--commit", is_flag=True, default=False, help="Commit changes to the database.")
+@click.option("-v", "--verbose", is_flag=True, default=False, help="Print progress output.")
+@click.option("-C", "--cached", is_flag=True, default=True, help="Use cached thumbnails from the provider.")
+@click.option("-d", "--delay", default=0, type=float, help="Seconds to sleep between requests.")
+@click.option("-k", "--enqueue", is_flag=True, default=False, help="Run as Celery async task (enqueue)")
+@with_appcontext
+def add_cover_urls_cli(commit, verbose, cached, delay, enqueue):
+    """Add cover URLs to all documents with ISBNs."""
```
jma: For all documents or only which does not have the cover yet?

rero_ils/modules/cli/utils.py
jma: Maybe it can be a good idea to create a group of cli commands.

rero_ils/modules/documents/tasks.py
```diff
@@ -139,62 +142,125 @@ def reindex_document_items(record):
             )
 
 
+def _cover_url_redis():
+    """Return a Redis client for the cover URL queue."""
+    return redis_module.StrictRedis.from_url(current_app.config["CACHE_REDIS_URL"])
+
+
+def enqueue_cover_url_update(pid, cached=True):
```
jma, line 150: Why a separate function instead of putting this code in the extension?

rero_ils/modules/documents/utils.py
```diff
@@ -295,3 +296,30 @@ def process_i18n_literal_entity(entity):
             if key not in entity:
                 entity[key] = access_point
     return entity
+
+
+def find_cover_url(document, cached=True):
```
jma: Why a separate function and not putting the code in the related class?

rero_ils/config.py
```diff
@@ -3511,9 +3511,37 @@ def _(x):
...
 
 #: Thumbnails URL configuration, used to construct the URL for thumbnails in
-#: the API responses. It should be the URL of the RERO-ILS instance, as
-#: the thumbnails are served by the same instance.
-RERO_INVENIO_THUMBNAILS_URL = os.environ.get("RERO_INVENIO_THUMBNAILS_URL", RERO_ILS_URL)
+#: the API responses. Always mirrors ``RERO_ILS_URL`` — do not override this
+#: directly. Set ``RERO_ILS_URL`` instead; ``finalize_app`` in
+#: ``rero_ils/modules/ext.py`` re-derives this value after all config sources
+#: (invenio.cfg, environment variables) have been applied.
+RERO_INVENIO_THUMBNAILS_URL = RERO_ILS_URL
```
jma, line 3518: Seems useless as the value is forced in the extension.

scripts/setup
```diff
@@ -393,37 +362,15 @@ then
 #     eval ${PREFIX} invenio reroils fixtures get_all_mef_records ${DOCUMENTS} ${CREATE_LAZY} ${ENQUEUE} -v
 fi
 
+info_msg "Pause cover URL queue worker to avoid race conditions"
+eval ${PREFIX} invenio reroils utils cover-url-queue-worker pause
+
 info_msg "- Documents: ${DOCUMENTS} ${CREATE_LAZY} ${DONT_STOP}"
 eval ${PREFIX} invenio reroils fixtures create --pid_type doc --schema 'https://bib.rero.ch/schemas/documents/document-v0.0.1.json' ${DOCUMENTS} --append ${CREATE_LAZY} ${DONT_STOP}
 
-if ${CREATE_ITEMS_HOLDINGS_SMALL}
```
jma, line 399: This is not a part of the commit message.

jma (approved):

rero_ils/modules/cli/utils.py
rerowep: used as callback in cli commands

rero_ils/modules/documents/tasks.py
rerowep: It's only called from the extension — the CLI path goes through a separate Celery task (add_cover_urls). So the question is fair.

The separation exists because enqueue_cover_url_update contains non-trivial infrastructure logic: it writes to a Redis set (for deduplication), checks a lock key, checks a pause flag, and conditionally fires a Celery task. That belongs in tasks.py where the Redis helpers and Celery tasks live, not in the extension which should only know about record lifecycle hooks.

If you inlined it, add_cover_url.py would need to import Redis directly, know about the queue/lock/pause key names, and call Celery — all infrastructure details that are properly encapsulated in tasks.py. The extension stays thin and testable: its tests mock enqueue_cover_url_update at the boundary, while tasks.py tests cover the queueing logic independently.

rero_ils/config.py
rerowep: This Variable comes from rero-invenio-thumbnails and it is good to have here with the command to understand.

scripts/setup
rerowep: added

PR #4079, opened 2026-04-20
fix(broker): harden RabbitMQ and Redis connections

rerowep:
* upgrade RabbitMQ from 3.x to 4.2-management in docker-services.yml
* add healthcheck for mq service
* add CELERY_BROKER_HEARTBEAT=60 and CELERY_BROKER_CONNECTION_RETRY_ON_STARTUP to detect stale connections
* add conn.ensure_connection() in process_bulk_queue to recover from dead pool connections after worker restarts
* remove unnecessary celery.connection() calls in tasks and cli/index; use unbound Queue instead to avoid connection leaks
* initialize all Redis clients once at app startup in ext.py via RERO_ILS_MONITORING_REDIS_INSTANCES config
* replace per-call Redis.from_url() in messages, imports and monitoring with app-level shared clients
* expose all configured Redis instances in monitoring API
* add test for /monitoring/redis endpoint

### DIFF
--- a/rero_ils/modules/api.py
+++ b/rero_ils/modules/api.py
@@ -578,6 +578,7 @@ class IlsRecordsIndexer(RecordIndexer):
             successful and a list of error responses
         """
         with current_celery_app.pool.acquire(block=True) as conn:
+            conn.ensure_connection(max_retries=current_app.config.get("CELERY_BROKER_CONNECTION_MAX_RETRIES", 10))
             consumer = Consumer(
                 connection=conn,
                 queue=self.mq_queue.name,
--- a/rero_ils/modules/ext.py
+++ b/rero_ils/modules/ext.py
@@ -39,6 +39,7 @@ from invenio_records.signals import (
 from invenio_records_rest.errors import JSONSchemaValidationError
 from invenio_search import current_search_client
 from jsonschema.exceptions import ValidationError
+from redis import Redis
 
 from rero_ils.filter import (
     address_block,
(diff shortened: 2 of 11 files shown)


PascalRepond (commented):
I don't know enough about Redis and RabbitMQ to review this PR. Should we wait until after v1.28.0 to merge? Bumping RabbitMQ to a next major version and changes in the cache and queues behaviour should probably be thoroughly tested before deploying to production.

rero_ils/modules/api.py
```diff
@@ -578,6 +578,7 @@ def process_bulk_queue(self, search_bulk_kwargs=None, stats_only=True):
             successful and a list of error responses
         """
         with current_celery_app.pool.acquire(block=True) as conn:
```
jma, line 580: Invenio use invenio-queues: https://github.com/inveniosoftware/invenio-queues/blob/7ec1c68c2a3461c4aab71650252b47e86d2cc896/invenio_queues/queue.py#L96. What is the best approach?

rero_ils/modules/ext.py
```diff
@@ -213,6 +214,18 @@ def init_app(self, app):
         NormalizerStopWords(app)
         self.init_config(app)
         app.extensions["rero-ils"] = self
+        instances = {}
+        for name, config_key in app.config.get("RERO_ILS_MONITORING_REDIS_INSTANCES", {}).items():
+            url = app.config.get(config_key, "")
+            if url and url.startswith(("redis://", "rediss://", "unix://")):
+                instances[name] = Redis.from_url(url)
+        app.extensions["rero_ils_redis_instances"] = instances
+        if "import_cache" in instances:
```
jma, line 223: Can you explain why you have two cases here?

jma (approved):

Issue #4091, opened 2026-05-07
Allow seeing and editing local fields for any resource in the organisation


PascalRepond:
### How it works

In the holdings/items views, since 1.28.0, the UI hides the local fields tabs if the resource doesn't belong to the user's currentLibrary. But local fields belong to the organisation, not the library. Some librarians manage multiple libraries and they always have to switch libraries to see the local fields or edit them on items/holdings that don't belong to their current library.

### Improvement suggestion

1. On items and holdings, I can always see the local fields tab.
2. I can only create/edit/delete local fields if the linked resources belong to one of my manageable libraries.
3. [nice-to-have] This logic should be in the backend permissions and not in the UI, for API/UI consistency.

PR #4092, opened 2026-05-07
fix(local_fields): permission is dependent of the linked resource

PascalRepond:
- Partially addresses #4091.


PascalRepond:
Needs a PR to adapt rero-ils-ui

PR #4114, opened 2026-06-10
fix(ext): return 409 on concurrent PUT instead of 500

### LINKED ISSUE #3652: DB Integrity error when double-saving a user
### Bug description:

If I create a user and click on "save" (or Enter) multiple times, the systems tries to save the record twice and encounters a database error.

### Expected behavior:

There should be a security in the user editor (same as others in #3540) that prevents multiple button clicks to save a resource multiple times. 

### Steps to Reproduce:

- Create a user
- Multi-click on the "save" button
- Get an error in Sentry "IntegrityError api_users.users_list -> duplicate key value violates unique constraint "uq_accounts_user_username"

jma:
SQLAlchemy optimistic locking raises StaleDataError when two requests
try to commit the same record revision. invenio-records-rest's pass_record
decorator intercepts this as PIDResolveRESTError before app-level handlers
can catch it. Two error handlers are registered as a temporary workaround
until a fix lands upstream in invenio-records-rest.

- Closes https://github.com/rero/rero-ils/issues/3652.

Depends on https://github.com/rero/rero-ils-ui/pull/1476

Co-Authored-by: Johnny Mariéthoz <johnny.mariethoz@rero.ch>

### DIFF
--- a/rero_ils/modules/ext.py
+++ b/rero_ils/modules/ext.py
@@ -22,7 +22,7 @@ import logging
 
 import jinja2
 from elasticsearch_dsl import connections
-from flask import Blueprint
+from flask import Blueprint, jsonify, request
 from flask_bootstrap import Bootstrap4
 from flask_login import current_user
 from flask_login.signals import user_loaded_from_cookie, user_logged_in, user_logged_out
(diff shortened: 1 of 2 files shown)


jma:
@coderabbitai review

anouillz (approved):

PascalRepond:
## Tests

- [x] [not blocking] There is still a problem on "create" with multiple clicks, for example in acquisition receipts [screen](https://github.com/user-attachments/assets/b0824d3c-6118-4214-94c0-4706b6082af0)
- [x] Multiple clicks on checkout breaks everything [screen](https://github.com/user-attachments/assets/26882e42-e700-4671-8f36-a33775950f6e), there are still multiple loans created, but they are now not shown in sentry and not in the patron profile
- [x] error and multiple clicks still possible on "send order", [screen](https://github.com/user-attachments/assets/a675a5f3-9bf5-43e8-a417-2524f9540f65)

PR #4138, opened 2026-06-25
fix(homepage): remove links to deleted templates

PascalRepond:


rero_ils/theme/templates/rero_ils/frontpage.html
jma, line 44: Should not be hardcoded.

jma (approved):

rerowep (approved):

PR #4139, opened 2026-06-25
feat(patrons): add available languages to settings

Garfield-fr:
* Expose the list of available UI languages in the logged-user endpoint so the frontend can build a language picker without needing a separate request.
* English is always prepended as the default entry; the rest comes from I18N_LANGUAGES in the app config.
* Restructure user data under a dedicated `user` key instead of spreading it directly into the response object.


Garfield-fr:
@coderabbitai review

jma (approved):

PR #4144, opened 2026-06-25
feat(stats): add item_type distribution for number of items

### LINKED ISSUE #4143: Stat reports: add item_type distrib for catalog/number of items
### Improvement suggestion

The field "item_type" from the patron resource should be available as a distribution in the statistics reports concerning the number of items. In the reports, it should be displayed as `item_type.name`

PascalRepond:
Also add the test coverage that was missing for the patron local_codes distribution feature, both at the report level (number of patrons and circulation) and in the configuration JSON schema validation.

Closes #4143.

### DIFF
--- a/rero_ils/modules/item_types/api.py
+++ b/rero_ils/modules/item_types/api.py
@@ -43,6 +43,15 @@ class ItemTypesSearch(IlsRecordsSearch):
 
         default_filter = None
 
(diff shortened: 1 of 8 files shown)


rerowep (commented):
Are there any scripts we have to run in migration for stat_cfg-v0.0.1.json‎ changes?

PR #4149, opened 2026-06-30
fix: error message when saving user

### LINKED ISSUE #3498: User editor: wrong error message when we try to save a user without specifying library
### Bug description:

When we add and save a new user with librarian roles and leave the library field empty, a wrong error message is displayed : 'forbidden'.
If we edit an existing user and save without the library, the message is 'missing libraries'.

image

### Expected behavior:

The message should be 'missing libraries'.

### Steps to Reproduce:

1. Add a new user.
2. Select roles.
3. Fill all the required fields
4. Leave the library field empty. Save

### Context

- Versions: v.1.19.0, production...
- Browser:

anouillz:
* Skips library-scope permission check when no libraries are given.
* Keeps the base create / update permission check.
* Lets patron validation return "Missing libraries" instead of "Forbidden".
* Closes #3498

### DIFF
--- a/rero_ils/modules/patrons/permissions.py
+++ b/rero_ils/modules/patrons/permissions.py
@@ -48,14 +48,26 @@ class AllowedByActionRestrictStaffByManageableLibrary(AllowedByActionRestrictByO
         if not isinstance(record, Patron):
             record = Patron(record)
 
-        record_roles = record.get("roles", [])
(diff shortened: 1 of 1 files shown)


anouillz:
@coderabbitai full review

rerowep (approved):

PascalRepond (approved):

PR #4151, opened 2026-07-01
fix(documents): allow delete for rolled-over orders

PascalRepond:
At rollover, orders of the deactivated budget keep their open status, so a document linked only to those lines could never be deleted. Ignore order lines whose account belongs to an inactive budget when computing the document deletion blockers.

* Filter open order lines on their account `is_active` flag, reusing the account index already kept in sync with the budget.


PascalRepond:
@coderabbitai review

anouillz (approved):

Garfield-fr (approved):

PR #4156, opened 2026-07-06
feat: migrate to Font Awesome Free v7

PascalRepond:


rerowep (approved):

Garfield-fr (approved):
To ensure consistency in the icons, shouldn't we use only `fa-solid`?
Should we change the `fa-regular` ones?

PascalRepond:
> To ensure consistency in the icons, shouldn't we use only `fa-solid`? Should we change the `fa-regular` ones?

@Garfield-fr We could if we decide we really only want solid icons, but certain icons are available for free and a little bit more elegant in regular (trash-can, file, etc.) so I think it's OK to mix and match for best final effect.

PR #4158, opened 2026-07-08
fix(ill): validate form before confirmation

### LINKED ISSUE #3725: Public ILL request form: better validation management
### How it works

When one fills in an ILL request from the public UI, there are modals:

1. To warn that the document is present in my organisation (conditionnal)
2. To warn that the request may include fees (always)

These are displayed even when the form is not valid. Only after are the probelmatic fields highlighted. This might lead the user to believe the form has been sent correctly. 

### Improvement suggestion

To be decided based on how this form works.

- On clicking "save", validate the form (wihtout saving it) before showing the modals. Show modals only if the form is valid and ready to be sent.
- OR Disable the "Save" button if the form is not valid.
- Nice-to-have: Add an alert message if the form is not valid.

anouillz:
* Prevents confirmation modal from opening when form is invalid.
* Closes #3725

### DIFF
--- a/rero_ils/theme/assets/js/reroils/ills_request.js
+++ b/rero_ils/theme/assets/js/reroils/ills_request.js
@@ -6,6 +6,11 @@ import $ from 'jquery';
 $(function() {
   $('#submit-btn').on('click', function(e) {
     e.preventDefault();
+    const form = document.getElementById('ill-public-form');
+    if (!form.checkValidity()) {
+      form.submit();
+      return;
+    }
     const url = $('#source-url').val();
     if (url) {
       const wHost = window.location.host;


anouillz:
@coderabbitai review

rero_ils/theme/assets/js/reroils/ills_request.js
```diff
@@ -6,6 +6,11 @@ import $ from 'jquery';
 $(function() {
   $('#submit-btn').on('click', function(e) {
     e.preventDefault();
+    const form = $('#ill-public-form')[0];
```
Garfield-fr: An ID is unique within a page. You don't need to use [0] to select the form.

Garfield-fr (changes requested):

Garfield-fr (approved):

PascalRepond (approved):

PR #4162, opened 2026-07-14
feat(documents): flag incomplete indexing

### LINKED ISSUE #4071: Files: inform the user when the full-text is too large to be indexed
### How it works

When indexing a document, there is a [limit](https://github.com/rero/rero-ils/blob/27d8481531fc2ce8ef6e82bffc4238ca261852d0/rero_ils/modules/documents/dumpers/indexer.py#L227) to the size of the files' full-text that can be indexed (for performance reasons). The user may not be aware of this limit and is not informed when it is reached.

### Improvement suggestion

- [ ] Explore the possibility to raise the limit in production without losing performance and allowing most document with many files to be indexed. (e.g. https://bib.rero.ch/professional/records/documents/detail/344871)
- [ ] When this `if` (https://github.com/rero/rero-ils/blob/27d8481531fc2ce8ef6e82bffc4238ca261852d0/rero_ils/modules/documents/dumpers/indexer.py#L244) is not true, there should be an info (maybe in the document metadata) to say that fulltext is too large to be indexed and it should be displayed somewhere in the document detailed view (public + pro).
- [ ] Document the fulltext size limit in the [ILS documentation](https://bib.rero.ch/help/record-files/)

anouillz:
* Adds indicator when attached file content exceeds the indexing limit.
* Closes #4071

### DIFF
--- a/rero_ils/modules/documents/dumpers/indexer.py
+++ b/rero_ils/modules/documents/dumpers/indexer.py
@@ -211,6 +211,7 @@ class IndexerDumper(Dumper):
         files = []
         full_text_size = 0
         full_text_size_max = current_app.config.get("RERO_ILS_FILES_FULL_TEXT_MAX", 10 * 1024 * 1024)
+        fulltext_indexing_incomplete = False
         for record_file in record.get_records_files():
             record_files_information = {}
             collections = record_file.get("metadata", {}).get("collections")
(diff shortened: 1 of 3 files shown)


PascalRepond (approved):

Garfield-fr (approved):

rerowep (approved):

PascalRepond:
Since we decided to refactor the whole file indexing mechanism, this fix won't be necessary.

PR #4165, opened 2026-07-15
fix(entities): link MEF by identifier, not type

### LINKED ISSUE #4135: Link to MEF task: do not check entity type
### How it works

Currently, a task links textual entities with identifiers to MEF. This task checks if the entity_type in the ILS data corresponds to the entity type in MEF. This is not needed.

### Improvement suggestion

When linking to a MEF record by identifier, do not take the entity type into account, the identifier should be enough. 

Warning: in the `contribution` if the MEF entity is the wrong type, it should still raise a validation error and be signaled to metadata admin.

rerowep:
* replace.py: identifier alone now links contribution/subjects/ genreForm entries to MEF; search falls back across every MEF family allowed for the field when the locally-declared type is wrong
* replace.py: contribution type mismatches are tracked in a new `type_mismatch` dict, no longer conflated with the `rero_only` retry cache, which was silently blocking later documents sharing the same identifier
* replace.py: document query now derives allowed entity types from `RERO_ILS_ENTITIES_TYPES_FIELDS`, so `bf:Work`-only subjects are no longer skipped from scanning
* sync.py: `sync_record` recursion is now capped to guard against a pathological MEF pid cycle; raises specific `RuntimeError`s instead of the bare `Exception` class
* utils.py: `get_mef_data_by_type` reports a clear "no result" error instead of a confusing IndexError-derived message
* proxy.py: `search()` no longer forwards cookies or arbitrary headers to the remote MEF server; deduplicated `_get_query_params` and removed a dead `_post_process_result_hit` override
* sync.py, replace.py: switched logging calls to lazy %-style formatting instead of f-strings
* cli.py: stopped reaching into a private method from outside the class
* Closes https://github.com/rero/rero-ils/issues/4135

### DIFF
--- a/rero_ils/modules/entities/remote_entities/replace.py
+++ b/rero_ils/modules/entities/remote_entities/replace.py
@@ -19,37 +19,63 @@ from rero_ils.modules.utils import set_timestamp as utils_set_timestamp
 from ..logger import create_logger
 from .api import RemoteEntity
 
+# (connect, read) timeout in seconds for each MEF request, so a stalled
+# connection cannot hang the whole run scan.
+MEF_REQUEST_TIMEOUT = (5, 60)
+
 
 class ReplaceIdentifiedBy:
-    """Entity replace identifiedBy with $ref.
+    """Replace a document's ``identifiedBy`` entities with a MEF ``$ref``.
 
-    Usage example:
-    replace_identified_by = ReplaceIdentifiedBy(
-        field='contribution',
-        dry_run=True,
-        verbose=True
-    )
-    changed, not_found rero_only = replace_identified_by.run()
-    # changed: count of changed identifiedBy
-    # not_found: count of not found identifiedBy
-    # rero_only: count of identifiedBy with RERO reference
-    #            (therefore not changed).
+    For a given field (``contribution``, ``subjects`` or ``genreForm``), every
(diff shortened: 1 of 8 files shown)


rero_ils/modules/entities/remote_entities/replace.py
```diff
@@ -19,37 +19,63 @@
...
+        replace = ReplaceIdentifiedBy(field="contribution", dry_run=True, verbose=True)
+        changed, not_found, rero_only = replace.run()
     """
 
     fields = ("contribution", "subjects", "genreForm")
+    # MEF families allowed for each field, per the document JSON schema.
+    field_mef_types = {
+        "contribution": ("agents",),
+        "subjects": ("agents", "concepts", "places"),
+        "genreForm": ("concepts",),
```
PascalRepond, line 59: Maybe larger question: usually in genreForms only concepts with a specific subfield are allowed (`"bnf_type": "genre/forme Rameau"`). Should we also enforce this for the replace?

PascalRepond (approved):

jma (approved):
To be honest I'm not sure of what this PR do. I found many debug message modified. Only some header type are forwarded to MEF and a max recursion value has been added.

PascalRepond:
Add "Closes https://github.com/rero/rero-ils/issues/4135"

PR #4167, opened 2026-07-16
fix(documents): combine multiple new_acq filters

### LINKED ISSUE #3371: Allow multiple location/library/organisation to be combined with `new_acquisition` filter
### How it works
When using a `new_acquisition` filter, existing filters location, library and organisation can be combined. However, only one of each can be used. 
This means we cannot search for new_acquisition from two or more different locations. Example: `new_acquisition=now-3M%2Fd&location=367&location=368` only looks in location 367.

### Improvement suggestion
Allow to combine multiple of the same filter with the `new_acquisition` filter. 

See: 
https://github.com/rero/rero-ils/blob/staging/rero_ils/modules/documents/query.py#L66-L71

Garfield-fr:
Repeating location/library/organisation with new_acquisition only kept the first value (request.args.get instead of getlist). Use a terms query so repeated values combine with OR, matching the standalone post-filters behavior.

* getlist() collects all repeated query param values
* Q("terms", ...) ORs them into the nested acquisition query
* Closes #3371.

### DIFF
--- a/tests/api/documents/test_documents_rest.py
+++ b/tests/api/documents/test_documents_rest.py
@@ -111,6 +111,8 @@ def test_documents_newacq_filters(
     system_librarian_martigny,
     rero_json_header,
     document,
+    document_sion_items,
+    export_document,
     holding_lib_martigny,
     holding_lib_saxon,
     loc_public_saxon,
@@ -150,7 +152,7 @@ def test_documents_newacq_filters(
     assert res.status_code == 201
 
     # check item creation and indexation
-    doc_list = url_for("invenio_records_rest.doc_list", view="global", pid="doc1")
+    doc_list = url_for("invenio_records_rest.doc_list", view="global", q="pid:doc1")
     res = client.get(doc_list, headers=rero_json_header)
     data = get_json(res)
     assert len(data["hits"]["hits"]) == 1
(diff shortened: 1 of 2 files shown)


rerowep (approved):

tests/api/documents/test_documents_rest.py
```diff
@@ -216,25 +216,29 @@ def datetime_milliseconds(date):
     data = get_json(res)
     assert data["hits"]["total"]["value"] == 0
 
-    # check new_acquisition filters with -- separator and timestamp
-    # Ex: 1696111200000--1700089200000
+    #   --> for loc1 or loc3, there is 1 document with 2 new acquisition
+    #       items (multiple `location` values are combined with OR)
```
PascalRepond, line 222: This test passes even without your fix.

We want to make sure that a query with multiple location filters will return both documents if one has a new_acq for loc1 and the other has a new_acq for loc3, for example:

doc1: new-acq in loc1 only
doc2: new-acq in loc3 only

FILTER: location=loc1               -> {doc1}        (1)
FILTER: location=loc3               -> {doc2}        (1)
FILTER: location=loc1&location=loc3 -> {doc1, doc2}  (2) (would fail without the change)

PascalRepond (changes requested):

tests/api/documents/test_documents_rest.py
Garfield-fr: Tests added

PascalRepond (approved):

PascalRepond (approved):

jma (approved):

PR #4169, opened 2026-07-16
fix(facet-count): count library facets by document

### LINKED ISSUE #4003: The organisation/library facet counts are incorrect
### How it works

Since https://github.com/rero/rero-ils/pull/3961, the counts in the library facet use nested aggregations, which means they count the number of holdings in each facet instead of the number of documents. This adds a discrepancy : [screenshot](https://github.com/user-attachments/assets/4ff27a6d-0e56-40fc-9401-3c02ce3deae4)

### Improvement suggestion

Decide how to go about this problem:

1. Complexity high:  Revert the nested aggregations implementation and add a simple field to the index that stores a set of pid combination organisation_library_location and is updated on each that allows us to dynamically pre-filter the org/library facet content.
2. Complexity medium: Revert the nested aggregations implementation and go back to the previous pre-filter implementation (but lose the ability to see all locations in the facet)
3. Complexity low: Transform the organisation facet into a separate filter that uses nested aggregation but doesn't need counts.

anouillz:
* Replaces nested facet aggregations with indexed organisation library location paths.
* Stores unique pid combinations in organisation_library_location.
* Uses those paths to count documents instead of holdings.
* Rebuilds the same facet tree for the UI.
* Replaces https://github.com/rero/rero-ils-ui/pull/1508 and https://github.com/rero/ng-core/pull/831.
* Closes #4003

Note: this change adds new indexed fields for document facet counts. Existing records need to be reindexed for the corrected organisation/library/location counts to be available.

### DIFF
--- a/rero_ils/config.py
+++ b/rero_ils/config.py
@@ -2114,38 +2114,27 @@ RECORDS_REST_FACETS = dict(
                     bool=dict()
                 ),
                 aggs=dict(
-                    nested_holdings=dict(
-                        nested=dict(
-                            path="nested_holdings"
-                        ),
-                        aggs=dict(
-                            organisation=dict(
-                                terms=dict(
-                                    field="nested_holdings.organisation.organisation_pid",
-                                    size=DOCUMENTS_AGGREGATION_SIZE,
-                                    min_doc_count=0,
-                                ),
-                                aggs=dict(
-                                    library=dict(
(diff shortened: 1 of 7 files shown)


anouillz:
Issue #4003 fixed in PR https://github.com/rero/rero-ils/pull/4180

PR #4186, opened 2026-07-27
build: update dependencies

PascalRepond:
Ruff 0.16 ships a much broader default rule set, which surfaced 332 new errors. Adapt the code and the linter configuration.

* Fixes the defects found on the way: `raise StopIteration` in a generator, `__all__` as a string, a `@property` on a method taking an argument, and dead test assertions.
* Retires 14 rules from the ignore lists by fixing their violations.
* Ignores `BLE001` and `DTZ`, deliberate patterns in the code base.

Updated dependencies (53 packages).

### DIFF
--- a/rero_ils/dojson/cli.py
+++ b/rero_ils/dojson/cli.py
@@ -17,9 +17,7 @@ def reverse():
     """Reverse the order of the data."""
 
     def processor(iterator):
-        items = []
-        for item in iterator:
-            items.append(item)
+        items = list(iterator)
         items.reverse()
         return items
 
@@ -35,11 +33,9 @@ def head(max):
     """Take only the first max items."""
 
     def processor(iterator):
-        n = 0
-        for item in iterator:
+        for n, item in enumerate(iterator):
             if n >= max:
-                raise StopIteration
-            n += 1
+                return
             yield item
 
     return processor
--- a/rero_ils/modules/api.py
+++ b/rero_ils/modules/api.py
@@ -378,10 +378,10 @@ class IlsRecord(Record):
             persistent_identifier.delete()
             if force:
                 db.session.delete(persistent_identifier)
-            self = super().delete(force=force)
+            record = super().delete(force=force)
             if dbcommit:
                 db.session.commit()
-            return self
+            return record
         raise IlsRecordError.NotDeleted()
 
     def update(self, data, commit=False, dbcommit=False, reindex=False):
@@ -429,10 +429,10 @@ class IlsRecord(Record):
         persistent_identifier = self.get_persistent_identifier(self.id)
         if persistent_identifier.is_deleted():
             raise IlsRecordError.Deleted()
-        self = super().revert(revision_id=revision_id)
+        record = super().revert(revision_id=revision_id)
         if reindex:
-            self.reindex(forceindex=False)
-        return self
+            record.reindex(forceindex=False)
+        return record
 
     def undelete(self, reindex=False):
         """Undelete the record."""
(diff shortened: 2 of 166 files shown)


rero_ils/dojson/cli.py
```diff
@@ -35,11 +33,9 @@ def head(max):
     """Take only the first max items."""
 
     def processor(iterator):
-        n = 0
-        for item in iterator:
+        for n, item in enumerate(iterator):
             if n >= max:
-                raise StopIteration
-            n += 1
+                return
```
rerowep, line 38: Why not raising a StopIteration, only returning will never show exhausted lists.

rero_ils/modules/stats/tasks.py
```diff
@@ -61,10 +61,6 @@ def collect_stats_reports(frequency="month"):
...
             report = stat_report.create_stat(values)
             to_return.append(report.pid)
-        except Exception as error:
-            logger.error(
-                f"Unable to generate report from config({pid}) :: {error}",
-                exc_info=True,
-                stack_info=True,
-            )
+        except Exception:
+            logger.exception(f"Unable to generate report from config({pid})", stack_info=True)
```
rerowep: If i remember right logger schould not us python f strings bur %s. A verify elsewhere.

rero_ils/modules/api.py
```diff
@@ -635,11 +635,10 @@ def _actionsiter(self, message_iterator):
                 message.reject()
             except Exception:
                 message.reject()
-                current_app.logger.error(
+                current_app.logger.exception(
```
rerowep, line 638: see above

rero_ils/modules/api.py
```diff
@@ -635,11 +635,10 @@ def _actionsiter(self, message_iterator):
...
                     f"Failed to {payload['op']}"
                     f" {payload.get('doc_type'), 'rec'} "
                     f"{payload.get('pid')}:{payload.get('id')}",
-                    exc_info=True,
```
rerowep, line 642: without exc_info we will not get traceback info and it could be difficult to find the exact error possition.

rero_ils/modules/jsonresolver.py
```diff
@@ -16,9 +16,9 @@ def resolve_json_refs(pid_type, pid):
     """
     try:
         persistent_id = PersistentIdentifier.get(pid_type, pid)
-    except Exception as error:
+    except Exception:
         current_app.logger.error(f"Unable to resolve {pid_type} pid: {pid}")
```
rerowep: see above

rerowep (changes requested):

rero_ils/modules/api.py
PascalRepond: exc_info is True by default for logger.exception.

rero_ils/modules/stats/tasks.py
PascalRepond: You are right but the ruff rule is not currently active and all loggers currently use f-strings. I fixed those specfic ones but globally this should be refactored as its own fix (maybe open an issue?).

rero_ils/dojson/cli.py
PascalRepond: This seems to be the correct practice since PEP479, see : https://docs.astral.sh/ruff/rules/stop-iteration-return/

Raising `StopIteration` inside a generator raises a `RuntimeError`. To raise a `StopIteration` at the consumer, we need to use a bare `return`. It stops the iterator inside the generator, the caller still sees an exhausted iterator, and the item count is unchanged.

rerowep (approved):

PR #4204, opened 2026-08-05
refactor(records-rest): drop workarounds fixed upstream

PascalRepond:
invenio-records-rest 6.0.2 handles concurrent PUT and weak ETags itself, so two of the three workarounds added for #3652 can go:

* the If-Match weak ETag normalization: check_etag() is now called with weak=True on every verb.
* the PIDResolveRESTError handler: pass_record no longer wraps the handler call, so a StaleDataError cannot reach it.

The StaleDataError handler stays. Upstream catches it in put() only, and leaving delete() out is an oversight, so a concurrent delete would still answer 500. It now reuses upstream's RecordConflictRESTError so that both verbs share the same 409 body.

Tests keep only the two behaviours worth checking locally: the 409 built in ext.py, and the weak ETag precondition, whose regression would silently allow concurrent overwrites.

invenio-records-rest is pinned to >=6.0.2 accordingly.


jma (approved):
Do not forget to test the UI interface.

PR #4218, opened 2026-08-18
feat(items): export temporary location name

### LINKED ISSUE #4212: Inventory list: serialise the temporary_location for the CSV/XLS exports
### How it works

The inventory list for exporting item list doesn't export the temporary item types.

### Improvement suggestion

Serialise the temporary location name (same as in the location serialiser), and add the column `item_temporary_location_name` to the exports, following `item_location_name`.

anouillz:
* Add temporary location names to inventory CSV and XLSX exports.
* Closes #4212

### DIFF
--- a/tests/api/items/test_items_serializer.py
+++ b/tests/api/items/test_items_serializer.py
@@ -94,6 +94,7 @@ def test_serializers(
         "item_type",
         "item_library_name",
         "item_location_name",
+        "item_temporary_location_name",
         "item_pac_code",
         "item_holding_pid",
         "item_price",
--- a/rero_ils/modules/items/serializers/__init__.py
+++ b/rero_ils/modules/items/serializers/__init__.py
@@ -51,6 +51,7 @@ _csv = ItemCSVSerializer(
         "item_type",
         "item_library_name",
         "item_location_name",
+        "item_temporary_location_name",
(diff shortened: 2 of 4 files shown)


rerowep (approved):

tests/api/items/test_items_serializer.py
```diff
@@ -135,6 +140,12 @@ def test_serializers(
     for field in fields:
         assert field in data
 
+    rows = list(DictReader(StringIO(data)))
+    item_data = next(row for row in rows if row["item_pid"] == item.pid)
+    assert item_data["item_temporary_location_name"] == loc_restricted_martigny["name"]
+    headers = list(rows[0])
+    assert headers.index("item_temporary_location_name") == headers.index("item_location_name") + 1
+
```
PascalRepond: For consistence, since we don't test the rows themselves for other fields, I don't think we need to test that for this one. 
`assert field in data` is enough, otherwise we would need to test every serialised field.

PascalRepond (approved):

tests/api/items/test_items_serializer.py
anouillz: Okay, I understand and so I just added the column name to the list so it can have the same tests as the others. Thank you

PR #4225, opened 2026-08-20
fix(patrons): drop empty entries in record links

rerowep:
A resource without any linked pid was still reported by `get_links_to_me(get_pids=True)`: the value tested for truthiness was a lazy object, and such an object is always true.

* patrons: the transaction pids are yielded and an exhausted generator is still truthy, so `transactions` was always present, holding the generator itself. Build the query locally and resolve it with `sorted_pids`, as the three other link types already do. The count branch reuses that query, which is what `get_transactions_count_for_patron` did.
* acq_orders: the empty entries were dropped before the queries were resolved, and an `elasticsearch_dsl` query defines neither `__bool__` nor `__len__`. Resolve the pids first, then drop the entries left empty.

The counted variant was already correct in both cases, so `reasons_not_to_delete` and the deletion rules do not change.

### DIFF
--- a/rero_ils/modules/patrons/api.py
+++ b/rero_ils/modules/patrons/api.py
@@ -27,11 +27,7 @@ from rero_ils.modules.minters import id_minter
 from rero_ils.modules.operation_logs.extensions import OperationLogObserverExtension
 from rero_ils.modules.organisations.api import Organisation
 from rero_ils.modules.patron_transactions.api import PatronTransaction
-from rero_ils.modules.patron_transactions.utils import (
-    create_subscription_for_patron,
-    get_transactions_count_for_patron,
-    get_transactions_pids_for_patron,
-)
+from rero_ils.modules.patron_transactions.utils import _build_transaction_query, create_subscription_for_patron
 from rero_ils.modules.providers import Provider
 from rero_ils.modules.tasks import process_bulk_queue
 from rero_ils.modules.templates.api import TemplatesSearch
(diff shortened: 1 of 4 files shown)


rero_ils/modules/patrons/api.py
```diff
@@ -334,14 +330,17 @@ def get_links_to_me(self, get_pids=False):
         )
         template_query = TemplatesSearch().filter("term", creator__pid=self.pid)
         ill_query = ILLRequestsSearch().filter("term", patron__pid=self.pid)
+        transaction_query = (
+            PatronTransactionsSearch().filter("term", patron__pid=self.pid).filter("term", status="open")
+        )
         if get_pids:
             loans = sorted_pids(loan_query)
-            transactions = get_transactions_pids_for_patron(self.pid, status="open")
```
PascalRepond, line 339: You could clean up these helpers from patron_transactions.utils since you don't use them anymore.

rero_ils/modules/patrons/api.py
```diff
@@ -334,14 +330,17 @@ def get_links_to_me(self, get_pids=False):
...
+            transactions = sorted_pids(transaction_query)
             templates = sorted_pids(template_query)
             ill_requests = sorted_pids(ill_query)
         else:
             loans = loan_query.count()
-            transactions = get_transactions_count_for_patron(self.pid, status="open")
```
PascalRepond, line 344: Same as above

rero_ils/modules/patrons/api.py
```diff
@@ -334,14 +330,17 @@ def get_links_to_me(self, get_pids=False):
         )
         template_query = TemplatesSearch().filter("term", creator__pid=self.pid)
         ill_query = ILLRequestsSearch().filter("term", patron__pid=self.pid)
+        transaction_query = (
```
PascalRepond: Make sure this does the exact same thing as the helper functions that were used before.

PascalRepond (approved):

jma (approved):

rero_ils/modules/patrons/api.py
rerowep: `get_transactions_pids_for_patron` is still used in selfcheck:257

PR #4226, opened 2026-08-20
perf(docker): speed up the dev and test stack

jma:
The development stack serves the test suite and `scripts/setup`, whose data is
always thrown away, so it can trade durability and production sizing for speed.

Three files, and nothing to change in the workflows or in `INSTALL.md`:
`docker compose up -d` is still the only command a caller runs, and it now
guarantees the whole stack by itself.

### Elasticsearch shards

The `record` template asks for 8 shards and 1 replica, sized for production. On
a development cluster that is hundreds of shards for a few thousand documents,
and every search fans out over 8 of them.

An `init` service puts a **legacy** template: it merges with `record` (order 0)
and overrides only these two settings, so every ICU/French analyzer is preserved.
A composable template would have disabled `record` entirely.

The test suite cannot rely on that service: the `search` fixture calls
`delete_template("*")` when a previous run left state behind, so
`tests/conftest.py` puts the template back itself, after each `put_templates`.

The shard count deliberately stays out of `scripts/setup`: it also runs on the
test servers, where the production defaults of `record` are the right ones.

### PostgreSQL durability

`fsync`, `synchronous_commit` and `full_page_writes` off, with
`wal_level=minimal`, which requires `max_wal_senders=0`. `autovacuum` stays on:
without it there is no ANALYZE, hence no statistics, and the queries fall back to
sequential scans.

These options are in `docker-compose.yml` and **not** in `docker-services.yml`,
which is the base of the `docker-compose.full.yml` production example: an example
carrying a database without durability is a trap left behind for whoever copies
it.

Trade-off: an unclean stop can corrupt the database, which then has to be
recreated by a new setup.

### The compose file is the gate

`docker compose up -d` returns as soon as the containers are *started*, not when
they answer, and not when a one-shot service has finished. So a setup launched
right after it could create its indices before the template lands, and keep the 8
shards. Measured three times out of three from a cold stack: `init` is still
running and `GET _template/dev-single-shard` returns **404** at the moment
`up -d` returns.

Two additions close that window inside the compose file, so that callers keep a
single command:

* **healthchecks on `db` and `cache`**, which had none — `mq` and `es` already
  have theirs — and `init` depends on all four being healthy;
* **a `gate` service** depending on `init` with
  `condition: service_completed_successfully`. `depends_on` conditions block a
  container's start, and `up -d` does not return before it has tried to start
  every service, so it now blocks until the template is installed.

Measured, from a cold stack:

| | |
| --- | --- |
| `up -d` | 8 s, exit 0, template **200** |
| template rejected by Elasticsearch | exit 1, `service "init" didn't complete successfully: exit 22` |
| a service that never turns healthy | exit 1 once its retries are exhausted, rather than hanging |

The `gate` reuses the image of `init` and runs `true`, so it pulls nothing extra
and leaves nothing running. `init` is named `init` rather than `es_init` because
it depends on the whole stack; it runs `curlimages/curl`, 10 MB, which takes
curl's arguments directly so no entrypoint override is needed — the Elasticsearch
image would have worked but weighs 353 MB, and none of the other images in the
stack ship curl.

Two mechanisms were tried first and are recorded in a comment so nobody
"simplifies" the gate away:

* **`docker compose wait init`** only considers *running* containers. Once `init`
  has exited — the normal case by the time a later step asks — it fails with
  `no containers for project "rero-ils"`. This is how the previous revision of
  this PR broke, and CI caught it.
* **`docker compose up -d --wait`** treats an exited one-shot container as a
  failure, returning 1 even when the container exited 0.

### One deletion

`docker/wait-for-services.sh` is removed. Nothing had called it since
`.travis.yml` disappeared in 2020, and the compose file now carries the
guarantee. For the record, reviving it would have meant fixing three defects that
made it ineffective anyway: `[[ $(...)==200 ]]` without spaces is not an equality
but a non-empty-string test, so the Elasticsearch and redis checks always passed;
`docker compose exec` without `-T` fails where there is no TTY, such as a CI
runner; and 5 retries at 2 s gave up long before a cold Elasticsearch answers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docker-compose.yml
```diff
@@ -18,3 +37,39 @@ services:
...
+    entrypoint: ["curl"]
+    command:
+      - "-sS"
+      - "-X"
+      - "PUT"
+      - "http://es:9200/_template/dev-single-shard"
+      - "-H"
+      - "Content-Type: application/json"
+      - "-d"
+      - '{"index_patterns":["*-*"],"order":100,"settings":{"number_of_shards":1,"number_of_replicas":0}}'
```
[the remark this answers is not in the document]
jma: Both points fixed in 0f3aacc7, but not in `scripts/setup`: it also runs on the test servers, where the production defaults of the `record` template are the right ones, so the shard count stays confined to the development stack.

**The readiness check** went into `docker/wait-for-services.sh`, which already gated postgres, Elasticsearch, redis and RabbitMQ — and which nothing had called since `.travis.yml` disappeared in 2020. It now also waits for `dev-single-shard`, and `INSTALL.md` plus both workflows call it. In the workflows the step sits just before the one that needs the services, not right after `up -d`, so the containers keep starting while uv, Python and the browsers install; the wait is then normally a no-op.

**`es_init` fails on HTTP errors**: `-sS` became `-fsS`.

Reviving that script surfaced three defects that would have made the gate useless:

* `[[ $(...)==200 ]]` without the spaces is not an equality — it tests that the word is not empty, so `[[ 404==200 ]]` is true. The Elasticsearch and redis checks always passed.
* `docker compose exec` without `-T` allocates a TTY, which fails outright on a CI runner.
* 5 retries at 2 s gave up after 10 s, while a cold Elasticsearch is allowed 80 s by its own compose healthcheck. Raised to 40.

Verified against a running stack: the four service checks pass, and the template check fails exactly as intended on a stack started before `es_init` existed.

docker-compose.yml
```diff
@@ -18,3 +37,50 @@ services:
...
       service: es
+    # Required to order es_init below. `_cluster/health` answers 200 whatever the
+    # state, yellow included, so the colour is not part of the test.
+    healthcheck:
+      test: ["CMD", "curl", "-sf", "http://localhost:9200/_cluster/health"]
+      interval: 5s
+      timeout: 5s
+      retries: 12
+      start_period: 20s
+  # The `record` template asks for 8 shards and 1 replica, sized for production.
```
rerowep, line 62: This command is already in test/conftest.py ???

INSTALL.md
```diff
@@ -61,6 +61,13 @@ screen ~/Library/Containers/com.docker.docker/Data/com.docker.driver.amd64-linux
 linut00001:~# sysctl -w vm.max_map_count=262144
 ```
 
+`docker-compose up -d` returns as soon as the containers are created, not when
+they answer. Wait for the services to be ready before going further:
+
+```console
+./docker/wait-for-services.sh
+```
+
```
PascalRepond: Not necessary here? The setup will fail if the containers aren't ready, and we just relaunch it.

PascalRepond (approved):

rerowep (approved):
Do we have to implement the healthcheck in kubernetes and docker stock?

INSTALL.md
jma: I've extended wait-for-services to be sure that the number of shards has been updated.

docker-compose.yml
jma: Yes we need both as we want to have this for the setup command for those that use this file locally.

Garfield-fr (approved):

PR #4233, opened 2026-08-25
chore: update dependencies

anouillz:
* Upgrade invenio-cache and Flask-Caching now that their compatibility issue is resolved.
* Update test configuration to use the new SimpleCache backend name required by Flask-Caching 2.5.
* Authenticate before serving cached responses so cached errors cannot
  block valid users and cached content cannot bypass access control.
* Verify authentication remains enforced when responses are cached.


PascalRepond (approved):
🚀

rerowep (approved):

PR #4239, opened 2026-08-31
refactor: remove legacy scripts and dead code

jma:
Nothing removed here has a caller left, save one wrapper that had one and no
work to do — its own section below. Most of the rest lost its last caller when
#4227 removed the dead options of `scripts/setup`, which is why this waited for
it. Now rebased on that merge, so the diff is this commit alone.

### Traced through the history, not just grepped

| removed | why |
| --- | --- |
| `load_virtua_transactions`, with `check_missing_fields` and `build_loan_record` as its only callees | loaded circulation transactions exported from Virtua; its `LOADEXTRAFILES` call site is gone |
| `create_operation_logs` | created in Feb 2021 as **`migrate_virtua_operation_logs`**, renamed three months later. It read the same Virtua export, `data/operation_logs.json` |
| `dump_operation_logs` | its counterpart, and broken: it wrote `str(record)` into a `JsonWriter`, whose `write` then calls `dumps()`. The output was a JSON array of **strings** holding Python reprs, which nothing could read back — not even `create_operation_logs` |
| `check_pid_dependencies`, `data/pid_dependencies_{big,small}.json` | only ever driven by `create_csvs -p` |
| `dump_search_mappings` | only ever driven by `setup --es-mapping`, the block that dumped the mappings before and after a setup |
| `create_request` | had callers until **June 2022**, when `create_loans` started routing `requests`, `rank_1` and `rank_2` through `create_loan`. Dead for four years |
| `scripts/create_csvs` | asks for the `cont` pid type, renamed `rement` long ago, and for `fixtures get_all_mef_records`, no longer registered. `set -e` aborts it before its last five `create_csv` calls |
| `scripts/dojson_virtua` | calls `dojson do marc21tojson_rero`; the registered rules are named `marc21_rero_to_json` and its five siblings, so it fails on an `AssertionError` |
| eight data files | the Virtua exports, the two `documents_*.xml` of a commented `marc21tojson` note, and the two pid-dependency files |

### Two manual tools

Neither is called by anything either, and neither has had a substantive change in
five years — only the mechanical sweeps, poetry to uv in 2025 and the license
headers in 2026.

| removed | last real change | what it did |
| --- | --- | --- |
| `scripts/all_year_days` | 2020-03-24 | replayed a command for every day of the year under `faketime`, to catch date-sensitive tests |
| `scripts/load_testing.py` | 2021-06-02 | hammered the REST API with concurrent updates |

### One wrapper with a caller but nothing to do

`scripts/console` is the exception to the rule above: it *was* called. Twenty-three
lines, `set -e` written twice, a decorative `PROGRAM:` banner, and one payload —
`invenio shell`. The poe task now runs that command directly, which is already
the house style for `format`, `lint`, `tests` and the four `pybabel` tasks.
`uv run poe console` is unchanged for the caller, so `INSTALL.md` needed no edit.

The only thing lost is a guard telling you to use `uv run`, which the documented
path already guarantees. Verified that the interactivity survives, since a REPL
behind a task runner is the one thing that could have broken: poe redirects
`stdin` only when a task declares input and `stdout` only when it captures, so a
`cmd` inherits the terminal — which is what `tests_debug = pytest -s -vv` already
relies on. `echo 'print("shell ok")' | uv run poe console` reaches IPython 9.16.1
and prints it.

`scripts/update` has the same copied header and is left alone: it chains
bootstrap, `alembic upgrade` and `update_mapping`, so it earns its existence.

### Kept, and one of them now says why

`destroy_operation_logs`, called by `scripts/setup_batches`. `create_csv`,
`bulk_load` and `bulk_save`, covered by `tests/unit/test_csv.py` and run by hand.
And `scripts/russian_dolls`, which has no caller either but was moved to pnpm
last month, so someone clearly uses it.

`scripts/convert_marc21xml.py` stays and gains a docstring, because it looked far
more misplaced than it is: the only Python module left among the shell scripts,
named as though it were a general MARC21 converter. It is the converter of the
`cogne` migration fixture, and it belongs exactly where it is:

* the production converters live in `rero_ils_migrations`, one module per
  campaign, resolved by dotted path from `Migration.conversion_code` — the two
  real migrations hold `converter_2024_bramois.BramoisConverter` and
  `converter_2025_rerodoc_vs.RerodocVSConverter`;
* that package is private and absent from PyPI, so it can never be a dependency
  of this repo, and the fixture needs a converter shipped here;
* it is not redundant with the same-named class in that package: `BaseConverter`
  declares `dedup` abstract, `converter_marc21xml_rero` never implements it, and
  the fixture runs `migrations data dedup cogne`. This one pairs that MARC21
  conversion with a working `dedup`;
* it sits outside `rero_ils/` because the wheel contains only `rero_ils/` —
  checked by building one — and fixture code has no business being shipped.

`scripts/` goes from 15 files to 10. Eight of the 10 have a verified caller: a
poe task, the E2E workflow, a `. source` from another script, or a test. The two
that do not are deliberate and named above — `__init__.py`, which is what makes
`scripts.convert_marc21xml` reachable by dotted path, and `russian_dolls`, kept
on the strength of its history rather than a caller.

After this, the only trace of Virtua left in the repository is in
`CHANGELOG.md`: seven lines of release notes, all of them Virtua migration bugs
closed between v1.2.0 and v1.4.0 in 2021. That check needs a precise pattern,
since
`git grep -i virtua` matches `virtual` in **14** files — virtualenv, virtual
memory, `VIRTUAL_ENV`, a `"virtual" Loan` in three tests. The one that means the
software is `git grep -inE 'virtua($|[^l])'`, and it returns nothing outside the
changelog.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


PascalRepond (approved):
🧹🌟

rerowep (approved):

Garfield-fr (approved):

Issue #4220, opened 2026-08-18
API: harmonise `hits.total` to a plain integer on all search endpoints


PascalRepond:
### How it works

`hits.total` has no single shape across the API. Four different shapes are served today:

| Shape | Where it comes from |
|---|---|
| `{"value": 42, "relation": "eq"}` | `rero_ils/modules/serializers/base.py:76` — our `serialize_search()` override forwards the raw search engine object. This covers every standard `RECORDS_REST_ENDPOINTS` list route. Same in `rero_ils/modules/sru/views.py:402` and `rero_ils/modules/imports/api.py:58`. |
| `{"value": 42}` — no `relation` | `rero_ils/modules/items/views/api_views.py:455` and `:477` |
| `42` — plain integer | `rero-invenio-files` and anything else built on `invenio-records-resources`, whose `total` is an integer by design (`invenio_records_resources/services/records/results.py:181`). Also `rero_ils/modules/imports/serializers/serializers.py:48`. |
| `"42"` — string | ES 6 era legacy. No current endpoint produces it, but the front-end still tolerates it. |

Two points worth noting:

**The plain integer is what upstream Invenio does, in both stacks.** `invenio-records-rest`'s own JSON serializer flattens the object (`invenio_records_rest/serializers/json.py:50`), and `invenio-records-resources` exposes `total` as an integer. Our `serialize_search()` override is the deviation, not the norm — `rero-mef`, which does not override it, already returns an integer.

**`relation` is not usable and is never used.** Both stacks set `track_total_hits=True` unconditionally (`invenio_records_rest/views.py:679`, `invenio_records_resources/services/records/service.py:165`), so the count is always exact and `relation` is always `"eq"`. It is read nowhere in `rero-ils-ui`, `ng-core` or `sonar-ui` — zero occurrences outside test fixtures — and two of our own endpoints already omit it, which would break any consumer that trusted the dict contract.

#### Consequences on the front-end

Because both shapes are live in the same application (standard resources return the dict, files resources return the integer), the front-end has to guess:

* `RecordService.totalHits()` in `ng-core` exists only to absorb the difference. It is called from **40 places** in `rero-ils-ui`, 30 of which have to prefix `+` to defeat its `number | string` return type.
* **25 further places** bypass it and read `result.hits.total.value` directly, silently yielding `undefined` on an integer total.
[...]

PR #4253, opened 2026-09-08
refactor(api): serve hits.total as a plain integer

PascalRepond:
* Serve `hits.total` as the plain integer upstream Invenio returns,
  in place of the four shapes the API answered with so far.
* Unwrap the search engine shape in the serializers alone, so the
  rest of the application keeps reading the engine response as is.
* Ask the search engine for an exact count on the routes that build
  their own query, where the total stopped growing at 10 000.
* Part of #4220.


rerowep (changes requested):
rero_ils/modules/sru/           — no track_total_hits
rero_ils/modules/documents/serializers/  — no
rero_ils/modules/serializers/   — no
rero_ils/modules/items/views/   — no

Recommendation: land the int, and add .extra(track_total_hits=True) to every custom search in the same PR — SRU above all. Int-without-exactness is strictly worse than the dict it replaces.

One tradeoff to decide consciously: track_total_hits=True disables ES's early-termination optimisation, so broad queries on a large documents index pay a real cost

PascalRepond:
> rero_ils/modules/sru/ — no track_total_hits rero_ils/modules/documents/serializers/ — no rero_ils/modules/serializers/ — no rero_ils/modules/items/views/ — no
> 
> Recommendation: land the int, and add .extra(track_total_hits=True) to every custom search in the same PR — SRU above all. Int-without-exactness is strictly worse than the dict it replaces.
> 
> One tradeoff to decide consciously: track_total_hits=True disables ES's early-termination optimisation, so broad queries on a large documents index pay a real cost

@rerowep SRU did not have exactness before: `"relation": "eq"` was hardcoded. And we don't need to use any `track_total_hits`, invenio-records-rest already does it.

scripts/bootstrap
```diff
@@ -86,7 +86,7 @@ then
   info_msg "Install RERO-ILS-UI from tgz: ${tgz_file}"
   npm install --no-save --only=prod --no-fund --no-audit  "${tgz_file}" --prefix "${static_folder}"
 else
-  npm install --no-save --only=prod --no-fund --no-audit  @rero/rero-ils-ui@21.1.0 --prefix "${static_folder}"
+  npm install --no-save --only=prod --no-fund --no-audit  @rero/rero-ils-ui@21.2.0 --prefix "${static_folder}"
```
jma: Is it required in this PR?

jma (approved):

PR #4256, opened 2026-09-10
feat(circulation): add requests exports

PascalRepond:
## 1. fix(circulation): restrict requests view by org

* Restrict the requests list to the librarians of its
  organisation, so that not any librarian could read the
  requests of any library, including one of another organisation.

## 2. feat(circulation): add requests exports

* Export the requests for a library as CSV, JSON or
  XLSX, so that librarians can work on the list outside of the
  professional interface.
* Build the export from the list the UI already uses,
  rather than from a dedicated search, so that both always show
  the same requests. This means that this specific export cannot
  use a standard `api_exports` without a much bigger refactor.
* Closes https://tree.taiga.io/project/rero21-reroils/us/3038.

## 3. refactor: share the display language resolution

* Move the resolution of the language used to display the i18n
  fields into `get_display_language` because the same three lines
  were repeated in the search query builder, in three exports and
  in the requests export just added.
* Keep the `lang` argument of the request, unused by any client yet,
  but useful when fetching exports directly from the API.

---

Linked to https://github.com/rero/rero-ils-ui/pull/1564

### DIFF
--- a/rero_ils/modules/decorators.py
+++ b/rero_ils/modules/decorators.py
@@ -17,6 +17,8 @@ from rero_ils.permissions import (
     login_and_patron,
 )
 
+from .libraries.api import Library
+from .patrons.api import current_librarian
 from .permissions import PermissionContext
 
 
@@ -44,6 +46,22 @@ def check_logged_as_librarian(fn):
     @wraps(fn)
     def wrapper(*args, **kwargs):
         login_and_librarian()
+        return fn(*args, **kwargs)
+
+    return wrapper
+
+
+def check_library_in_organisation(fn):
+    """Decorator to check the `library_pid` is part of the user organisation.
+
+    If the library belongs to another organisation: return 403 (forbidden)
+    """
+
+    @wraps(fn)
+    def wrapper(*args, **kwargs):
+        library = Library.get_record_by_pid(kwargs.get("library_pid"))
+        if library and (not current_librarian or library.organisation_pid != current_librarian.organisation_pid):
+            return jsonify({"status": "error: Forbidden"}), 403
         return fn(*args, **kwargs)
 
     return wrapper
--- a/rero_ils/modules/loans/api_views.py
+++ b/rero_ils/modules/loans/api_views.py
@@ -3,21 +3,47 @@
 
 """Blueprint used for loans."""
 
-from flask import Blueprint, abort, jsonify
+from flask import Blueprint, abort, jsonify, request
 from flask_login import login_required
 
-from rero_ils.modules.decorators import check_logged_as_librarian
+from rero_ils.modules.decorators import (
+    check_library_in_organisation,
+    check_logged_as_librarian,
+)
 from rero_ils.modules.items.api import Item
 from rero_ils.modules.items.models import ItemCirculationAction
 from rero_ils.modules.items.views.api_views import (
     check_logged_user_authentication,
     jsonify_error,
 )
+from rero_ils.modules.libraries.api import Library
 from rero_ils.modules.loans.api import Loan
+from rero_ils.modules.loans.serializers import (
+    csv_requests_search,
+    json_requests_search,
+    requests_to_validate_rows,
+    xlsx_requests_search,
+)
 from rero_ils.modules.loans.utils import get_circ_policy, sum_for_fees
 from rero_ils.modules.patrons.api import current_librarian, current_patrons
 
 api_blueprint = Blueprint("api_loan", __name__, url_prefix="/loan")
+
+
+@api_blueprint.route("/requests/<library_pid>", methods=["GET"])
+@check_logged_as_librarian
+@check_library_in_organisation
+def requests_export(library_pid):
+    """HTTP GET request to export the requests to validate of a library.
+
+    The exported requests are the same as the ones listed by the professional
+    interface. The output format is given by the `format` argument.
+    """
+    serializers = {"csv": csv_requests_search, "json": json_requests_search, "xlsx": xlsx_requests_search}
+    if not (serializer := serializers.get(request.args.get("format", "csv"))):
+        abort(400, "Unsupported export format")
+    library = Library.get_record_by_pid(library_pid)
+    return serializer(None, requests_to_validate_rows(library) if library else [])
 
 
 @api_blueprint.route("/<loan_pid>/circulation_policy", methods=["GET"])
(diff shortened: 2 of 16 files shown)


rero_ils/modules/decorators.py
```diff
@@ -49,6 +51,22 @@ def wrapper(*args, **kwargs):
...
+    """Decorator to check the `library_pid` is part of the user organisation.
+
+    If the library belongs to another organisation: return 403 (forbidden)
+    """
+
+    @wraps(fn)
+    def wrapper(*args, **kwargs):
+        library = Library.get_record_by_pid(kwargs.get("library_pid"))
+        if library and (not current_librarian or library.organisation_pid != current_librarian.organisation_pid):
+            return jsonify({"status": "error: Forbidden"}), 403
```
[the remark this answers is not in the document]
PascalRepond: These errors are technical, they are never translated in the app.

rero_ils/modules/loans/serializers/__init__.py
```diff
@@ -51,3 +60,36 @@
...
+        "document_title",
+        "document_contributions",
+        "item_first_call_number",
+        "item_second_call_number",
+        "enumerationAndChronology",
+        "requested_by",
+        "item_location",
+        "pickup_location",
+        "request_date",
+    ]
```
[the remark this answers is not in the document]
PascalRepond: For now, the headers are never localized in the app. This would be a more global refactor and does not belong here.

rero_ils/modules/loans/serializers/__init__.py
```diff
@@ -51,3 +60,36 @@
...
+xlsx_requests_search = search_responsify_file(
+    _requests_csv,
+    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+    file_extension="xlsx",
+    file_prefix="export-requests",
+    content_converter=xlsx_converter(
+        "rero_ils/exports/requests.xml",
+        worksheet_name="Requests",
+    ),
+)
```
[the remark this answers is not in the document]
PascalRepond: Not standard practice in the rest of the app.

rero_ils/modules/loans/serializers/csv.py
```diff
@@ -109,3 +109,28 @@ def generate_csv():
...
+        """
+
+        def generate_csv():
+            """Generate the CSV content as a generator."""
+            line = Line()
+            writer = DictWriter(line, dialect="excel", quoting=QUOTE_ALL, fieldnames=self.csv_included_fields)
+            writer.writeheader()
+            yield line.read()
+            for row in search_result:
+                writer.writerow(row)
```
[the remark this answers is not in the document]
PascalRepond: Not standard pracice in the rest of the app, needs a global refactor if it were to be fixed.

rero_ils/theme/templates/rero_ils/exports/requests.xml
```diff
@@ -0,0 +1 @@
+{% include "rero_ils/exports/_xlsx_sheet.xml" %}
```
anouillz, line 1: It would be nice to also give the name of the columns that may be a number, a date or a pid. That way they will be treated as such on excel and the user does not need to format them manually. You could look at the other templates.

anouillz (approved):

rero_ils/theme/templates/rero_ils/exports/requests.xml
PascalRepond: You're right, but for this one export all fields are strings. `request_date` (date-time) should not be treated as an Excel date, because it would lose the time.

rero_ils/modules/loans/serializers/base.py
```diff
@@ -0,0 +1,99 @@
+# SPDX-FileCopyrightText: Fondation RERO+
```
jma, line 1: I would not put this in a base.py. By convention serializers/base.py contains the base class of a serializer. Use utils.py or similar instead.

rero_ils/modules/loans/serializers/base.py
```diff
@@ -0,0 +1,99 @@
...
+
+from rero_ils.modules.documents.api import DocumentsSearch
+from rero_ils.modules.documents.extensions import TitleExtension
+from rero_ils.modules.holdings.api import HoldingsSearch
+from rero_ils.utils import get_display_language
+
+from ..api import Loan
+
+
+def _document_contributions(document, language):
```
jma, line 17: Maybe this code already exists in the project.

rero_ils/modules/loans/serializers/json.py
```diff
@@ -135,3 +137,17 @@ def _enrich_buckets(buckets, search_class, attribute):
             aggregations["misc_status"]["buckets"] = [
                 {"key": term, "doc_count": hit["doc_count"]} for term, hit in misc_aggr.items() if hit.get("doc_count")
             ]
+
+
+class RequestsToValidateJSONSerializer:
```
jma, line 142: Why this serializer is not derived by an other serializer? If the idea is to dumps a json, from my memory you can do it using json.dumps in a lambda function directly in the config.py.

rero_ils/modules/loans/api_views.py
```diff
@@ -3,23 +3,49 @@
...
 from rero_ils.modules.loans.utils import get_circ_policy, sum_for_fees
 from rero_ils.modules.patrons.api import current_librarian, current_patrons
 
 api_blueprint = Blueprint("api_loan", __name__, url_prefix="/loan")
 
 
+@api_blueprint.route("/requests/<library_pid>", methods=["GET"])
+@check_logged_as_librarian
+@check_library_in_organisation
+def requests_export(library_pid):
```
jma, line 36: export_requests?

rero_ils/modules/loans/api_views.py
```diff
@@ -3,23 +3,49 @@
...
+)
 from rero_ils.modules.loans.utils import get_circ_policy, sum_for_fees
 from rero_ils.modules.patrons.api import current_librarian, current_patrons
 
 api_blueprint = Blueprint("api_loan", __name__, url_prefix="/loan")
 
 
+@api_blueprint.route("/requests/<library_pid>", methods=["GET"])
+@check_logged_as_librarian
+@check_library_in_organisation
```
jma, line 35: It would be better to use ContentNegotiatedMethodView.

jma (approved):

rero_ils/modules/loans/serializers/base.py
PascalRepond: Not exactly as is.

PR #4257, opened 2026-09-11
fix(entities): query each MEF type endpoint

rerowep:
* Add the missing places endpoint.
* Group the pids by entity type, each type has its own endpoint.
* Raise on a non 200 answer to not skip the updated records.


rero_ils/modules/entities/remote_entities/sync.py
```diff
@@ -199,27 +203,32 @@ def get_mef_pids(search_query, chunk_size=1000):
...
+                        # each MEF entity type has its own endpoint: a pid sent
+                        # to the endpoint of another type is reported back as
+                        # an unknown record.
+                        pids_by_type = {}
+                        for pid, entity_type in chunk:
+                            if mef_type := entity_types.get(entity_type):
+                                pids_by_type.setdefault(mef_type, []).append(pid)
+                        for mef_type, mef_pids in pids_by_type.items():
+                            url = f"{get_mef_url(mef_type)}/mef/updated"
+                            # ask the mef server to return only the updated pids form a given date
```
PascalRepond, line 223: form -> from*

PascalRepond (approved):

PR #4259, opened 2026-09-14
fix(items): display circulation information

### LINKED ISSUE #4243: Display circulation information even for items types that are "available"
### Bug description:

Items that have an item type set as "available" display "on shelf" in the status/availability instead of the "circulation information" from the item type.

### Expected behavior:

Display the circulation information from the item type when it exists even when the item is "available".

anouillz:
* Show circulation details for available item types.
* Closes #4243.


rero_ils/modules/items/api/circulation.py
```diff
@@ -1280,6 +1280,11 @@ def availability_text(self):
         label = self.status
         if self.is_issue and self.issue_status != ItemIssueStatus.RECEIVED:
             label = self.issue_status
+        if label == ItemStatus.ON_SHELF:
+            return [
+                *circ_category.get("circulation_information", []),
```
rerowep, line 1285: We are sure we have always a valid circ_category here?

PascalRepond (approved):
Tested locally 👍

rero_ils/modules/items/api/circulation.py
anouillz: Yes. If `circ_category` was not valid, it would've raised an error in the `negative_availability` check a few lines above.

PR #4262, opened 2026-09-17
fix(requests): hide not received serial issues

### LINKED ISSUE #3601: Issues that don't have status recieved should not be displayed in the requests list
### How it works
Currently, serial issues that don't have the status recieved are not displayed in the public UI but they can still be requested in the API/pro UI and are then displayed in the requests lists in the pro UI.

As these issues do not exist physically yet, they should not be shown in the requests lists when requested (same as an item that is already on loan or in transit).

### Improvement suggestion
Filter out item/issues that do not have the status "recieved" to be displayed in the pro requests list.

- Either automatically validate requests on items/issues with a specific status in the backend so that they are LoanState.IN_TRANSIT_FOR_PICKUP, and are not shown in this list
- Or use the item data in the ui to filter out (ONLY IN THIS LIST) the loans for these items.

anouillz:
Serial issues that are requested but were not received yet do not appear on the request list anymore. They will appear once their status becomes "received".

* Exclude items whose issue status is not "received" from the request- to-validate item lookup query.
* Closes #3601.


anouillz:
@coderabbitai review

rero_ils/modules/loans/api.py
```diff
@@ -370,6 +370,13 @@ def item_by_pid(pid, known_items):
                     ItemsSearch()
                     .filter("term", pid=pid)
                     .filter("term", status=ItemStatus.ON_SHELF)
+                    # exclude issues that have not been physically received yet
+                    .filter(
+                        Q(
+                            "bool",
+                            must_not=[Q("exists", field="issue") & ~Q("term", issue__status=ItemIssueStatus.RECEIVED)],
+                        )
+                    )
```
PascalRepond: More readable version:

```python
.exclude("terms", issue__status=[ItemIssueStatus.EXPECTED, ItemIssueStatus.LATE, ItemIssueStatus.DELETED])
```

PascalRepond (approved):

rero_ils/modules/loans/api.py
```diff
@@ -370,6 +370,10 @@ def item_by_pid(pid, known_items):
                     ItemsSearch()
                     .filter("term", pid=pid)
                     .filter("term", status=ItemStatus.ON_SHELF)
+                    # exclude issues that have not been physically received yet
```
rerowep, line 373: Now the definition is not the same as the function anymore. Should we add a parameter to exclude not received items?

rerowep (changes requested):

rerowep (approved):

